PR #2, opened 2019-08-28
Rename cartridge.pre script without dot prefix

rosik:


rosik:
I don't want to merge it anymore. Renaming one file wont help UX, and will break many readmes. Let it stay hidden

PR #4, opened 2019-08-29
Switch from tap to luatest

printercu:
Enable luacheck on CI, fix warnings in templates


Makefile
```diff
@@ -0,0 +1,25 @@
...
+	tarantoolctl rocks install luatest 0.2.0
+	tarantoolctl rocks install luacheck
+
+.PHONY: lint
+lint: bootstrap
+	.rocks/bin/luacheck ./
+
+.PHONY: test
+test: luatest pytest
+	.rocks/bin/luatest
```
sharonovd: Looks like luatest will be called twice this way

PR #9, opened 2019-08-29
Start and stop all instances

printercu:


cartridge
```diff
@@ -1953,12 +1953,29 @@ local cmd_start = {
...
+        When INSTANCE_NAME is not provided it reads `cfg` file and starts all
+        defined instances.
+    ]=]):format(self_name):gsub('(\n?)' .. (' '):rep(8), '%1'),
 }
 
+-- Fetches app_name from .rockspec file.
+-- Extracted from cartridge.argparse, but searches for rockspec only in current
+-- directory.
+local function get_app_name()
+    local rockspecs = fio.glob('*-scm-1.rockspec')
```
RunsFor: What if there is no such file in the directory. How it will behave? I think it should throw user an error (or usage)

cartridge
```diff
@@ -2038,6 +2098,11 @@ end
 -- is successfully bootstraped: after tarantool executes main script
 -- it sends `READY=1` message to the NOTIFY_SOCKET.
 function cmd_start.callback(args)
+    if args.instance_name == nil then
```
RunsFor: Is it possible to run single application without specifying instance name and configuration path using simple command:

```bash
cartridge start myapp
```

printercu: Yep, there are examples in readme

cartridge
printercu: It can be set in different ways. Final value is checked on line 2070

cartridge
RunsFor: I tried and got an error:

```bash
$ cartridge create --template cartridge --name myapp
$ cd myapp
$ rm myapp-scm-1.rockspec
$ cartridge start myapp
error: Failed to open file /Users/i.konyuhov/Workspace/cartridge-cli/myapp/instances.yml: No such file or directory
```

cartridge
```diff
@@ -2030,6 +2053,43 @@ local function check_pid_running(pid)
...
+        ):map(function(x) return read_configuration(x) end):totable()
+        return fun.chain(unpack(configs)):tomap()
+    else
+        return yaml.decode(read_file(path))
+    end
+end
+
+-- Read configuration at `path` and fetch instance names.
+local function get_configured_isntances(path, app_name)
+    assert(app_name and #app_name > 0, 'app_name must be given')
```
RunsFor: Oh, I see, It actually asserts that app_name must be given. But I think this should be done on the `parse` step.

cartridge
RunsFor: Continued [here](https://github.com/tarantool/cartridge-cli/pull/9/files#r321174638)

cartridge
printercu: You need to create configuration file with instances list. Otherwise it can not know which instances do you want it to start.

cartridge
printercu: I don't think so. There it tries to figure out app name when it was not explicitly given. It just returns nothing when it can not do this.

RunsFor: But you don't check if there is `app_name` given, when instance_name provided. I.e:

```
cartridge start .instance_name
```

cartridge
RunsFor: I just want to start single application (maybe test something). Should it be necessary to write additional configuration file (or provide additional cmd arguments) for such case?

printercu: `$ ./init.lua` should do the work

printercu: `cartridge start .instance_name` should also work

cartridge
printercu: It was not required at that time. I've changed behaviour, app_name is always required and it's checked in `.parse` method.

PR #15, opened 2019-09-02
Use tarantool docker images in ci

printercu:


Makefile
```diff
@@ -19,6 +19,11 @@ luatest: bootstrap
 pytest: bootstrap
 	python3.6 -m pytest -vvl
 
+.PHONY: ci_prepare
+ci_prepare:
+	git config --global user.email "test@tarantool.io"
+	git config --global user.name "Tar Antool"
```
RunsFor: `Taran Tool` ?

printercu: :) just copy-pasted. will change to Test Tarantool

printercu:
I want to simplify gitlab-ci.yml and speedup builds. Your suggestion is not on this way.

Scripts you're pointing at are for installation and providing docker entry point. They don't affect `tarantool` command.

rosik:
Your PR doesn't simplify anything but moves the complexity to another place. Simplify it without bringing additional dependencies

PR #16, opened 2019-09-02
Cache downloaded sdk between ci jobs

printercu:
image


RunsFor:
Should we consider to migrate to travis-ci?

printercu:
We need at least jobs for tarantool-enterprise to be run on private gitlab runners.

PR #26, opened 2019-09-27
Colorize logs and prefix with instance name for multiple foreground instances

printercu:
image


knazarov:
I see that the actual traceback is not colored. I suggest to store the last known error level for instance’s log lines and apply it to current line if we don’t see the level marker

printercu:
Updated
image

PR #28, opened 2019-10-03
Pack DEB

dokshina:


test/e2e/rpm.sh
```diff
@@ -47,7 +36,7 @@ curl -w "\n" -X POST http://127.0.0.1:8081/admin/api --fail -d@- <<QUERY
     "mutation {
         j1: join_server(
             uri:\"$IP:3301\",
-            roles: [\"app.roles.custom\"]
+            roles: [\"vshard-router\", \"app.roles.custom\"]
```
rosik: Revert it please. The idea is to check that vshard-router is a dependency of custom role

PR #49, opened 2019-10-30
Add cartridge-cli when packing an app

printercu:


test/python/test_pack.py
```diff
@@ -53,14 +51,19 @@
...
 ])
 
+if tarantool_enterprise_is_used():
+    original_file_tree |= set([
+        'cartridge',
+        'tarantool',
+        'tarantoolctl',
+    ])
+    original_rocks_content |= set([
+        '.rocks/share/tarantool/rocks/cartridge-cli/1.0.0-1/bin/cartridge',
```
sharonovd: 1.0.0 ?

printercu:
Decided not to proceed with this pr. cartridge-cli should be installed system-wide

PR #60, opened 2019-11-11
Start instances in foreground by default

printercu:


README.md
```diff
@@ -52,11 +52,11 @@ Options
...
                         Defaults to TARANTOOL_CFG or ./instances.yml
 
-    --foreground        Do not daemonize
+    --daemonize / -d    Start in background
 ```
 
-It starts a `tarantool` instance in background with enforced env-vars and
-waits until the app's main script is finished.
+It starts a `tarantool` instance with enforced env-vars.
+With `--daemonize` options it also waits until the app's main script is finished.
```
lenkis: With the `--daemonize` option... [add "the" and remove "-s"]

PR #77, opened 2019-12-06
Remove redundant package.path modification

printercu:


rosik:
All tests are red, it wasn't redundant :-)

olegrok:
As soon as CI is red and there were no changes since December, I think this PR should be closed.

Feel free to reopen

printercu:
These path modifications are useful only for tests. They are useless in the real installations because rock content is saved to `.rocks/share/tarantool/cartridge-cli/` and executable is placed to `.rocks/bin/`.

PR #90, opened 2019-12-23
Change the parse priority of option sources

HustonMmmavr:


README.md
```diff
@@ -340,8 +340,11 @@ Default options for the `cartridge` command can be overridden in
 run_dir: tmp/run
 cfg: cartridge.yml
 apps_path: /usr/local/share/tarantool
+script: init.lua
 ```
 
+Also options from `.cartridge.yml` are overridden by corresponding to them environment variables.
+
```
printercu: Let's move this to the previous sentence "... can be overridden with `TARANTOOL_*` environment variables or in ...". There is also same text in help message in lua file at `cmd_start` definition.

PR #91, opened 2019-12-23
New pack flow

### LINKED ISSUE #86: Consider an alternative to old-way .ignore
The main idea of .cartridge.ignore was to clear useless artefacts after building (e.g. node_modules/some docs temporary files)

However, now .cartridge.ignore is used before build. So, user can't control that will be packed after build.

Let's consider an alternative to old-way approach. It could be a ".cartridge.pack.ignore" or something else.

dokshina:
Fix backward compatible with `.cartridge.ignore + .cartridge.pre` 
Support new  Use `cartridge.pre-build` and `cartridge.post-build` files to specify additional actions to be done before and after `tarantoolctl rocks make`
Separate docker packing tests
Use parametrized `project` fixture to test both original and deprecated flows
Use coloured error, warn and log messages

Closes #86

### DIFF
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -1171,20 +1171,27 @@ local function form_distribution_dir(source_dir, dest_dir)
     check_filemodes(dest_dir)
 end
 
+local function run_hook(dir, filename)
+    info('Running %s', filename)
+    assert(fio.path.exists(fio.pathjoin(dir, filename)))
+
+    local ret = os.execute(
+        'set -e\n' ..
+        string.format('cd %q\n', dir) ..
+        string.format('. ./%s', filename)
+    )
+
+    if ret ~= 0 then
+        die('Failed to execute %s', filename)
+    end
+
+    remove_by_path(fio.pathjoin(dir, filename))
+end
+
 local function build_application(dir)
     -- pre build
     if fio.path.exists(fio.pathjoin(dir, PREBUILD_SCRIPT_NAME)) then
-        info('Running %s', PREBUILD_SCRIPT_NAME)
-        local ret = os.execute(
-            'set -e\n' ..
-            string.format('cd %q\n', dir) ..
-            string.format('. ./%s', PREBUILD_SCRIPT_NAME)
-        )
-        if ret ~= 0 then
-            die('Failed to execute pre-build stage')
-        end
-
-        remove_by_path(fio.pathjoin(dir, PREBUILD_SCRIPT_NAME))
+        run_hook(dir, PREBUILD_SCRIPT_NAME)
     end
 
     -- build
@@ -1204,17 +1211,7 @@ local function build_application(dir)
 
     -- post build
     if fio.path.exists(fio.pathjoin(dir, POSTBUILD_SCRIPT_NAME)) then
-        info('Running %s', POSTBUILD_SCRIPT_NAME)
-        local ret = os.execute(
-            'set -e\n' ..
-            string.format('cd %q\n', dir) ..
-            string.format('. ./%s', POSTBUILD_SCRIPT_NAME)
-        )
-        if ret ~= 0 then
-            die('Failed to execute post-build stage')
-        end
-
-        remove_by_path(fio.pathjoin(dir, POSTBUILD_SCRIPT_NAME))
+        run_hook(dir, POSTBUILD_SCRIPT_NAME)
     end
 end
 
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -791,6 +791,11 @@ local function detect_name_release_version(source_dir, raw_name, raw_version)
     return name, release, version
 end
 
+-- * ----------- Special filenames ------------
+
+local PREBUILD_SCRIPT_NAME = 'cartridge.pre-build'
+local POSTBUILD_SCRIPT_NAME = 'cartridge.post-build'
+
 -- * --------------- Preinstall ---------------
 
 local CREATE_USER_SCRIPT = [[
@@ -932,12 +937,16 @@ COPY . ${dir}
 
 WORKDIR ${dir}
 
-RUN if [ -f .cartridge.pre ]; then \
-       set -e && . .cartridge.pre && rm .cartridge.pre; \
+RUN if [ -f ${prebuild_script_name} ]; then \
+       set -e && . ${prebuild_script_name} && rm ${prebuild_script_name}; \
     fi
 
 RUN if ls *.rockspec 1> /dev/null 2>&1; then \
         tarantoolctl rocks make; \
+    fi
+
+RUN if [ -f ${postbuild_script_name} ]; then \
+        set -e && . ${postbuild_script_name} && rm ${postbuild_script_name}; \
     fi
 
 USER tarantool:tarantool
@@ -1068,39 +1077,6 @@ local function generate_version_file(source_dir, dest_dir, app_name, app_version
     version_file:close()
 end
 
-local function pattern_form(pattern)
-    if pattern == '' or-- blank line
-            string.startswith(pattern, '#') then -- comment
-        return nil, false
-    end
-
-    if string.startswith(pattern, '\\#') then -- escape #
-        pattern = pattern:sub(2, #pattern)
-    end
-     -- trim space
-    pattern = pattern:gsub("%s+", "")
-
-    local negative = false
-    if string.startswith(pattern, '!') then
-        pattern = pattern:sub(2, #pattern)
-        negative = true
-    end
-    return pattern, negative
-end
-
-local function path_form(path, destdir)
-    if string.startswith(path, './') then
-        path = path:sub(3, #path)
-    end
-    -- if this is a folder, then added to end /
-    if fio.path.is_dir(fio.pathjoin(destdir, path)) then
-        path = path .. '/'
-    end
-
-
-    return path
-end
-
 local function matching(str, pattern)
     -- case: pattern <simple>, str [<simple> | <simple/>]
     if not string.endswith(pattern, '/') and string.endswith(str, '/') then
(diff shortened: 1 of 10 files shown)


olegrok:
I suggest you to split your patch into several parts:
- color log messages with this fixes (https://github.com/tarantool/cartridge-cli/pull/91/commits/86ec474c3451223d2c4ac5dd20a6a6275259cde4)
- separate docker pack tests (without new flow)
- new pack flow

olegrok:
> * color log messages with this fixes ([86ec474](https://github.com/tarantool/cartridge-

I think you can collect all your fixes to this one (I also found - https://github.com/tarantool/cartridge-cli/pull/91/commits/2f1d7498a2084d1e3089603a84a007ab16addeaa#diff-65269a74ae0126ab3c284f41ebda16adR294)

This line is not part of "pack flow"

cartridge-cli.lua
```diff
@@ -1112,7 +1133,7 @@ local function remove_ignored(destdir)
     local matched = {}
     for _, file in ipairs(files) do
         local pretty_file = path_form(file, destdir)
-        print(pretty_file)
+        -- print(pretty_file)
```
olegrok: Remove it, if it was used only for debug purposes

cartridge-cli.lua
```diff
@@ -1187,26 +1209,44 @@ local function form_distribution_dir(source_dir, dest_dir)
                  "normally ignored are shipped to the resulting package. ")
     end
 
-    -- deleting files matching patterns from .cartridge.ignore
-    remove_ignored(dest_dir)
+    if not opts.deprecated_build_flow then
```
olegrok: Let's move "deprecated build flow" into separate commit that will be easy to revert in future

test/python/conftest.py
```diff
@@ -1,9 +1,29 @@
...
+
 @pytest.fixture(scope='module')
 def module_tmpdir(request):
     dir = py.path.local(tempfile.mkdtemp())
     request.addfinalizer(lambda: dir.remove(rec=1))
     return str(dir)
+
+
+@pytest.fixture(scope="module")
+def project_path(module_tmpdir):
```
olegrok: I suggest you to add finalizer `request.addfinalizer(lambda: shutil.rmtree(path, ignore_errors=True))` to clean directories after test will be finished

test/python/test_docker_pack.py
```diff
@@ -0,0 +1,288 @@
...
+from utils import recursive_listdir
+from utils import assert_dir_contents
+from utils import assert_filemodes
+
+
+# #######
+# Helpers
+# #######
+def find_image(docker_client, project_name):
+    for image in docker_client.images.list():
```
olegrok: Why "get" isn't enough https://docker-py.readthedocs.io/en/stable/images.html#docker.models.images.ImageCollection.get ?

test/python/test_docker_pack.py
```diff
@@ -0,0 +1,288 @@
...
+    command = '/bin/bash -c "{}"'.format(command.replace('"', '\\"'))
+    output = docker_client.containers.run(
+        image_name,
+        command,
+        remove=True
+    )
+    return output.decode("utf-8").strip()
+
+
+def wait_for_container_start(container, timeout=10):
```
olegrok: I think it's useful to have in project "wait_for" helper.

`wait_for(function, args, timeout)`

test/python/utils.py
```diff
@@ -20,10 +23,117 @@ def tarantool_version():
...
+    '.editorconfig',
+    '.gitignore',
+    '.luacheckrc',
+    'deps.sh',
+    'init.lua',
+    'instances.yml',
+    'app',
+    'app/roles',
+    'app/roles/custom.lua',
+    # 'test',
```
olegrok: Why have you commented this lines?

If it's redundant just drop it

cartridge-cli.lua
```diff
@@ -2300,6 +2300,14 @@ local function app_pack(args)
             "Using `.cartridge.ignore` and `.cartridge.pre` files is deprecated in 1.3.0 " ..
                 "and will be removed in 2.0.0"
         )
+
+        if args.type == 'docker' then
+            die(
```
olegrok: Let's just ignore it with warning message

cartridge-cli.lua
```diff
@@ -1222,6 +1154,7 @@ local function form_distribution_dir(source_dir, dest_dir)
     end
     local git = which('git')
     if git ~= nil and fio.path.exists(fio.pathjoin(dest_dir, '.git')) then
+        info('Running `git clean`')
```
olegrok: Let's call "git describe --tag" before and save it to special env variable (e.g. GIT_APPLICATION_VERSION)

olegrok: It could be useful for some users

cartridge-cli.lua
```diff
@@ -1230,47 +1163,63 @@ local function form_distribution_dir(source_dir, dest_dir)
...
     -- check application files mode
+    info('Check application file modes')
     check_filemodes(dest_dir)
 end
 
 local function build_application(dir)
-    if fio.path.exists(fio.pathjoin(dir, '.cartridge.pre')) then
-        info("Running .cartridge.pre")
+    -- pre build
+    if fio.path.exists(fio.pathjoin(dir, PREBUILD_SCRIPT_NAME)) then
```
olegrok: This logic is too similar to `post build`... May be it's a good idea to move it into "run_hook" function

test/python/conftest.py
```diff
@@ -16,99 +16,14 @@ def module_tmpdir(request):
 
 @pytest.fixture(scope="module")
 def project_path(module_tmpdir):
-    return create_project(module_tmpdir, project_name, 'cartridge')
+    path = create_project(module_tmpdir, project_name, 'cartridge')
 
+    # add cartridge.post-build file to remove test/ and tmp/ contents
+    with open(os.path.join(path, 'cartridge.post-build'), 'w') as f:
```
olegrok: Where is a test for "pre-build"?

test/python/test_pack.py
```diff
@@ -113,33 +112,40 @@ def rpm_archive_with_custom_units(module_tmpdir, project_path):
...
+            project['path']
         ],
         cwd=module_tmpdir
     )
     assert process.returncode == 0, \
         "Error during creating of rpm archive with project"
 
-    archive_name = find_archive(module_tmpdir, 'rpm')
+    archive_name = find_archive(module_tmpdir, project['name'], 'rpm')
     assert archive_name is not None, "RPM archive isn't founded in work directory"
```
olegrok: founded -> found

test/python/test_pack.py
```diff
@@ -256,3 +264,27 @@ def test_packing_without_git(tmpdir):
         project_path,
     ]
     process = subprocess.run(cmd, cwd=tmpdir)
+
+
+def test_packing_with_wrong_filemodes(tmpdir):
```
olegrok: I think this test should be moved to separate patch/commit

It's not about "new package flow"

test/python/conftest.py
dokshina: This diff is outdated, I just checked that this test is here.

cartridge-cli.lua
dokshina: I like your idea, maybe we should write it in the VERSION file.
But in fact, if application version isn't passed explicitly, it will be detected from `git describe --tag`.

Anyway, I would like to discuss it in separated issue.

olegrok: https://github.com/tarantool/cartridge-cli/issues/100

test/python/conftest.py
```diff
@@ -14,101 +14,94 @@ def module_tmpdir(request):
...
+            "#!/bin/sh",
+            "rm -rf test tmp third_party"
+        ]
+        f.write('\n'.join(postbuild_script_lines))
+
+    # add custom-module to rocks content
+    rocks_content = get_rocks_content(project_name)
+    rocks_content.add('.rocks/share/tarantool/rocks/custom-module')
+
+    return {
```
olegrok: You return it as a module

Let's add a "Project" as class

olegrok: with options, methods and so on

dokshina: I like your idea, but it's not about new flow, so let's implement it in separated PR.

olegrok: don't forget that

test/python/test_pack.py
dokshina: See #102

cartridge-cli.lua
```diff
@@ -932,14 +937,18 @@ COPY . ${dir}
 
 WORKDIR ${dir}
 
-RUN if [ -f .cartridge.pre ]; then \
-       set -e && . .cartridge.pre && rm .cartridge.pre; \
+RUN if [ -f ${prebuild_script_name} ]; then \
```
olegrok: Please, group this several Runs into one. There are too may layers in result image - it's not ok

olegrok:
In general, I think your patch is OK. Please, fix my several notes.

Then split your patch into three commits (not separate PRs):
- main part with "new flow" 
- using "colored formatted" messages instead of prints
- tests fixes: project_path -> project.path

I want your commits to be atomic

dokshina:
I don't see any reason to split this patch to different commits.
It's not too large and will be squashed anyway.

I've already squashed some commits to easily separate this PR, and it was an important change. I don't think that separating this commits again is required.

cartridge-cli.lua
dokshina: I'm not sure that this action should be performed in the one RUN.
Are this layers too big?
I need more arguments to understand if it's required.

olegrok: The layer is quite cheap docker primitive, however "best practices" and examples shows us that we should decrease amount of layers. 

https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers

olegrok: In separate PR - ok

PR #92, opened 2019-12-25
Fix rocks manifest error on packing

### LINKED ISSUE #80: Get warning about rocks manifest for template application
## Error message after packing

> warning: can't process rocks manifest file. Dependency information can't be shipped to the resulting package.

## Reproducer:
```bash
tarantoolctl rocks install cartridge-cli
.rocks/bin/cartridge create
.rocks/bin/cartridge pack tgz myproject
```

dokshina:
Don't use `luarocks` module to find and parse manifest file
Write Tarantool version for opensource
Improve tests
XXX: doesn't work for packing to docker

closes #80

### DIFF
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -9,9 +9,6 @@ local log = require('log')
 local socket = require('socket')
 local yaml = require('yaml')
 
-local luarocks_searcher = require('luarocks.search')
-local luarocks_path = require('luarocks.path')
-
 local self_name = fio.basename(arg[0])
 
 local function VERSION()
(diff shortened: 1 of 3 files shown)


cartridge-cli.lua
```diff
@@ -922,19 +919,58 @@ ENV PATH="/usr/share/tarantool/tarantool-enterprise:${"$"}{PATH}"
...
-    query.exact_name = false
+    local function read_manifest_file(filepath)
+        local res = {}
+
+        local file_content = read_file(filepath)
+        file_content = file_content:gsub("^#![^\n]*\n", "")
+
+        local chunk, ok, err
+        if _VERSION == "Lua 5.1" then -- Lua 5.1
+            chunk, err = loadstring(file_content, filepath)
```
olegrok: Are we plan to support plain "Lua 5.1"?

"load" is available in LuaJIT and Tarantool. Also loadstring is deprecated function

olegrok: https://stackoverflow.com/questions/37641153/loadstring-function-replacement-in-latest-version

cartridge-cli.lua
```diff
@@ -922,19 +919,58 @@ ENV PATH="/usr/share/tarantool/tarantool-enterprise:${"$"}{PATH}"
...
+    end
 
-    local manifest_path = luarocks_path.rocks_dir(fio.pathjoin(project_dir, '.rocks'))
+    local dependencies = {}
+    -- XXX: fix manifest filepath compution
+    local manifest_filepath = fio.pathjoin(project_dir, '.rocks/share/tarantool/rocks/manifest')
 
-    local results = {}
-    luarocks_searcher.manifest_search(results, manifest_path, query)
+    if fio.path.exists(manifest_filepath) then
```
olegrok: Shouldn't we check that it's a file? (it could be a user created function, by mistake)

test/python/test_pack.py
```diff
@@ -353,19 +353,38 @@ def has_tail(path): return len(path.split('/', maxsplit=1)) > 1
...
     version_filepath = os.path.join(distribution_dir, 'VERSION')
     assert os.path.exists(version_filepath)
 
+    version_file_content = {}
     with open(version_filepath, 'r') as version_file:
-        version_content = version_file.read()
-        for key in original_keys:
-            assert '{}='.format(key) in version_content
+        file_lines = version_file.read().strip().split('\n')
+        print('LINES:', file_lines)
```
olegrok: Is it a debug?

test/python/test_pack.py
```diff
@@ -353,19 +353,38 @@ def has_tail(path): return len(path.split('/', maxsplit=1)) > 1
...
 
     version_filepath = os.path.join(distribution_dir, 'VERSION')
     assert os.path.exists(version_filepath)
 
+    version_file_content = {}
     with open(version_filepath, 'r') as version_file:
-        version_content = version_file.read()
-        for key in original_keys:
-            assert '{}='.format(key) in version_content
+        file_lines = version_file.read().strip().split('\n')
```
olegrok: may be `splitlines()`?

cartridge-cli.lua
```diff
@@ -9,9 +9,6 @@ local log = require('log')
 local socket = require('socket')
 local yaml = require('yaml')
 
-local luarocks_searcher = require('luarocks.search')
```
olegrok: Make sure that we don't use luarocks 3 in another parts of cartridge-cli

cartridge-cli.lua
artur-barsegyan: What needs for the second argument? As I understand, `loadstring` receives only one argument

olegrok: > What needs for the second argument? As I understand, `loadstring` receives only one argument

This argument improves verbosity in case of error in loadstring function

test/python/test_pack.py
dokshina: Yes, thanks

test/python/test_pack.py
dokshina: If file ends with `\n\n`, `splitlines` will return `[..., '']` (last element is empty string).
I prefer using `strip + split('\n')`

cartridge-cli.lua
dokshina: The second argument is needed to form correct error msg.

Anyway, I don't use `loadstring` function anymore, but for `load` call I pass filename as a `source` parameter. I'm not sure, that it's right, but it's not so important.

cartridge-cli.lua
dokshina: `luarocks.*` modules was used only for version detection, it you mean that.

cartridge-cli.lua
dokshina: I added a check.

olegrok:
Your patch looks good to me, thanks!

Please, squash your changes before push

PR #95, opened 2019-12-25
Refactor tests

dokshina:


test/python/utils.py
```diff
@@ -20,11 +28,271 @@ def tarantool_version():
...
+    assert len(systemd_files) == 2
+    assert '{}.service'.format(project_name) in systemd_files
+    assert '{}@.service'.format(project_name) in systemd_files
+
+
+def check_package_files(basedir, project_path):
+    # check if only theese files are delivered
+    for filename in recursive_listdir(basedir):
+        assert any([
+            filename.startswith(prefix) or prefix.startswith(filename)
```
olegrok: why?

test/python/test_docker_pack.py
```diff
@@ -0,0 +1,288 @@
...
+from utils import recursive_listdir
+from utils import assert_dir_contents
+from utils import assert_filemodes
+
+
+# #######
+# Helpers
+# #######
+def find_image(docker_client, project_name):
+    for image in docker_client.images.list():
```
olegrok: why not "get"?

test/python/test_docker_pack.py
```diff
@@ -0,0 +1,288 @@
...
+    command = '/bin/bash -c "{}"'.format(command.replace('"', '\\"'))
+    output = docker_client.containers.run(
+        image_name,
+        command,
+        remove=True
+    )
+    return output.decode("utf-8").strip()
+
+
+def wait_for_container_start(container, timeout=10):
```
olegrok: I think it could be useful to add "wait_for" function to helpers

test/python/test_docker_pack.py
```diff
@@ -0,0 +1,288 @@
...
+
+    return False
+
+
+# ########
+# Fixtures
+# ########
+@pytest.fixture(scope="module")
+def docker_client():
+    client = docker.from_env()
```
olegrok: Should we disable docker tests or show beauty message in case then docker is unavailable

test/python/test_docker_pack.py
```diff
@@ -0,0 +1,288 @@
...
+    # check if distribution dir was created
+    command = '[ -d "{}" ] && echo true || echo false'.format(container_distribution_dir)
+    output = run_command_on_image(docker_client, image_name, command)
+    assert output == 'true'
+
+    # get distribution dir contents
+    arhive_path = os.path.join(tmpdir, 'distribution_dir.tar')
+    with open(arhive_path, 'wb') as f:
+        bits, _ = container.get_archive(container_distribution_dir)
+        for chunk in bits:
```
olegrok: Looks a bit strange?
Why don't we use f.write(bits)?

test/python/test_docker_pack.py
```diff
@@ -0,0 +1,288 @@
...
+        arch.extractall(path=os.path.join(tmpdir, 'usr/share/tarantool'))
+    os.remove(arhive_path)
+
+    assert_dir_contents(
+        recursive_listdir(os.path.join(tmpdir, 'usr/share/tarantool/', project_name)),
+        skip_tarantool_binaries=True
+    )
+
+    assert_filemodes(tmpdir)
+    container.remove()
```
olegrok: We don't destroy container if tests failes because of assert_filemodes. Consider adding of finalizer and moving a bit higher

test/python/test_docker_pack.py
```diff
@@ -0,0 +1,288 @@
...
+
+    if tarantool_enterprise_is_used():
```
olegrok: I like the idea of split such logic into two separate tests `test_docker_pack_community`/`test_docker_pack_enterprise` with conditional running

test/python/test_docker_pack.py
dokshina: I just copied this code from documentation, I think it could be important to use cycle here.

test/python/test_docker_pack.py
dokshina: `docker_image` finalizer removes all containers based on this image.

test/python/utils.py
dokshina: Because `recursive_listdir` result contains smth like `[..., /usr, /usr/share, /usr/share/tarantool/, /usr/share/tarantool/myapp ...]`.

test/python/test_docker_pack.py
dokshina: Don't want to hardcode image name.

test/python/utils.py
olegrok: You can use "substring in string"

test/python/test_docker_pack.py
dokshina: Maybe, but I have no idea where it could be useful except of testing docker.

test/python/utils.py
dokshina: But I want to check absolute filepaths.
What about smth like `filename = '/tmp/whoops/usr/share/tarantool/myapp'`? 
`'/usr/share/tarantool/myapp' in filename` will return `true`.

olegrok: Why do you use relpath in recursive listdir?

test/python/test_docker_pack.py
dokshina: I'm not sure about it.

Maybe it would be better to separate `tarantool` dependency or binaries checks to a local function.

test/python/utils.py
dokshina: Because files are unpacked to some temporarily directory in tests.

After package installation they will be unpacked to the `/` dir, and absolute path will be `/<relpath>`.

test/python/test_docker_pack.py
dokshina: I'm not sure.
I think it's important to test docker so I don't like the idea of disabling this tests.
And I sure that python will fail with some beauty message anyway.

test/python/test_docker_pack.py
olegrok: Got it.

It's an iterator/stream, not whole data

test/python/utils.py
```diff
@@ -20,11 +28,271 @@ def tarantool_version():
...
+
+
+def find_archive(path, arch_ext):
+    with os.scandir(path) as it:
+        for entry in it:
+            if entry.name.endswith('.' + arch_ext) and entry.is_file():
+                return os.path.join(path, entry.name)
+
+
+def recursive_listdir(root_dir):
```
olegrok: Try to use `glob.glob('**', recursive=True)`

dokshina: It sounds good, but in fact `glob.glob('path/to/dir/**')` will return `[...'path/to/dir/subdir', .'path/to/dir/subdir/file', ...]` paths. So, this call requres some additional actions to extract relpath and it can't replace using `recursive_listdir` function.

test/python/test_pack.py
```diff
@@ -331,221 +116,11 @@ def rpm_archive_with_custom_units(module_tmpdir, project_path, prepare_ignore):
...
-        for directory in directories:
-            files.add(os.path.join(rel_dir, directory))
-
-        for filename in filenames:
-            files.add(os.path.join(rel_dir, filename))
-
-    return files
-
-
-def assert_tarantool_dependency_rpm(filename):
```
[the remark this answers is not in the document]
olegrok: You moved it to utils https://github.com/tarantool/cartridge-cli/pull/95/files#diff-316f09b497bb77b8db4d532696a927aeR225

test/python/utils.py
```diff
@@ -20,11 +28,271 @@ def tarantool_version():
...
+    distribution_dir = os.path.join(basedir, 'usr/share/tarantool', project_name)
+    assert os.path.exists(distribution_dir)
+    assert_dir_contents(recursive_listdir(distribution_dir))
+
+    # check systemd dir content
+    check_systemd_dir(basedir)
+
+    # check tmpfiles conf
+    project_tmpfiles_conf_file = os.path.join(basedir, 'usr/lib/tmpfiles.d', '%s.conf' % project_name)
+    assert open(project_tmpfiles_conf_file).read().find('d /var/run/tarantool') != -1
```
olegrok: You don't close a file here

test/python/utils.py
```diff
@@ -20,11 +28,271 @@ def tarantool_version():
...
+def assert_files_mode_and_owner_rpm(filename):
+    DIRNAMES_TAG = 1118
+    DIRINDEXES_TAG = 1116
+
+    expected_tags = [
+        'basenames', DIRNAMES_TAG, DIRINDEXES_TAG, 'filemodes',
+        'fileusername', 'filegroupname'
+    ]
+
+    rpm = rpmfile.open(filename)
```
olegrok: You don't close this file

You can use `with rpmfile.open()...`

test/python/utils.py
```diff
@@ -20,11 +28,271 @@ def tarantool_version():
...
+        min_version = re.findall(r'\d+\.\d+\.\d+', tarantool_version())[0]
+        max_version = str(int(re.findall(r'\d+', tarantool_version())[0]) + 1)
+
+        deps = depends_str.group(1)
+        assert 'tarantool (>= {})'.format(min_version) in deps
+        assert 'tarantool (<< {})'.format(max_version) in deps
+
+
+def assert_tarantool_dependency_rpm(filename):
+    rpm = rpmfile.open(filename)
```
olegrok: File is not closed

test/python/utils.py
```diff
@@ -20,11 +28,271 @@ def tarantool_version():
...
+
+
+def validate_version_file(distribution_dir):
+    original_keys = [
+        'TARANTOOL',
+        project_name,
+        # default app dependencies
+        'luatest',
+        'cartridge',
+        # known cartridge dependencies
```
olegrok: `ddl` for new versions...

I'm not sure that we should hardcode it in test - it could be different for different versions

test/python/test_docker_pack.py
```diff
@@ -0,0 +1,288 @@
...
+        dir_contents = [
+            os.path.normpath(filename)
+            for filename in files_list
+        ]
+
+        assert 'tarantool' in dir_contents
+        assert 'tarantoolctl' in dir_contents
+    else:
+        # check if tarantool was installed
+        command = 'yum list installed 2>/dev/null | grep tarantool'
```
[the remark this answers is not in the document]
dokshina: I want to check if Tarantool was installed. If not, I will get an error on running `tarantool --version`,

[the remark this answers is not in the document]
olegrok: Yes, but "yum" is an implementation details. One day we will want to use tarantool installed from source and it will break tests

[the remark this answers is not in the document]
olegrok: If you get an error on running `tarantool --version` - tarantool is not installed

test/python/test_docker_pack.py
```diff
@@ -0,0 +1,288 @@
...
+        assert installed_version == expected_version
+
+
+def test_base_dockerfile_with_env_vars(project_path, module_tmpdir, tmpdir):
+    dockerfile_with_env = os.path.join(tmpdir, 'Dockerfile')
+    with open(dockerfile_with_env, 'w') as f:
+        f.write('''
+            FROM centos:8
+            # comment this string to use cached image
+            # ENV TEST_VARIABLE=${TEST_VARIABLE}
```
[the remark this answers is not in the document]
dokshina: The main idea of this test is to check if `cartridge pack` command don't fail on using `${name}` constructions in the base Dockerfile. 

It's about `expand` function specifics. It interprets `${name}` construction as a template for substitution `name` variable value, so I don't pass the base Dockerfile to this function and this test will fail if somebody start to do it.

So, I don't want to check if Docker ENV option works, I just want to protect base Dockerfile from passing to `expand` function.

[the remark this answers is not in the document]
olegrok: May be you can add small note about it

test/python/utils.py
dokshina: O, thanks! This fix was moved to the #91 PR, I will move it here.

test/python/utils.py
dokshina: Yes, you right, thanks!

test/python/utils.py
dokshina: I was thinking about it and tried to list application directory and then filter out some files, but it turned out to be not so easy. I decided that this change needs separated PR in the future.

Now, the `cartridge` version is hardcoded in application template, and you should fix this files list after changing it.

test/python/test_docker_pack.py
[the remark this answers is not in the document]
dokshina: If we start to install Tarantool from source, it will require a lot of changes as well as changing tests. As I now, building Tarantool from source isn't planned.

test/python/utils.py
olegrok: I suggest you to file an issue to fix it. If cartridge version is hardcoded - ok. For now it's ok

test/python/test_docker_pack.py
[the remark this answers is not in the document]
olegrok: You've already use 'tarantool --version' in `is_enterprise` function. It's an attempt to generalize the approach to versions management. I think it could be reused in several places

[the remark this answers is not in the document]
dokshina: `is_enterprise ` function checks Tarantool version on local machine, if Tarantool is not install there, test fill fail earlier.

Maybe it's better to use `tarantool --version` output instead of using `yum info tarantool` on version check.
But I don't want to change it now for one more reason - we are planning changes that allow to freeze Tarantool version in the result package or Docker image, so this check will be changed anyway (at least because Enterpise version would be tested too).

test/python/utils.py
dokshina: https://github.com/tarantool/cartridge-cli/issues/96

test/python/test_docker_pack.py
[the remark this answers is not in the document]
dokshina: Added a comment

olegrok:
Your patch is ok in general

I have several nits, but I hope we consider them in following iterations

PR #99, opened 2019-12-26
Support deprecated packing flow

dokshina:
Support deprecation flow and a new flow as well
On `cartridge pack` call check which flow is used, warn on deprecated
flow and check if only ine flow special filws are exists
Pass deprecated_build_flow flag to pack funtions
Deny using deprecated (`.cartridge.ignore` and `.cartridge.pre`) files for `pack docker`
Update README
Add deprecated_project fixture to test if it works correctly

### DIFF
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -796,6 +796,11 @@ end
 local PREBUILD_SCRIPT_NAME = 'cartridge.pre-build'
 local POSTBUILD_SCRIPT_NAME = 'cartridge.post-build'
 
+-- deprecated files
+
+local DEP_PREBUILD_SCRIPT_NAME = '.cartridge.pre'
+local DEP_IGNORE_FILE_NAME = '.cartridge.ignore'
+
 -- * --------------- Preinstall ---------------
 
 local CREATE_USER_SCRIPT = [[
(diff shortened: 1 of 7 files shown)


cartridge-cli.lua
```diff
@@ -1077,6 +1082,39 @@ local function generate_version_file(source_dir, dest_dir, app_name, app_version
     version_file:close()
 end
 
+local function pattern_form(pattern)
+    if pattern == '' or-- blank line
```
olegrok: `or--` add a space between, please

cartridge-cli.lua
```diff
@@ -1210,8 +1301,18 @@ local function build_application(dir)
     end
 
     -- post build
-    if fio.path.exists(fio.pathjoin(dir, POSTBUILD_SCRIPT_NAME)) then
-        run_hook(dir, POSTBUILD_SCRIPT_NAME)
+    if opts.deprecated_build_flow then
```
olegrok: Let's define a global variable for it to avoid ugly propagation of this flag via options

dokshina: I don't like using global variables, it's not a constant.
I think it's better to pass options explicitly.

cartridge-cli.lua
```diff
@@ -2150,7 +2259,7 @@ local function pack_docker(source_dir, _, name, release, version, opts)
     generate_version_file(source_dir, distribution_dir, name, version)
 
     local dockerfile_path = fio.pathjoin(tmpdir, 'Dockerfile')
-    construct_dockerfile(dockerfile_path, name, from)
+    construct_dockerfile(dockerfile_path, name, from, opts)
```
olegrok: This is a redundant change

PR #102, opened 2019-12-26
Test packing with wrong filemodes

### LINKED ISSUE #101: Test packing with wrong filemodes
Add a test to check if `cartridge pack` command fails if application contains files with wrong modes

dokshina:
Closes #101

### DIFF
--- a/test/python/test_pack.py
+++ b/test/python/test_pack.py
@@ -256,3 +256,27 @@ def test_packing_without_git(tmpdir):
         project_path,
     ]
     process = subprocess.run(cmd, cwd=tmpdir)
+
+
+def test_packing_with_wrong_filemodes(tmpdir):
+    project_name = 'test-project'
+
+    # create project
+    cmd = [
+        os.path.join(basepath, "cartridge"), "create",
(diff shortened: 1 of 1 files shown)


test/python/test_pack.py
```diff
@@ -256,3 +256,27 @@ def test_packing_without_git(tmpdir):
...
+    ]
+    process = subprocess.run(cmd, cwd=tmpdir)
+    assert process.returncode == 0, "Error during creating the project"
+    project_path = os.path.join(tmpdir, project_name)
+
+    # add file with invalid (700) mode
+    filepath = os.path.join(project_path, 'wrong-mode-file.lua')
+    with open(filepath, 'w') as f:
+        f.write("return 'Hi'")
+    os.chmod(filepath, 0o700)
```
olegrok: What's about chown "root" after?

dokshina: It doesn't matter, file owner is set to `root` on packing.

olegrok: OK, thanks for explanation

PR #105, opened 2020-01-07
Do 'git clean' on submodules during packing

knazarov:
Currently if an app has submodules, the 'git clean' stage that we do
in cartridge-cli doesn't touch them, which leads to garbage
accumulating in the resulting package. Sometimes it may even lead to
failed compilation (e.g. when using CMake due to leftover
CMakeCache.txt).

This patch adds recursive cleaning of submodules during packing.

### DIFF
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -1254,6 +1254,7 @@ local function form_distribution_dir(source_dir, dest_dir)
         -- Clean up all files explicitly ignored by git, to not accidentally
         -- ship development snaps, xlogs or other garbage to production.
         call("cd %q && %s clean -f -d -X", dest_dir, git)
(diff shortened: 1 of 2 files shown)


cartridge-cli.lua
```diff
@@ -1254,6 +1254,7 @@ local function form_distribution_dir(source_dir, dest_dir)
         -- Clean up all files explicitly ignored by git, to not accidentally
         -- ship development snaps, xlogs or other garbage to production.
         call("cd %q && %s clean -f -d -X", dest_dir, git)
+        call("cd %q && %s submodule foreach --recursive %s clean -f -d -X", dest_dir, git, git)
```
dokshina: Please, add a comment on this line.

PR #110, opened 2020-01-13
Pack application in docker

### LINKED ISSUE #98: Pack application in docker
## Application build flow

Application build comprises these steps:

* `cartridge.pre-build` (or `.cartridge.pre`);
* `tarantoolctl rocks make`;
* `cartridge.post-build`;

And then, we perform some actions, described in the `cartridge-cli` source code:

* writing VERSION file (it contains Trantool, SDK, application and installed rocks modules versions);
* (cleanning up files mentioned in `.cartridge.ignore`)

## What's the problem?

There are two types of problem:

### For all distribution types except docker

Building rocks modules takes place on the local machine. As a result, the package contains executables specific for the OS where `cartridge pack` command was ran.

### For docker distribution type:

Building rocks is running in the container.
It solves the problem of wrong type of executables, but on the other hand, writing VERSION file and applying `.cartridge.ignore` can't be ran in the container.

## Possible decision

We discussed a few ways to solve this problems, and I suggest using one of these:

* Application build stages are running in the docker container. The main purpose of this action is to get application distribution directory with the rocks modules specific for the target platform. (For packing to `tar.gz` user should have an opportunity to specify target system)

* Then, application files with installed rocks modules are copied on the local machine to perform some actions like applying `.cartridge.ignore` and writing installed rocks versions to the VERSION file.

* Next, `systemd` and `tmpfiles` directories are formed for `rpm` and `deb`.

* Finally, application files (with rocks modules and other staff) are packed in the archive or copied to the resulting image.

This approach solves the problem of wrong platform-specific rocks.
Forming distribution dir (`/usr/share/tarantool/<appname>`) is independent of distribution type, and all `cartridge-cli` specific actions are performed on the local machine.
Docker is used only for transforming current directory contents to the directory contents required for the target platform.

It will also help us to implement `runtime.txt` (see #88) functionality in the future.
It also will make possible to use deprecated build files (`.cartridge.pre` + `.cartridge.ignore`) with `pack docker` command.

dokshina:
This is the main part of #98

### DIFF
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -1249,15 +1249,14 @@ Depends: ${deps}
 
 -- * ------------------- Tmpfiles --------------------
 
-local TMPFILES_CONFIG = [[
-d /var/run/tarantool 0755 tarantool tarantool
-]]
+local TMPFILES_CONFIG = 'd /var/run/tarantool 0755 tarantool tarantool'
 
 -- * ------------------- Dockerfile -------------------
 
-local DOCKERFILE_FROM_DEFAULT = 'FROM centos:8'
+local DOCKERFILE_FROM_DEFAULT = 'FROM centos:8\n'
 
-local DOCKERFILE_TAIL_TEMPLATE = [[
+local DOCKERFILE_PREPARE = [[
+### Prepare
 SHELL ["/bin/bash", "-c"]
 
 RUN yum install -y git gcc make cmake unzip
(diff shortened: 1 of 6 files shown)


cartridge-cli.lua
```diff
@@ -2438,6 +2452,8 @@ local function app_pack(args)
     else
         die("Unknown package type: %s", args.type)
     end
+
+    clean_build_dir(pack_state.build_dir)
```
olegrok: IMO if build_dir is a temporary dir, we should drop it via rmtree

cartridge-cli.lua
```diff
@@ -2438,6 +2452,8 @@ local function app_pack(args)
     else
         die("Unknown package type: %s", args.type)
```
olegrok: In case of an error, we do not clear the directory... But I suppose that we should clean up any build garbage in case of error

cartridge-cli.lua
```diff
@@ -1651,7 +1676,240 @@ local function run_hook(dir, filename)
...
+    local install_tarantool_dockerfile_part
+    if app_state.tarantool_is_enterprise then
+        local version_filepath = fio.pathjoin(get_tarantool_dir(), 'VERSION')
+        local version_filepath_content = read_file(version_filepath)
+
+        local sdk_version = string.match(version_filepath_content, 'TARANTOOL_SDK=(%S+)\n')
+        if sdk_version == nil then
+            return nil, string.format('Failed to get SDK version from %s file', version_filepath)
+        end
+        sdk_version = sdk_version:gsub('-macos', '')
```
olegrok: what? why? I think macos shouldn't be mention at all

cartridge-cli.lua
```diff
@@ -1651,7 +1676,240 @@ local function run_hook(dir, filename)
...
+    })
+    local ok, err = call(build_image_command)
+    if not ok then
+        return false, string.format('Failed to build image: %s', err)
+    end
+
+    info('Base image tagged as: %s', app_state.base_image_fullname)
+
+    -- Build application in the base image
+    info('Build application in %s', app_state.base_image_fullname)
```
olegrok: I think it's a copypaste error because of `app_state.base_image_fullname` in both places

cartridge-cli.lua
dokshina: When you develop locally on your macos, your TE VERSION file contains string like `TARANTOOL_SDK=2.2.1-65-g34831ed-macos`. This version is used to detect TE version that should be delivered in the result package, and it shouldn't contain `macos` suffix. In a future, we plan to allow user to specify Tarantool version for the result package, but now it's detected from the version used for development.

cartridge-cli.lua
dokshina: Yes, but I'm not sure that it's bad to show both these messages.

olegrok:
Please, provide a unit test that checks Dockerfile content.

It could be useful use "white-box" testing and clearly understandable what was changed/broken/added

test/python/test_pack.py
```diff
@@ -407,3 +419,23 @@ def test_packing_without_path_specifying(project_without_dependencies, tmpdir):
 
     filepath = find_archive(project.path, project.name, 'rpm')
     assert filepath is not None,  'Package not found'
+
+
+@pytest.mark.parametrize('pack_format', ['rpm', 'deb', 'tgz', 'docker'])
+def test_build_in_docker_download_token_for_ee(project_without_dependencies, pack_format, tmpdir):
```
olegrok: `_without_download_token_for_ee`?

cartridge-cli.lua
```diff
@@ -1249,15 +1249,14 @@ Depends: ${deps}
...
 
-local TMPFILES_CONFIG = [[
-d /var/run/tarantool 0755 tarantool tarantool
-]]
+local TMPFILES_CONFIG = 'd /var/run/tarantool 0755 tarantool tarantool'
 
 -- * ------------------- Dockerfile -------------------
 
-local DOCKERFILE_FROM_DEFAULT = 'FROM centos:8'
+local DOCKERFILE_FROM_DEFAULT = 'FROM centos:8\n'
```
olegrok: Please, consider choosing base image, some users could deploy their projects on centos 7 then it should be more native for them to use centos:7 as base image

cartridge-cli.lua
```diff
@@ -1270,64 +1269,89 @@ RUN groupadd -r tarantool \
...
-    TARANTOOL_PID_FILE=/var/run/tarantool/${name}.${instance_name}.pid \
-    TARANTOOL_CONSOLE_SOCK=/var/run/tarantool/${name}.${instance_name}.control \
-    tarantool ${dir}/init.lua
 ]]
 
 local DOCKER_INSTALL_OPENSOURCE_TARANTOOL_TEMPLATE = [[
-# install opensource Tarantool
+### Install opensource Tarantool
 RUN curl -s \
         https://packagecloud.io/install/repositories/tarantool/${tarantool_repo_version}/script.rpm.sh | bash \
```
olegrok: Hmm... But in case 2.1 tarantool_repo_version = 2x (https://packagecloud.io/install/repositories/tarantool/2x/script.rpm.sh)

test/python/test_pack.py
dokshina: Good, thanks!

cartridge-cli.lua
dokshina: The result package will be the same anyway.
I sure that we need to support building on different platforms, but it will be done in the separated PR, because this change is very-very big and difficult.

Now, there is only one problem with using centos - `makedir` binary abspath in `ExecStartPre` section of unit file will be specific for centos, and it wouldn't work on some debian distributions. But it can be fixed by using custom `unit-template` and `instantiated-unit-template`.

cartridge-cli.lua
dokshina: OMG, thanks!

I will fix it in a separated patch.

dokshina: See #146

cartridge-cli.lua
```diff
@@ -968,6 +968,21 @@ local function tarantool_is_enterprise()
...
 
+local function get_sdk_version()
+    assert(tarantool_is_enterprise())
+
+    local version_filepath = fio.pathjoin(get_tarantool_dir(), 'VERSION')
+    local version_filepath_content = read_file(version_filepath)
+
+    local sdk_version = string.match(version_filepath_content, 'TARANTOOL_SDK=(%S+)\n')
+    if sdk_version == nil then
+        die('Failed to get SDK version from %s file', version_filepath)
```
olegrok: Why not `nil, err`? We should respect error handling convention

dokshina: This function is called outside of pack handler, this is the part of setting up `app_state`.

test/unit/dockerfile_constructors.lua
```diff
@@ -0,0 +1,225 @@
...
+    local expected_dockerfile = table.concat({
+        dockerfile_base_layers,
+        get_prepare_layers(),
+        get_install_tarantool_opensource_layers('2x'),
+        get_copy_code_layers(app_name),
+        get_dockerfile_runtime_layers(app_name),
+    }, '\n')
+
+    assert_lines_are_equal(constructor(), expected_dockerfile)
+
```
olegrok: redundant empty line

dokshina: Thanks, fixed.

PR #112, opened 2020-01-16
Fix RPM header for centos:8

dokshina:
The new (4.14.2) RPM version used in centos:8 image requires
PAYLOADDIGEST tag in package header.
This patch adds PAYLOADDIGEST and PAYLOADDIGESTALGO tags to package
header to prevent package signature check fail

### DIFF
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -1556,6 +1556,8 @@ local HEADERIMMUTABLE=63
 -- But I didn't find documentation for some tags (PAYLOADDIGEST, PAYLOADDIGESTALGO),
 -- so, I got these values from the rpm repo:
 -- - https://github.com/rpm-software-management/rpm/blob/master/lib/rpmtag.h
+-- payload digest explanation can be found here:
+-- - https://github.com/rpm-software-management/rpm/issues/163
 --
 local SIGNATURE_TAG_TABLE = {
     SIG_SIZE = 1000,
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -1548,6 +1548,15 @@ local HEADERIMMUTABLE=63
 -- There are way more tags in the spec than what I've included here
 -- both for signature header and regular header. Most of them are
 -- optional.
+--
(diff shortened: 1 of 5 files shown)


cartridge-cli.lua
```diff
@@ -1985,7 +2002,9 @@ local function pack_rpm(source_dir, dest_dir, name, release, version, opts)
         {'FILEDIGESTS', 'STRING_ARRAY', fileinfo.filedigests},
         {'FILELINKTOS', 'STRING_ARRAY', fileinfo.filelinktos},
         {'RPMVERSION', 'STRING', '4.11.3'},
```
olegrok: This feature was implemented in https://github.com/rpm-software-management/rpm/commit/91aa0786cf3b2e34de01c586427952de6d0d9b40 "4.14.0-rc1"

I suggest you to bump this version... I'm not sure that it can't cause any conflicts

cartridge-cli.lua
```diff
@@ -1538,6 +1538,15 @@ local HEADERIMMUTABLE=63
...
 -- both for signature header and regular header. Most of them are
 -- optional.
+--
+-- Explanation and values for most of this tags can be found in documentation:
+-- - http://ftp.rpm.org/max-rpm/s1-rpm-file-format-rpm-file-format.html
+-- - https://docs.fedoraproject.org/ro/Fedora_Draft_Documentation/0.1/html/RPM_Guide/ch-package-structure.html
+--
+-- But I didn't find documentation for some tags (PAYLOADDIGEST, PAYLOADDIGESTALGO),
+-- so, I got these values from the rpm repo:
+-- - https://github.com/rpm-software-management/rpm/blob/master/lib/rpmtag.h
```
olegrok: And may be this PR? https://github.com/rpm-software-management/rpm/issues/163

knazarov: I propose not to. It's not mandatory, and it will complicate packing.

cartridge-cli.lua
dokshina: This flag is optional, so I suggest removing it from header. It just confuses.

PR #114, opened 2020-01-16
pack: extend custom version with git commit

### LINKED ISSUE #113: Extend "version" with git commit
Now: `cartridge pack rpm /opt/project --version 1.2.3-264-g8bce594e` leads to `project-1.2.3-0.rpm`

Expected: `project-1.2.3-264-g8bce594e.rpm`

olegrok:
pack: extend custom version with git commit

This patch allows to pass version as "major.minor.patch-count-hash"
    
Now user can pass `cartridge pack rpm --version 1.2.3-1-g8bce594e` and get rpm with name `project-1.2.3-1-g8bce594e.rpm`
    
Also hash is an optional argument and could be omitted: `cartridge pack rpm --version 1.2.3-1` -> `project-1.2.3-1.rpm`
    
Closes #113

### DIFF
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -692,6 +692,7 @@ local function normalize_version(str)
     local patterns = {
         "(%d+)%.(%d+)%.(%d+)-(%d+)-(%g+)",
         "(%d+)%.(%d+)%.(%d+)-(%d+)",
+        "(%d+)%.(%d+)%.(%d+)-(%g+)",
         "(%d+)%.(%d+)%.(%d+)",
         "(%d+)%.(%d+)",
         "(%d+)"
@@ -700,9 +701,13 @@ local function normalize_version(str)
     for _, pattern in ipairs(patterns) do
         local major, minor, patch, count, hash = string.match(str, pattern)
         if major ~= nil then
-            local release = count or '0'
-            if hash ~= nil then
-                release = string.format('%s-%s', release, hash)
+            local release = '0'
+            if count ~= nil and hash ~= nil then
+                release = string.format('%s-%s', count, hash)
+            elseif count ~= nil then
+                release = tostring(count)
+            elseif hash ~= nil then
+                release = tostring(hash)
             end
             return format_version(major, minor, patch), release
         end
@@ -778,7 +783,7 @@ local function detect_name_release_version(source_dir, raw_name, raw_version)
     if raw_version then
         version, release = normalize_version(raw_version)
         if version == nil then
-            die("Passed version '%s' should be semantic (major.minor.patch[-commit])",
+            die("Passed version '%s' should be semantic (major.minor.patch[-count][-commit])",
                 raw_version)
         end
         release = release or '0'
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -801,7 +801,7 @@ local function detect_name_release_version(source_dir, raw_name, raw_version)
         die("Application must have `init.lua` in its root directory")
     end
 
-    return name, release, version
+    return name, version, release
 end
 
 -- * ----------- Special filenames ------------
@@ -1063,7 +1063,7 @@ local function get_rock_versions(project_dir)
     return dependencies
 end
 
-local function generate_version_file(source_dir, dest_dir, app_name, app_version)
+local function generate_version_file(dest_dir, app_name, app_version, app_release)
     local version_file, _ = fio.open(
         fio.pathjoin(dest_dir, 'VERSION'),
         {'O_TRUNC', 'O_WRONLY', 'O_CREAT'},
@@ -1089,8 +1089,7 @@ local function generate_version_file(source_dir, dest_dir, app_name, app_version
         version_file:write(string.format('TARANTOOL=%s\n', _TARANTOOL))
     end
 
-    local _, _, app_commit = detect_git_version(source_dir)
-    version_file:write(string.format("%s=%s-%s\n", app_name, app_version, app_commit or ''))
+    version_file:write(string.format("%s=%s-%s\n", app_name, app_version, app_release))
 
     local rocks_versions, err = get_rock_versions(dest_dir)
     if rocks_versions == nil then
@@ -1409,7 +1408,7 @@ end
 
 -- * ---------------- TAR.GZ packing ----------------
 
-local function pack_tgz(source_dir, dest_dir, name, release, version)
+local function pack_tgz(source_dir, dest_dir, name, version, release)
     local tgz_file_name = string.format("%s-%s-%s.tar.gz", name, version, release)
     tgz_file_name = fio.pathjoin(dest_dir, tgz_file_name)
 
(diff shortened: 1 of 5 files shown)


test/python/test_pack.py
```diff
@@ -263,7 +263,18 @@ def test_packing_without_git(tmpdir):
         "--version", "0.1.0",
         project_path,
     ]
-    process = subprocess.run(cmd, cwd=tmpdir)
```
dokshina: I just realized that I missed `process.returncode == 0` assertion here.

The main idea of this test is to check that `cartridge pack` doesn't fail if application is not a git repository.

test/python/test_pack.py
```diff
@@ -263,7 +263,18 @@ def test_packing_without_git(tmpdir):
...
+    subprocess.run(cmd, cwd=tmpdir)
+    assert 'test-project-0.1.0-0.rpm' in os.listdir(tmpdir)
+
+    # pass extended version
```
dokshina: Why didn't you add a new test for this feature?

BTW, maybe we should add `project_without_dependencies` or smth like this to use it in this kind of tests.

cartridge-cli.lua
```diff
@@ -682,19 +682,25 @@ end
...
             patch = patch or '0'
-
-            return {major, minor, patch}
+            if commit_num ~= nil then
+                commit = tostring(commit_num)
+            end
+            if commit_hash ~= nil then
+                commit = commit .. '-' .. tostring(commit_hash)
+            end
+            return {major, minor, patch}, commit
```
dokshina: Maybe it's time to get rid of this horrible multi-return.
We already discussed that normalized version can be a string and I hope it will.

BTW, in the code `release` name is used to save `commit` value, it confuses a little.

cartridge-cli.lua
```diff
@@ -682,19 +682,25 @@ end
...
         if major ~= nil then
             minor = minor or '0'
             patch = patch or '0'
-
-            return {major, minor, patch}
+            if commit_num ~= nil then
+                commit = tostring(commit_num)
+            end
+            if commit_hash ~= nil then
+                commit = commit .. '-' .. tostring(commit_hash)
```
dokshina: Why don't you use `string.format()`?

olegrok: Fixed

cartridge-cli.lua
olegrok: Replace `{major, minor, patch` to one "major.minor.patch" string

Change "commit" to "release"

test/python/test_pack.py
olegrok: Now, I check "process.returncode == 0"

But I don't see any matter to rename this test or split into several. It's not much and save time because "prepare step" is executed only once

test/python/test_pack.py
olegrok: See explanation here https://github.com/tarantool/cartridge-cli/pull/114/files#r367650797

cartridge-cli.lua
olegrok: Also may be we can return simply "version-release"... But this code is quite fragile, I don't want to do it in this patch

test/python/test_pack.py
dokshina: Your test is not about “packing application without git”. The “prepare step” just removes dependencies from project, it can be a fixture with session scope.

cartridge-cli.lua
```diff
@@ -680,21 +681,32 @@ end
...
-            patch = patch or '0'
-
-            return {major, minor, patch}
+            local release
+            if commit_num ~= nil then
+                release = tostring(commit_num)
+            end
+            if commit_hash ~= nil then
+                release = string.format('%s-%s', release, commit_hash)
+            end
```
dokshina: Maybe release should be set to 0 by deafault?

test/python/test_pack.py
olegrok: Ok

Moved to `test_packing_with_version`

cartridge-cli.lua
olegrok: fixed

cartridge-cli.lua
```diff
@@ -723,16 +733,15 @@ local function detect_version(source_dir)
         return nil
```
dokshina: I sure that this function should be renamed to `detect_git_version`, it detects version only from git.

olegrok: Done

cartridge-cli.lua
dokshina: If user specifies `--version 0.1.0-g8bce594e`, he get `<name>-0.1.0-0-g8bce594e.<type>` package, not  `<name>-0.1.0-g8bce594e.<type>` as he expected.

Maybe it would be better to do smth like this?

```lua

            local release = '0'

            if count ~= nil and hash ~= nil then
                release = string.format('%s-%s', count, hash)
            elseif count ~= nil then
                release = tostring(count)
            elseif hash ~= nil then
                release = tostring(hash)
            end

            return format_version(major, minor, patch), release
```

test/python/test_pack.py
```diff
@@ -264,6 +243,31 @@ def test_packing_without_git(tmpdir):
         project_path,
     ]
     process = subprocess.run(cmd, cwd=tmpdir)
+    assert process.returncode == 0
+    assert 'empty-project-0.1.0-0.rpm' in os.listdir(tmpdir)
```
dokshina:
```python
project_name = project_without_dependencies['name']
assert '{}-0.1.0-0.rpm'.format(project_name) in os.listdir(tmpdir)
```

Just to prevent unexpected test failure on changing `project_without_dependencies` name.

cartridge-cli.lua
```diff
@@ -767,21 +776,20 @@ local function detect_name_release_version(source_dir, raw_name, raw_version)
     end
 
     if raw_version then
-        if not normalize_version(raw_version) then
-            die("Passed version '%s' should be semantic (major.minor.patch)",
+        version, release = normalize_version(raw_version)
+        if version == nil then
+            die("Passed version '%s' should be semantic (major.minor.patch[-commit])",
```
dokshina: Maybe `"Passed version '%s' should be semantic (major.minor.patch[-count][-commit])"`?

cartridge-cli.lua
```diff
@@ -1076,8 +1084,8 @@ local function generate_version_file(source_dir, dest_dir, app_name, app_version
         version_file:write(string.format('TARANTOOL=%s\n', _TARANTOOL))
     end
 
-    local _, _, app_commit = detect_version(source_dir)
-    version_file:write(string.format("%s=%s-%s\n", app_name, table.concat(app_version, '.'), app_commit or ''))
+    local _, _, app_commit = detect_git_version(source_dir)
```
dokshina: I think it shouldn't be called here.

We already detected git version and saved that.

I suggest 

```lua
function generate_version_file(source_dir, dest_dir, app_name, app_version, app_release)
...
version_file:write(string.format("%s=%s-%s\n", app_name, app_version, app_release)
...
end
...

```

```

cartridge-cli.lua
olegrok: it could be quite strange, but ok, let's support it

test/python/test_pack.py
olegrok: Yes, thanks, fixes

cartridge-cli.lua
olegrok: Fixed

cartridge-cli.lua
olegrok: I hope I nothing break

Fixed

PR #115, opened 2020-01-17
Store common packing parameters in global variable

dokshina:
This is the first part of impementing #98

Since application packing has more and more parameters (such as `name`, `version` or `deprecated_flow`), their propagation looks uglier.

@olegrok suggested storing all these parameters in the global variable, and we implemented this for `deprecated_flow` flag. But packing application in docker requires additional parameters, so I decided to store all common packing flow parameters in the global variable `pack_state`. Some parameters, specific for distribution type, are still propagated by options.

### DIFF
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -1119,9 +1119,9 @@ local function generate_version_file(distribution_dir)
     end
 
     version_file:write(string.format(
-        "%s=%s-%s\n", pack_state.name,
-        pack_state.version,
-        pack_state.release
+        "%s=%s\n",
+        pack_state.name,
+        pack_state.version_release
     ))
 
     local rocks_versions, err = get_rock_versions(distribution_dir)
@@ -1451,10 +1451,9 @@ end
 
 local function pack_tgz()
     local tgz_file_name = string.format(
-        "%s-%s-%s.tar.gz",
+        "%s-%s.tar.gz",
         pack_state.name,
-        pack_state.version,
-        pack_state.release
+        pack_state.version_release
     )
     tgz_file_name = fio.pathjoin(pack_state.dest_dir, tgz_file_name)
 
@@ -1517,17 +1516,16 @@ local function pack_rock()
     if rockspec then
         content = read_file(fio.pathjoin(distribution_dir, rockspec))
         content = string.gsub(content, "(.-version%s-=%s-['\"])(.-)(['\"].*)",
-                '%1' .. string.format('%s-%s', pack_state.version, pack_state.release) .. '%3')
+                '%1' .. pack_state.version_release .. '%3')
         if not content then
             die('Rockspec %s is not valid! Version not found!')
         end
     end
 
     local name_of_rockspec = string.format(
-        '%s-%s-%s.rockspec',
+        '%s-%s.rockspec',
         pack_state.name,
-        pack_state.version,
-        pack_state.release
+        pack_state.version_release
     )
 
     local new_rockspec = fio.pathjoin(distribution_dir, name_of_rockspec)
@@ -1537,10 +1535,9 @@ local function pack_rock()
     fio.chdir(distribution_dir)
 
     local rock_filename = string.format(
-        '%s-%s-%s.*.rock',
+        '%s-%s.*.rock',
         pack_state.name,
-        pack_state.version,
-        pack_state.release
+        pack_state.version_release
     )
 
     call_or_die('tarantoolctl rocks pack %s ', new_rockspec)
@@ -2033,10 +2030,9 @@ local function pack_rpm(opts)
     local rpm_file_name = fio.pathjoin(
         pack_state.dest_dir,
         string.format(
-            "%s-%s-%s.rpm",
+            "%s-%s.rpm",
             pack_state.name,
-            pack_state.version,
-            pack_state.release
+            pack_state.version_release
         )
     )
 
@@ -2149,14 +2145,14 @@ end
 -- data.tar.xz    : package files
 -- control.tar.xz : control files (control, preinst etc.)
 --
-local function form_deb_control_dir(dest_dir, name, version, release)
+local function form_deb_control_dir(dest_dir, name, version)
     fio.mktree(dest_dir)
 
     -- control
     local control_filepath = fio.pathjoin(dest_dir, 'control')
     local control_params = {
         name = name,
-        version = ('%s-%s'):format(version, release),
+        version = version,
         maintainer = 'Tarantool Cartridge Developer',
         arch = 'all',
         desc = ('Tarantool Cartridge app %s'):format(name),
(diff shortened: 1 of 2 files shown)


cartridge-cli.lua
```diff
@@ -1050,9 +1051,11 @@ local function get_rock_versions(project_dir)
     return dependencies
 end
 
-local function generate_version_file(source_dir, dest_dir, app_name, app_version)
+local function generate_version_file(distribution_dir)
+    info('Generate VERSION file')
+
     local version_file, _ = fio.open(
-        fio.pathjoin(dest_dir, 'VERSION'),
+        fio.pathjoin(distribution_dir, 'VERSION'),
```
olegrok: destination? or distribution?

dokshina: In this project name `distribution_dir` is used for directory where application files are placed (`/usr/share/tarantool/<appname>` contents).

I use `dest_dir` (destination directory) name to specify path where the result package should be placed.

cartridge-cli.lua
```diff
@@ -1061,32 +1064,37 @@ local function generate_version_file(source_dir, dest_dir, app_name, app_version
...
         version_file:write(string.format('TARANTOOL=%s\n', _TARANTOOL))
     end
 
-    local _, _, app_commit = detect_version(source_dir)
-    version_file:write(string.format("%s=%s-%s\n", app_name, table.concat(app_version, '.'), app_commit or ''))
+    local _, _, app_commit = detect_version(pack_state.path)
+    version_file:write(string.format(
+        "%s=%s-%s\n",
+        pack_state.name,
+        table.concat(pack_state.version, '.'),
```
olegrok: `table.concat`? Fix this rebase artefact please

cartridge-cli.lua
```diff
@@ -1396,11 +1407,12 @@ end
...
 -- * ---------------- TAR.GZ packing ----------------
 
-local function pack_tgz(source_dir, dest_dir, name, release, version)
+local function pack_tgz()
     local tgz_file_name = string.format(
         "%s-%s-%s.tar.gz",
-        name, table.concat(version, '.'), release)
-    tgz_file_name = fio.pathjoin(dest_dir, tgz_file_name)
+        pack_state.name,
+        table.concat(pack_state.version, '.'), pack_state.release)
```
olegrok: table.concat

cartridge-cli.lua
```diff
@@ -1434,54 +1446,58 @@ end
...
 
-    local rockspec = find_rockspec(destdir)
+    local rockspec = find_rockspec(distribution_dir)
     local content = ''
     if rockspec then
-        content = read_file(fio.pathjoin(destdir, rockspec))
+        content = read_file(fio.pathjoin(distribution_dir, rockspec))
         content = string.gsub(content, "(.-version%s-=%s-['\"])(.-)(['\"].*)",
-                '%1' .. string.format('%s-%s', table.concat(version, '.'), release) .. '%3')
+                '%1' .. string.format('%s-%s', table.concat(pack_state.version, '.'), pack_state.release) .. '%3')
```
olegrok: table.concat

cartridge-cli.lua
```diff
@@ -1434,54 +1446,58 @@ end
...
             die('Rockspec %s is not valid! Version not found!')
         end
     end
 
-    local name_of_rockspec = string.format('%s-%s-%s.rockspec', name, table.concat(version, '.'),
-                    release)
+    local name_of_rockspec = string.format(
+        '%s-%s-%s.rockspec',
+        pack_state.name,
+        table.concat(pack_state.version, '.'),
```
olegrok: table.concat

cartridge-cli.lua
```diff
@@ -2322,6 +2343,21 @@ end
 
 local function app_pack(args)
     local name, release, version = detect_name_release_version(args.path, args.name, args.version)
+
+    -- collect general application info
```
olegrok: This is quite inconsistent with `app_create`...

I think that logic of this functions should be uniform

cartridge-cli.lua
olegrok: Okay

cartridge-cli.lua
dokshina: `app_create` function is much more simple. It calls the only one function `instantiate_template` and pass in this function only one application parameter - name.

I think, that we shouldn't complicate `app_create` function.

cartridge-cli.lua
```diff
@@ -340,6 +340,7 @@ local function read_file(path)
     while true do
         local val = file:read(1024)
         if val == nil then
+            pcall(function() file:close() end)
```
olegrok: pcall(file.close, file)

cartridge-cli.lua
```diff
@@ -1253,8 +1261,8 @@ local function check_filemodes(dir)
     end
 end
 
-local function form_distribution_dir(source_dir, dest_dir)
-    assert(fio.copytree(source_dir, dest_dir))
+local function form_distribution_dir(dest_dir)
+    assert(fio.copytree(pack_state.path, dest_dir))
```
olegrok: I think `assert` is too rude

cartridge-cli.lua
dokshina: I think it looks ugly and confuses a little.

cartridge-cli.lua
dokshina: I agree with you.

cartridge-cli.lua
olegrok: Ok

cartridge-cli.lua
olegrok: Ok, for console application - ok.

If it is in context of server app, I think I will disagree with you

cartridge-cli.lua
```diff
@@ -1066,31 +1070,35 @@ local function generate_version_file(dest_dir, app_name, app_version, app_releas
...
-            version_file:write(fio.open(tnt_version):read())
+            version_file:write(read_file(tnt_version))
         end
     else
         -- write TARANTOOL version
         version_file:write(string.format('TARANTOOL=%s\n', _TARANTOOL))
     end
 
-    version_file:write(string.format("%s=%s-%s\n", app_name, app_version, app_release))
+    version_file:write(string.format(
```
olegrok: May be we should check that our data were successfully written to file?

Don't ignore errors

cartridge-cli.lua
```diff
@@ -2351,7 +2386,23 @@ local function check_if_deprecated_build_flow_is_ised(source_dir)
...
+    pack_state.version_release = string.format('%s-%s', version, release)
+
+    -- collect pack-specific application info
+    pack_state.dest_dir = fio.abspath('.')
+    pack_state.from = args.from
+    pack_state.download_token = args.download_token
+    pack_state.docker_build_args = args.docker_build_args
+    pack_state.deprecated_flow = check_if_deprecated_build_flow_is_ised()
+    pack_state.tarantool_is_enterprise = tarantool_is_enterprise()
+
```
olegrok: Let's add asserts here to be sure that no one of these parameters is empty. We will exploded before could start some destructive actions

PR #116, opened 2020-01-17
Allow to specify build directory

dokshina:
This is the second part of impementing #98.

* By default, temporary directory for application building is created in
  `~/.cartridge/tmp`
* User can specify custom build directory for project in `CARTRIDGE_BUILDDIR`
  environment variable. If this directory doesn't exists, it will be created, used
  for building the application and then removed.
* If directory already exists, `CARTRIDGE_BUILDDIR/build.cartridge` repository 
  will be used for build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,16 +112,17 @@ the application version is detected from `git describe`.
 
 #### Build directory
 
-By default, application build is performed in the directory `<app_path>/build.cartridge`, so the
+By default, application build is performed in the temporarily directory in the
+`~/.cartridge/tmp/`, so the
 packaging process doesn't affect the contents of your application directory.
 
-You can specify custom build directory for your project in `TARANTOOL_BUILDDIR`
+You can specify custom build directory for your project in `CARTRIDGE_BUILDDIR`
 environment variable. If this directory doesn't exists, it will be created, used
 for building the application and then removed.
 **Note**, that specified directory can't be project subdirectory.
 
-If you specify existent directory in `TARANTOOL_BUILDDIR` environment variable,
-`TARANTOOL_BUILDDIR/build.cartridge` repository will be used for build and then
+If you specify existent directory in `CARTRIDGE_BUILDDIR` environment variable,
+`CARTRIDGE_BUILDDIR/build.cartridge` repository will be used for build and then
 removed. This directory will be cleaned before building application.
 
 #### General packing flow and options
--- a/README.md
+++ b/README.md
@@ -110,10 +110,21 @@ The result will be named as `<name>-<version>.<type>`.
 By default, the application name is detected from the rockspec, and
 the application version is detected from `git describe`.
 
-#### General packing flow and options
+#### Build directory
 
-A package is first created in a temporarily directory, so the packaging process
-doesn't affect the contents of your application directory.
+By default, application build is performed in the directory `<app_path>/build.cartridge`, so the
+packaging process doesn't affect the contents of your application directory.
+
+You can specify custom build directory for your project in `TARANTOOL_BUILDDIR`
+environment variable. If this directory doesn't exists, it will be created, used
+for building the application and then removed.
+**Note**, that specified directory can't be project subdirectory.
+
+If you specify existent directory in `TARANTOOL_BUILDDIR` environment variable,
+`TARANTOOL_BUILDDIR/build.cartridge` repository will be used for build and then
+removed. This directory will be cleaned before building application.
+
+#### General packing flow and options
 
 A package build comprises these steps:
 
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -907,7 +907,10 @@ local DEP_IGNORE_FILE_NAME = '.cartridge.ignore'
 
 -- build directory
 
-local BUILD_DIRECTORY_NAME = 'build.cartridge'
+local HOME_DIR = os.getenv('HOME') or '/home'
+local DEFAULT_BUILD_DIRECTORY_NAME = 'build.cartridge'
+local CARTRIDGE_TMP_PATH = fio.pathjoin(HOME_DIR, '.cartridge/tmp')
+local BUILD_DIRECTORY_NAME_TEMPLATE = 'cartridge-build-%s'
 
 -- * --------------- Preinstall ---------------
 
@@ -1351,26 +1354,8 @@ local function check_filemodes(dir)
     end
 end
 
-local function copy_app_files(dest_dir)
-    make_tree(dest_dir)
-
-    for _, name in ipairs(listdir(pack_state.path)) do
-        local source_path = fio.pathjoin(pack_state.path, name)
-        local dest_path = fio.pathjoin(dest_dir, name)
-
-        if source_path ~= pack_state.build_dir then
-            if fio.path.is_dir(source_path) then
-                make_tree(dest_path)
-                copytree(source_path, dest_path)
-            else
-                copyfile(source_path, dest_path)
-            end
-        end
-    end
-end
-
 local function form_distribution_dir(dest_dir)
-    copy_app_files(dest_dir)
+    copytree(pack_state.path, dest_dir)
 
     local rocks_dir = fio.pathjoin(dest_dir, '.rocks')
     if fio.path.exists(rocks_dir) then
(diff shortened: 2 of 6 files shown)


cartridge-cli.lua
```diff
@@ -2341,6 +2334,52 @@ local function check_if_deprecated_build_flow_is_ised()
...
+    -- By default, application is built in the tmpdir.
+    -- User can specify build directory in TARANTOOL_BUILDDIR env variable.
+    -- There are two cases:
+    -- - specified directory doesn't exists: we just create it and remove after the build
+    -- - directory already exists:
+    --   - ${TARANTOOL_BUILDDIR}/cartridge.build will be the build directory
+    --   - sub-directory cartridge.build is (re)created and used for application build
+    --   - after the build, ${TARANTOOL_BUILDDIR}/cartridge.build  is removed
+
+    local specified_build_dir = os.getenv('TARANTOOL_BUILDDIR')
```
olegrok: Let's validate this path

dokshina: I validate if it's a directory if the path exists.

What kind of validation did you mean?

olegrok: My bad... Yes, that's ok

cartridge-cli.lua
```diff
@@ -2434,6 +2474,9 @@ local function app_pack(args)
     else
         die("Unknown package type: %s", args.type)
     end
+
+    -- clean build directory
+    fio.rmtree(pack_state.build_dir)
```
olegrok: In case if application die, we don't remove build_dir

I think `pack_*` functions should respect `nil, err` convention. And we will have one entry point where can process an error

olegrok: You can create a table with "handlers" e.g.
```lua
local handlers = {
   ['docker'] = pack_docker,
   ['rpm'] = pack_rpm,
   ...
}

local pack = handlers[mode]
if pack == nil then
    die()
end

local ok, err = pack(args)
fio.rmtree(builddir)
if not ok then
    die(err)
else
   info('Success!')
end
```

dokshina: I like your idea, but I sure that it should be done in separated patch.

Moreover, I suggest adding named constants for distribution types to use them in code and error messages.

README.md
```diff
@@ -110,10 +110,15 @@ The result will be named as `<name>-<version>.<type>`.
 By default, the application name is detected from the rockspec, and
 the application version is detected from `git describe`.
 
-#### General packing flow and options
+#### Build directory
+
+By default, application build is performed in the temporarily directory, so the packaging process doesn't affect the contents of your application directory.
```
olegrok: temporarily -> temporary?

README.md
```diff
@@ -110,10 +110,15 @@ The result will be named as `<name>-<version>.<type>`.
...
+
+You can specify custom build directory for your project in `TARANTOOL_BUILDDIR` environment variable. If this directory doesn't exists, it will be created, used for building the application and then removed.
 
-A package is first created in a temporarily directory, so the packaging process
-doesn't affect the contents of your application directory.
+If you specify existent directory in `TARANTOOL_BUILDDIR` environment variable, `TARANTOOL_BUILDDIR/cartridge.build` repository will be used for build and then removed. If `TARANTOOL_BUILDDIR/cartridge.build` is already exists, it will be cleaned before building application.
```
olegrok: Please, split long paragraphs into several lines. Markdown supports it.

It could be quite hard to edit in future in some editors

cartridge-cli.lua
```diff
@@ -2385,6 +2378,52 @@ local function check_if_deprecated_build_flow_is_ised()
...
+    if not fio.path.is_dir(build_dir) then
+        die("Specified build directory is not a directory: %s", build_dir)
+    end
+
+    -- This little hack is used to prevent deletion user files
+    -- from specified build directory on cleanup.
+    -- Moreover, this subdirectory is defenitely clean,
+    -- so we wouldn't have any problems
+    build_dir = fio.pathjoin(build_dir, BUILD_SUBDIRECTORY_NAME)
+    fio.rmtree(build_dir)
```
olegrok: This may be extremely unexpected for the user - let's log this action. `Directory is already exists - clean it`

knazarov:
Temporary directory name should be auto-generated and be random. It shouldn't be located in the project directory, but at least in home dir.

cartridge-cli.lua
dokshina: See #127

cartridge-cli.lua
```diff
@@ -845,6 +905,13 @@ local POSTBUILD_SCRIPT_NAME = 'cartridge.post-build'
 local DEP_PREBUILD_SCRIPT_NAME = '.cartridge.pre'
 local DEP_IGNORE_FILE_NAME = '.cartridge.ignore'
 
+-- build directory
+
+local HOME_DIR = os.getenv('HOME') or '/home'
```
olegrok:
```bash
> env
...
HOME=/Users/o.babin
...
```

I think it should be `~/` instead of `/home`

test/python/test_pack.py
```diff
@@ -325,3 +326,32 @@ def test_rpm_checksig(rpm_archive):
...
+    env['CARTRIDGE_BUILDDIR'] = os.path.join(project.path, 'sub', 'sub', 'directory')
+    process = subprocess.run(cmd, cwd=tmpdir, env=env)
+    assert process.returncode == 1
+
+    # pass correct directory as a builddir
+    builddir = os.path.join(tmpdir, 'build')
+    env['CARTRIDGE_BUILDDIR'] = builddir
+    process_output = subprocess.check_output(cmd, cwd=tmpdir, env=env)
+    process_output = process_output.decode()
+    assert re.search(r'[Bb]uild directory .+{}'.format(builddir), process_output) is not None
```
olegrok: `[Bb]`?

cartridge-cli.lua
```diff
@@ -2427,15 +2487,15 @@ local function pack_docker(opts)
     info('Resulting image tagged as: %s', image_fullname)
 end
 
-local function check_if_deprecated_build_flow_is_ised()
+local function check_if_deprecated_build_flow_is_ised(app_path)
```
olegrok: Why do you replace previous behaviour?

cartridge-cli.lua
```diff
@@ -1389,10 +1448,22 @@ local function build_application(dir)
...
-        remove_by_path(fio.pathjoin(dir, DEP_IGNORE_FILE_NAME))
-        remove_by_path(fio.pathjoin(dir, DEP_PREBUILD_SCRIPT_NAME))
-        info('Remove .git directory')
-        remove_by_path(fio.pathjoin(dir, '.git'))
+
+        -- remove special files
+        for _, filename in ipairs({DEP_IGNORE_FILE_NAME, DEP_PREBUILD_SCRIPT_NAME}) do
+            local filepath = fio.pathjoin(dir, filename)
+            if fio.path.exists(filepath) then
+                info('Remove %s', filename)
```
olegrok: Could you please provide an example of log of packaging some simple module?

I suppose that such verbosity could be redundant. And this messages should have "verbose" log level

cartridge-cli.lua
dokshina: `fio` thinks that `~` is a `~` directory in the local path. E.g. if you call `fio.abspath('~')` from `/path/to/me`, you get `/path/to/me/~`.

test/python/test_pack.py
dokshina: `Build` or `build`. I don't want to change this test on log message changing.

cartridge-cli.lua
dokshina: What?

cartridge-cli.lua
dokshina:
```

Comment (dokshina) [reply cartridge-cli.lua]: It's just two messages. I think it's better than deleting these files without any message.

File: cartridge-cli.lua (hunk 6 above)
Comment (olegrok) [reply cartridge-cli.lua]: looks like a fio bug

Comment (olegrok) [reply cartridge-cli.lua]: but ok. Let's use `env HOME`

File: cartridge-cli.lua (hunk 9 above)
Comment (olegrok) [reply cartridge-cli.lua]: `.../.cartridge/tmp/cartridge-build-8f7e3cefe056829c`

Why '...' at start?

In general ok

File: cartridge-cli.lua [hunk 10]
```diff
@@ -2480,10 +2597,18 @@ local function check_pack_state(state)
...
+    end
+
+    if not fio.path.is_dir(args.path) then
+        die("Specified path %s is not a directory", args.path)
+    end
+
     local name, version, release = detect_name_version_release(args.path, args.name, args.version)
 
     -- collect general application info
-    pack_state.path = args.path
```
olegrok: may be we are ready to

```lua
pack_state = {
     param1 = param1,
     ...
}
```
?

PR #117, opened 2020-01-18
refactoring: simplify detect_git_version function

olegrok:
After bed66e7234515eb06d192c4eff4cc04348022b03
cartridge-cli supports `git describe` version format.

Let's reuse `normalize_version` function and drop a piece of redundant code

### DIFF
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -731,22 +731,14 @@ local function detect_git_version(source_dir)
         return nil
     end
 
-    local version, release, commit = string.match(
-        string.strip(raw_version), "^(.*)-(%d+)-(%g+)$")
-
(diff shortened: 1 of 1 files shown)


cartridge-cli.lua
```diff
@@ -731,22 +731,14 @@ local function detect_git_version(source_dir)
...
     if version == nil then
-        return nil
-    end
-
-    local normalized_version = normalize_version(version)
-    if normalized_version == nil then
-
         warn("Detected version '%s' ignored, " ..
               "because it doesn't look like proper " ..
               "version (major.minor.patch)", version)
```
dokshina: It should look like `(major.minor.patch[-count][-commit])`.

olegrok: Fixed

PR #119, opened 2020-01-21
Refactor tests

dokshina:


test/python/project.py
```diff
@@ -0,0 +1,235 @@
...
+
+    # add custom-project to version_file_keys
+    project.version_file_keys.add(SUBMODULE_NAME)
+
+
+ignored_data = [
+    {'dir': '',                  'file': 'ignored.txt'},
+    {'dir': '',                  'file': 'asterisk'},
+    {'dir': '',                  'file': 'ignored.lua'},
+    {'dir': '',                  'file':  'ignored_by.format'},
```
olegrok: Redundant space between `:` - `'file':  'ignored_by.format'}`

test/python/test_docker_pack.py
```diff
@@ -244,7 +237,7 @@ def test_docker_e2e(project, docker_image, tmpdir, docker_client):
         image_name,
         environment=environment,
         ports={'8082': '8082'},
-        name='{}-instance-1'.format(project['name']),
```
olegrok: Why do you remove a `project` from this test?

.gitlab-ci.yml
```diff
@@ -42,6 +42,8 @@ variables:
 prebuild:
   extends: .dind
   stage: prebuild
+  variables:
+    DOCKER_BUILDKIT: 1
```
olegrok: May be define it in `.dind`?

test/python/conftest.py
```diff
@@ -1,256 +1,133 @@
...
+# ################
+# Project fixtures
+# ################
+# There are three types of projects:
+# * light_project ({original,deprecated}_light_project):
+#   Cartridge CLI creates project with cartridge dependency by default.
+#   It's known that installing cartridge rocks is a long operation,
+#   so we don't want to perform it on every test.
+#   These fixtures are used to decrease packing time.
+#   They don't have cartridge dependency,
```
olegrok: nit: "don't have `a/the` cartridge"

test/python/project.py
```diff
@@ -0,0 +1,235 @@
...
+
+from utils import create_project
+from utils import recursive_listdir
+from utils import tarantool_enterprise_is_used
+
+
+CARTRIDGE_PACK_SPECIAL_FILES = {
+    'cartridge.pre-build',
+    'cartridge.post-build',
+    '.cartridge.ignore',
```
olegrok: May be it's better to add a note that two last files are from "deprecated" mode

test/python/test_docker_pack.py
```diff
@@ -150,9 +146,9 @@ def test_docker_pack(project, docker_image, tmpdir, docker_client):
     os.remove(arhive_path)
 
     assert_dir_contents(
-        files_list=recursive_listdir(os.path.join(tmpdir, 'usr/share/tarantool/', project['name'])),
-        exp_files_list=project['distribution_files_list'],
-        exp_rocks_content=project['rocks_content'],
+        files_list=recursive_listdir(os.path.join(tmpdir, 'usr/share/tarantool/', project.name)),
+        exp_files_list=project.distribution_files,
```
olegrok: usually `exp` is shortcut for `exponent`. Let's call it explicitly `expected`

test/python/test_docker_pack.py
dokshina: To simplify the test.

dokshina: The project is created on `docker_image` fixture setup.

olegrok: May be pass `project name` with image name to eliminate such hardcode?

dokshina: Yes, I think that docker_image object can contain field `project_name`.

test/python/project.py
```diff
@@ -0,0 +1,240 @@
...
+        # create project and save its path
+        self.path = create_project(basepath, name, template)
+
+        # project files list
+        self.project_files = recursive_listdir(self.path)
+
+        # files that should be delivered in the result package
+        self.distribution_files = filter_out_files_removed_on_pack(self.project_files)
+        self.distribution_files.add('VERSION')
+        if tarantool_enterprise_is_used():
```
olegrok: You use this function twice in this ... function

May be let's store it to var and reuse

test/python/project.py
```diff
@@ -59,10 +59,13 @@ def __init__(self, name, basepath, template='cartridge'):
         # project files list
         self.project_files = recursive_listdir(self.path)
 
+        # save tarantool_enterprise_is_used() result to variable
```
olegrok: It's a bit trivial

PR #122, opened 2020-01-22
Fix error on submodules cleanup

### LINKED ISSUE #120: rpm pack fails on git clean
```bash
docker run --name $tmp_name $image /bin/bash -c \
        "tarantoolctl rocks install cartridge-cli 1.3.1 && \
        /.rocks/bin/cartridge pack rpm /opt/project --version ${version} --name project"
```

```
Packing rpm file
Packing CPIO in: /tmp/CTtXas
Running `git clean`
/usr/libexec/git-core/git-submodule: line 509: 0: Bad file descriptor
ERROR: Failed to execute 'cd "/tmp/CTtXas/usr/share/tarantool/project" && /usr/bin/git submodule foreach --recursive /usr/bin/git clean -f -d -X'
```

v1.3.1
centos:7

dokshina:
Closes #120

### DIFF
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -1258,8 +1258,20 @@ local function form_distribution_dir(source_dir, dest_dir)
         -- Clean up all files explicitly ignored by git, to not accidentally
         -- ship development snaps, xlogs or other garbage to production.
         call("cd %q && %s clean -f -d -X", dest_dir, git)
+
+        info('Running `git clean` for submodules')
         -- Recursively cleanup all submodules
-        call("cd %q && %s submodule foreach --recursive %s clean -f -d -X", dest_dir, git, git)
+        local ret = os.execute(string.format(
+            "cd %q && %s submodule foreach --recursive %s clean -f -d -X",
+            dest_dir, git, git
+        ))
+        if not ret then
+            warn(
+                'Failed to run `git clean` for submodules: %s. ' ..
+                'It is possible that some of the extra files normally ignored '..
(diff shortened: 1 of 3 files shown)


cartridge-cli.lua
```diff
@@ -1258,8 +1258,20 @@ local function form_distribution_dir(source_dir, dest_dir)
         -- Clean up all files explicitly ignored by git, to not accidentally
         -- ship development snaps, xlogs or other garbage to production.
         call("cd %q && %s clean -f -d -X", dest_dir, git)
+
+        info('Running `git clean` for submodules')
         -- Recursively cleanup all submodules
-        call("cd %q && %s submodule foreach --recursive %s clean -f -d -X", dest_dir, git, git)
+        local ret = os.execute(string.format(
```
olegrok: Let's add a "raw_call" that has an interface similar to `call`, but don't call die.

May be returns `ok, err`

Firstly, it can be used in `call`, secondly it

dokshina: I prefer changing `call` function interface to return `output, err`.

We can use `call_or_die` function instead of current `call`.

```lua
local output, err = call('git clean')
if output == nil then warn('Failed to run git clean: %s', err) end

call_or_die('ls')
```
On the other hand, the main reason of adding `call` function was using command output.
Maybe it would be better to refactor `call` function to use `os.execute` and return `ok, err`, and add `check_output` function to use command output.

```lua
local ok, err = call('git clean')
call_or_die('ls')
local output, err = check_output('git describe')
```

PR #125, opened 2020-01-28
pack: fix rock pack

olegrok:
This patch fixes pack into rock.

In some cases cartridge could return following error
```
Type mismatch on field version: invalid value
'1.5.6-284-g10a507e4' does not match '[%w.]+-[%d]+'
(using rockspec format 1.0)
```

Luarocks requires a specific type of version format.
We should cut 'git format' to simple '[%w.]+-[%d]+'.

Also luarocks requires existance of "source.url" path
and faild with non-informative error otherwise.
So '/dev/null' in tests was replaced to 'file://.'

### DIFF
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -1518,7 +1518,10 @@ end
 local function get_rockspec_version(extended_version)
     local version, release = normalize_version(extended_version)
     release = tonumber(release:split('-', 1)[1]) or '0'
-    return version .. '-' .. release
+    local result = version .. '-' .. release
+    info('Version %s is transformed to %s to be compatible with luarocks format',
+        extended_version, result)
+    return result
 end
 
 local function pack_rock()
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -1514,6 +1514,13 @@ end
 
 -- * ---------------- ROCK packing ----------------
 
+-- luarocks requires specific format of version format: '[%w.]+-[%d]+'
+local function get_rockspec_version(extended_version)
+    local version, release = normalize_version(extended_version)
+    release = tonumber(release:split('-', 1)[1]) or '0'
+    return version .. '-' .. release
+end
+
 local function pack_rock()
     local tmpdir = fio.tempdir()
     local distribution_dir = fio.pathjoin(tmpdir, pack_state.name)
@@ -1533,10 +1540,11 @@ local function pack_rock()
 
     local rockspec = find_rockspec(distribution_dir)
     local content = ''
+    local rockspec_version = get_rockspec_version(pack_state.version_release)
     if rockspec then
         content = read_file(fio.pathjoin(distribution_dir, rockspec))
         content = string.gsub(content, "(.-version%s-=%s-['\"])(.-)(['\"].*)",
-                '%1' .. pack_state.version_release .. '%3')
+                '%1' .. rockspec_version .. '%3')
         if not content then
             die('Rockspec %s is not valid! Version not found!')
         end
@@ -1545,7 +1553,7 @@ local function pack_rock()
     local name_of_rockspec = string.format(
         '%s-%s.rockspec',
         pack_state.name,
-        pack_state.version_release
+        rockspec_version
     )
 
     local new_rockspec = fio.pathjoin(distribution_dir, name_of_rockspec)
@@ -1557,7 +1565,7 @@ local function pack_rock()
     local rock_filename = string.format(
         '%s-%s.*.rock',
         pack_state.name,
-        pack_state.version_release
+        rockspec_version
     )
 
     call_or_die('tarantoolctl rocks pack %s ', new_rockspec)
--- a/test/python/test_pack.py
+++ b/test/python/test_pack.py
@@ -287,12 +287,16 @@ def test_packing_with_git_file(project_with_git_file, tmpdir):
                          [
                              ('0.1.0', 'rpm', '0.1.0-0.rpm'),
                              ('0.1.0', 'deb', '0.1.0-0.deb'),
+                             ('0.1.0', 'rock', '0.1.0-0.src.rock'),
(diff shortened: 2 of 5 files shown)


test/python/test_pack.py
```diff
@@ -142,6 +144,26 @@ def rpm_archive_with_custom_units(module_tmpdir, light_project):
...
+
+    # luarocks requires existence of source.url
+    with fileinput.FileInput(project.rockspec_path, inplace=True) as file:
+        for line in file:
+            print(line.replace('/dev/null', 'file://.'), end='')
+
+    cmd = [os.path.join(basepath, "cartridge"), "pack", "rock", project.path]
+    process = subprocess.run(cmd, cwd=module_tmpdir)
+    assert process.returncode == 0, \
+        "Error during creating of rpm archive with project"
```
dokshina: `rpm` -> `rock`

test/python/test_pack.py
```diff
@@ -142,6 +144,26 @@ def rpm_archive_with_custom_units(module_tmpdir, light_project):
...
 
 
+@pytest.fixture(scope="module")
+def rock_archive(module_tmpdir, light_project):
+    project = light_project
+
+    # luarocks requires existence of source.url
+    with fileinput.FileInput(project.rockspec_path, inplace=True) as file:
+        for line in file:
+            print(line.replace('/dev/null', 'file://.'), end='')
```
dokshina: Please, add a comment about this street magic.

cartridge-cli.lua
```diff
@@ -1514,6 +1514,13 @@ end
 
 -- * ---------------- ROCK packing ----------------
 
+-- luarocks requires specific format of version format: '[%w.]+-[%d]+'
+local function get_rockspec_version(extended_version)
```
dokshina: Maybe we should log this action?

test/python/test_pack.py
```diff
@@ -224,6 +246,17 @@ def test_deb_pack(deb_archive, tmpdir):
...
 
+def test_rock_pack(rock_archive):
+    assert zipfile.is_zipfile(rock_archive.filepath), 'rock should be a valid zip-archive'
+    with zipfile.ZipFile(rock_archive.filepath, 'r') as zip_archive:
+        files = zip_archive.namelist()
+        # Tarantool 1.10 and Tarantool 2.2 use Luarocks 2 and Luarocks 3 respectively
+        # The content is this cases a bit different
+        # This checks is quite simple but shows that rock-archive is not empty
+        assert 'init.lua' in files, \
+            'Rock archive content is not as expected'
```
dokshina: Can you check that archive contains all `project.distribution_files`?

test/python/test_pack.py
```diff
@@ -224,6 +246,17 @@ def test_deb_pack(deb_archive, tmpdir):
             assert 'chown root:root /usr/lib/tmpfiles.d/{}.conf'.format(project.name) in postinst_script
 
 
+def test_rock_pack(rock_archive):
+    assert zipfile.is_zipfile(rock_archive.filepath), 'rock should be a valid zip-archive'
+    with zipfile.ZipFile(rock_archive.filepath, 'r') as zip_archive:
+        files = zip_archive.namelist()
+        # Tarantool 1.10 and Tarantool 2.2 use Luarocks 2 and Luarocks 3 respectively
```
dokshina: This is very sad, I sure that it would be great to check that archive contains correct name, version and other base information.

test/python/test_pack.py
olegrok: Thanks! Fixed

cartridge-cli.lua
olegrok: Done

test/python/test_pack.py
olegrok: This check will mean that we check that `tarantoolctl rocks pack` works correctly. I believe it's beyond responsibility of 'cartrige-cli'. I think such simple check should be enough

test/python/test_pack.py
olegrok: The package process is different for different luarocks versions and different rockspec content. I check that `cartridge-cli` doesn't fail on `pack rock` and I don't test `luarock pack`

cartridge-cli.lua
olegrok: Done

test/python/test_pack.py
dokshina: The main idea of checking package contents is checking that some file was filtered out by `cartridge-cli` before running `tarantoolctl rocks pack`.

test/python/test_pack.py
dokshina: Yes, but `cartridge-cli` is responsible for passing correct information to `luarocks`. We sure (hope) that `luarocks` works correctly, so this check means inly checking name and version that `cartridge-cli` passes to it.

test/python/test_pack.py
olegrok: No. In case of luarocks your logic doesn't work.

It depends on build type. You can specify list of needed files explicitly in your rockspec. You can choose "make" or "cmake" build type - and cmake and make will be responsible in this case.

olegrok: You can specify `build='none'` - and you will copy your sources

test/python/test_pack.py
olegrok: All our possible actions in case of `rock pack` is to pass correct rockspec to `luarock pack` and validate rockspec/files content before run `luarocks pack` - to show more informative error... But it's an overkill

test/python/test_pack.py
dokshina: But you definitely know that test project rockspec contains `source  = {{ url = 'file://.' }} build = {{ type = 'none' }}`.

olegrok:
After discussing I'm completely sure that `pack rock` should be completely removed from cartridge-cli. We are not able to adequate interface over standard luarocks what is compatible with luarocks 2 and 3

PR #127, opened 2020-01-29
Refactor error handling on `pack` and `create`

dokshina:
See @olegrok [suggestion](https://github.com/tarantool/cartridge-cli/pull/116#discussion_r368242804)

### DIFF
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -450,6 +450,7 @@ local function load_variables_from_file(filepath)
     local file_content, err = read_file(filepath)
     if file_content == nil then return nil, err end
 
+    -- remove shebang
     file_content = file_content:gsub("^#![^\n]*\n", "")
 
     local chunk, load_err = load(file_content, filepath, "t", res)
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -752,7 +752,7 @@ local function call(command, ...)
 end
 
 -- Runs command using `os.execute` and dies if command fails
-local function call_or_die(command, ...)
+local function call_or_die(command, ...)  -- luacheck: no unused
     local ok, err = call(command, ...)
 
     if not ok then
@@ -2876,14 +2876,16 @@ local function app_pack(args)
     end
 
     local ok_pack, err_pack = pack_handler(opts)
-
     if not ok_pack then
-        info('Failed to pack application...')
+        warn('Failed to pack application')
     end
 
     -- clean build directory
     info('Remove build directory %s', pack_state.build_dir)
-    remove_by_path(pack_state.build_dir)
+    local ok, err = remove_by_path(pack_state.build_dir)
+    if not ok then
+        warn('Failed to clean up build directory %s: %s', pack_state.build_dir, err)
+    end
 
     if not ok_pack then
         die('Failed to pack application: %s', err_pack)
@@ -3022,7 +3024,8 @@ node_modules
 ]]
 
 local function instantiate_template(template_dir, dest_dir, app_name)
-    local files = find_files(template_dir)
+    local files, err = find_files(template_dir)
+    if files == nil then return false, err end
 
     local context = {project_name=app_name,
                      project_name_lower=string.lower(app_name)}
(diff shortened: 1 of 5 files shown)


test/python/test_create.py
```diff
@@ -26,3 +28,19 @@ def test_project(default_project):
 
     process = subprocess.run(['.rocks/bin/luatest'], cwd=project.path)
     assert process.returncode == 0, "luatest failed"
+
+
+def test_project_recreation(default_project):
```
olegrok: May be it's better to move this test into separate PR and push it firstly?

I suppose that "refactoring" shouldn't change behaviour

cartridge-cli.lua
```diff
@@ -739,7 +752,7 @@ local function call(command, ...)
 end
 
 -- Runs command using `os.execute` and dies if command fails
-local function call_or_die(command, ...)
+local function call_or_die(command, ...)  -- luacheck: no unused
```
olegrok: Let's drop it

If you don't use it

cartridge-cli.lua
```diff
@@ -841,18 +854,21 @@ local function detect_git_version(source_dir)
...
 
 local function detect_name(source_dir)
-    local rockspec = find_rockspec(source_dir)
-    if rockspec ~= nil then
-        return string.match(rockspec, '^(%g+)%-scm%-1%.rockspec$')
-    end
+    local rockspec, err = find_rockspec(source_dir)
+    if rockspec == nil then return nil, err end
+
+    return string.match(rockspec, '^(%g+)%-scm%-1%.rockspec$')
```
olegrok: hm...

Why "scm", why "scm-1"? E.g in watchdog we use "scm-2" (https://github.com/tarantool/watchdog/blob/master/watchdog-scm-2.rockspec)

cartridge-cli.lua
```diff
@@ -1566,61 +1667,81 @@ local function pack_tgz()
...
-        content = read_file(fio.pathjoin(distribution_dir, rockspec))
+        local err
+        content, err = read_file(fio.pathjoin(distribution_dir, rockspec))
+        if content == nil then return false, err end
+
         content = string.gsub(content, "(.-version%s-=%s-['\"])(.-)(['\"].*)",
                 '%1' .. pack_state.version_release .. '%3')
         if not content then
-            die('Rockspec %s is not valid! Version not found!')
+            return false, 'Rockspec %s is not valid! Version not found!'
```
olegrok: I suppose you miss a format here (because of "%s")

cartridge-cli.lua
```diff
@@ -2080,40 +2209,56 @@ local function pack_cpio(opts)
...
+    for _, f in ipairs({'unpacked', 'files'}) do
+        local filepath = fio.pathjoin(pack_state.build_dir, f)
+        local ok, err = remove_by_path(filepath)
+        if not ok then return false, err end
+    end
 
     local fileinfo = generate_fileinfo(pack_state.build_dir)
 
-    return archive, fileinfo, payloadsize
+    return true, {
```
olegrok: I think in "positive" case we should return only one argument. Two arguments is sign of an error. Let's return `nil, err` in case of error and table otherwise

cartridge-cli.lua
```diff
@@ -2670,34 +2873,48 @@ local function app_pack(args)
...
             docker_build_args = args.docker_build_args,
-        })
-    else
-        die("Unknown package type: %s", args.type)
+        }
+    end
+
+    local pack_handler = pack_handlers[args.type]
+    if pack_handler == nil then
+        local handler_err = string.format("Pack handler for %s distribution type not found", args.type)
```
olegrok: Do you really ask me to report a bug in case if I write `cartridge pack rar`?

cartridge-cli.lua
```diff
@@ -2670,34 +2873,48 @@ local function app_pack(args)
...
+        die(
+            "Whoops! It looks like something is wrong with this version of Cartridge CLI. " ..
+            "Please, report a bug on https://github.com/tarantool/cartridge-cli/issues/new. " ..
+            "The error is: %s.", handler_err
+        )
+    end
+
+    local ok_pack, err_pack = pack_handler(opts)
+    if not ok_pack then
+        warn('Failed to pack application')
```
olegrok: Where is an error message here (err_pack)?

cartridge-cli.lua
```diff
@@ -2846,13 +3063,74 @@ local function instantiate_template(template_dir, dest_dir, app_name)
...
-        write_file(destname, content, mode)
+        local ok, err = write_file(destname, content, mode)
+        if not ok then return false, err end
     end
+
+    return true
 end
 
+local function create_app_directory_and_init_git(dest_dir, template, name)
+    assert(fio.path.exists(dest_dir))
```
olegrok: assert?

cartridge-cli.lua
dokshina: In your case you will get an error from `app_pack_parse` function, but if there is no pack handler for one of allowed types, it's internal error.

cartridge-cli.lua
dokshina: Yes, this function expects that `dest_dir` is already exists.

test/python/test_create.py
dokshina: Ok, I will do it.

cartridge-cli.lua
dokshina: Ok, I agree.

cartridge-cli.lua
dokshina: Yes, thanks!

cartridge-cli.lua
dokshina: The main idea of this warning was to explain user why his build process suddenly started to clean build directory...
I will think about it and make log more user-friendly and informative.

cartridge-cli.lua
dokshina: I changed the name detecting approach.
I don't use rockspec filename anymore, I use `package` value from rockspec.

cartridge-cli.lua
olegrok: ok

cartridge-cli.lua
olegrok: I mean that it should be `warn('Failed to pack application: %s', err_pack)`

cartridge-cli.lua
```diff
@@ -2832,13 +3048,74 @@ local function instantiate_template(template_dir, dest_dir, app_name)
...
+    local git = which('git')
+
+    if git ~= nil then
+        info("Initializing git repo in: %s", dest_dir)
+
+        repeat  -- until true
+            local ok, err = call("cd %s && %s init .", dest_dir, git)
+            if not ok then
+                warn('Failed to initialize git repo: %s', err)
+                break
```
olegrok: Why not simple `return`? 

Why `break`?

dokshina: If some additional actions will be added after git init, it will require changing all these `break` keywords to `return`.

cartridge-cli.lua
dokshina: I thought that it should be smth like:
```
Packing rpm file
WARNING: Failed to pack application
Remove build directory ~/.cartridge/tmp/cartridge-build-fc3c12ba73e6b743
ERROR: Failed to pack application: Some error occured
```

olegrok:
```
WARNING: Failed to pack application
ERROR: Failed to pack application: Some error occured
```

It's a duplication

dokshina: Do you suggest smth like this?

```
ERROR: Failed to pack application: Some error occured
Remove build directory...
```

olegrok: Yes

dokshina: The main problem is that `die` function writes message with `ERROR` prefix and then calls `os.exit(1)`. Your suggestion requires adding `err` function that writes `ERROR` message and doesn't call `os.exit(1)` (to allow performing cleanup). Then, `os.exit` should be called from `app_pack`.

I don't want to separate these actions, I suggest using:

```
WARNING: Failed to pack application                                 # Notify user that smth wrong
Remove build directory...                                                      # Cleanup
ERROR: Failed to pack application: Some error occured   # Explain the error
```

PR #129, opened 2020-01-29
Update luatest to 0.5.0, add luacov

printercu:


examples/getting-started-app/getting-started-app-scm-1.rockspec
```diff
@@ -7,7 +7,7 @@ source  = {
 dependencies = {
     'tarantool',
     'lua >= 5.1',
-    'luatest == 0.4.0-1',
+    'luatest == 0.5.0-1',
     'ldecnumber == 1.1.3-1',
     'cartridge == 1.2.0-1',
```
olegrok: Let's bump it to 2.0.1

printercu: It should be updated along with getting started readme

PR #133, opened 2020-02-03
Cartridge build command

### LINKED ISSUE #132: Implement `cartridge build` command
The main idea of this feature is to simplify local development.
`cartridge build` command is designed to be used for building application locally.

It will allow to start applications this way:

```bash
$ cartridge create --name myapp
$ cd myapp
$ cartridge build
$ cartridge start
```

### RFC

Application build comprises these steps:

* Running `cartridge.pre-build` (or `.cartridge.pre`);
* Running `tarantoolctl rocks make` to install required rocks modules;

It's important to remember that the main purpose of `cartridge.pre-build` script is  installing non-standard rocks modules, so it can't contain some commands like `yum install` etc. 

#### Command

`cartridge build [<path>]` - builds an application locally. 

* `path` - path to application (default: `'.'`)

This command runs `cartridge.pre-build` script and then calls `tarantoolctl rocks make` directly in the project directory.

dokshina:
Closes #132

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,8 +111,11 @@ cartridge pack rpm ./myapp
 
 ### Building an application
 
-You can call `cartridge build` command to build application locally.
+You can call `cartridge build [<path>]` command to build application locally.
 It can be useful for local development.
+
+This command requires one argument - path to the application.
+By default - it's `.` (current directory).
 
 These steps will be performed on running this command:
 
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ It can be useful for local development.
 
 These steps will be performed on running this command:
 
-* running `cartridge.pre-build` or `.cartridge.pre` [DEPRECATED];
+* running `cartridge.pre-build` (or [DEPRECATED] `.cartridge.pre`);
 * running `tarantoolctl rocks make`.
 
 ### Application packing details
--- a/README.md
+++ b/README.md
@@ -74,6 +74,14 @@ For more details, say:
 cartridge --help
 ```
 
+These commands are supported:
+
+* `create` - create a new app from template;
+* `pack` - pack application into a distributable bundle;
+* `build` - build application for local development;
+* `start` - start a Tarantool instance(s);
+* `stop` - stop a Tarantool instance(s).
+
 ### Applications lifecycle
 
 Create an application from a template:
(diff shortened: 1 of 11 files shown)


templates/cartridge/deps.sh
```diff
@@ -1,10 +1,9 @@
 #!/bin/sh
```
olegrok: May be we should remove `deps.sh` and use hooks instead?

knazarov: This is reasonable. But then we need to also change the getting started example.

dokshina: This script is used to install dev dependencies, such as luatest and luacheck. I think that it shouldn’t be removed, it allows to fix dev dependencies versions in application template.
But I agree that getting-started example should be changed to use a new build command.

olegrok: Let's do it in separate PR

README.md
```diff
@@ -82,12 +90,35 @@ Create an application from a template:
...
 ```
 
+### Building an application
+
+You can call `cartridge build` command to build application locally.
+It can be useful for local development.
+
+These steps will be performed on running this command:
+
+* running `cartridge.pre-build` or `.cartridge.pre` [DEPRECATED];
```
olegrok: It could confuse users. It's not clear `.cartridge.pre` is deprecated or both '`cartridge.pre-build` or `.cartridge.pre`'

cartridge-cli.lua
```diff
@@ -2726,6 +2749,13 @@ local function check_if_deprecated_build_flow_is_ised(app_path)
         end
     end
 
+    if deprecated_build_flow_is_ised then
```
olegrok: I believe that moving of this part is redundant

test/python/test_pack.py
```diff
@@ -149,65 +150,81 @@ def rpm_archive_with_custom_units(module_tmpdir, light_project):
...
+        'rpm', '--checksig', '-v', rpm_archive.filepath
+    ]
+    process = subprocess.run(cmd)
+    assert process.returncode == 0, "RPM signature isn't correct"
+
 
 def test_deb_pack(deb_archive, tmpdir):
     project = deb_archive.project
 
+    extract_dir = os.path.join(tmpdir, 'extract')
```
olegrok: Describe please

cartridge-cli.lua
dokshina: This warning message was printed only on application packing. I sure that we should warn on building application too.

test/python/test_pack.py
dokshina: You mean describing to you or adding a comment?)

Anyway, to correctly check package files archive, it should be extracted to the empty directory. Since I use `tmpdir` for all files used in test (project files etc.), I need to create subdirectory.

olegrok: Second one. Ok. However, it's not clear at first

cartridge-cli.lua
olegrok: Ok, yes

README.md
dokshina: I agree with you.

Maybe

running `cartridge.pre-build` (or  [DEPRECATED] `.cartridge.pre`);

olegrok: Yes, it's better

test/python/test_pack.py
dokshina: I added a comment, but I'm not sure that it's clear enough.

PR #136, opened 2020-02-04
Add default path to cartridge pack command

dokshina:
Allow to run `cartridge pack <type>` in the project directory without specifying path `.`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - By default, temporary directory for application building is created in
   `~/.cartridge/tmp`
 - Commands usage messages are prettified
+- `path` argument for `cartridge pack` command isn't required.
+  By default, current directory is used.
 
 ## [1.3.2] - 2020-01-23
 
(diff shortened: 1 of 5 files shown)


cartridge-cli.lua
```diff
@@ -2999,6 +3000,14 @@ function cmd_pack.parse(arg)
     args.type = parameters[1]
     args.path = parameters[2]
 
+    if args.path == nil then
+        args.path = '.'
```
olegrok: Why not `fio.cwd()`?

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - By default, temporary directory for application building is created in
   `~/.cartridge/tmp`
 - Commands usage messages are prettified
+- `path` argument for `cartridge pack` command isn't required.
```
olegrok: I assume not only `pack` but also `build`

dokshina: This is **changed** section, this behaviour was *changed* only for `pack`. Build command was *added* and it's mentioned in **added** section.

cartridge-cli.lua
dokshina: Wow, thanks a lot, I missed this function! 
I will fix it anywhere in the code.

PR #138, opened 2020-02-05
Flush after printing messages

dokshina:


cartridge-cli.lua
```diff
@@ -325,22 +325,26 @@ end
 local function die(fmt, ...)
     local msg = "ERROR: " .. string.format(fmt, ...)
     print(colored_msg(msg, ERROR_COLOR_CODE))
```
olegrok: May be move to

```lua
local function print_and_flush(msg)
    io.write(msg, '\n')
    io.flush()
end
```
?

dokshina: Done

PR #139, opened 2020-02-05
Release 1.4.0

dokshina:


cartridge-cli/VERSION.lua
```diff
@@ -1 +1 @@
-return '1.3.2'
+return '1.4.0'
```
olegrok: Why not 1.3.3?

dokshina: Because there are new features - `cartridge build` and build directory specifying.

PR #141, opened 2020-02-07
[getting-started][readme] Add info about luacov

lenkis:


examples/getting-started-app/README.md
```diff
@@ -743,3 +746,33 @@ The output will look like this:
 Ran 2 tests in 0.021 seconds, 2 successes, 0 failures
 OK
 ```
+You can also use `luacov` to check the code coverage of the tests.
+
+Install `luacov`:
+
+``` bash
+getting-started-app $ tarantoolctl rocks install luacov
```
dokshina: I'm not sure that we should install `luacov` this way:

* we have already installed it by running `./deps.sh`;
* a few lines above we use `.rocks/bin/luatest` (installed by `./deps.sh` as well);
* I sure that using `deps.sh` is much better, because it contains fixed versions of rocks, and `rocks install` uses `scm-1` that can occasionally become non-compatible with the old version.

PR #142, opened 2020-02-07
Improve parsing commands args

dokshina:
This PR solves next problem: if you want to pass some boolean option (e.g. `daemonize`) to `cartridge` CLI, you should do it this way:

```bash
cartridge start --run_dir ${RUN_DIR} --daemonize  # OK
cartridge start --daemonize --run_dir ${RUN_DIR}  # FAIL
```

The second command (with boolean option passed not at the end of input) will fail because internal `argparse` module that is used in `cartridge-cli` doesn't support this feature.

Now, there is internal `argparse` wrapper that rearranges boolean options to be parsed correctly.

Moreover, is supports arguments schema, that allows to parse positional args as well as options.

```lua
parse_command_args('--name NAME --debug TYPE PATH', {
    opts = {
        name = 'string',
        debug = 'boolean',
    },
    args = {
        'type',
        'path',
    }
})
-- wil return
{
    name = 'NAME',
    debug = true,
    type = 'TYPE',
    path = 'PATH',
}
```

It also allows to use `--long-option` as well as `--long_option`. In both cases this is the `long_option`.

### DIFF
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -289,6 +289,11 @@ local function raw_option_name(arg)
     return arg:gsub('^%-%-?', '')
 end
 
+local function prettify_option_name(opt_name)
+    local pretyy_opt_name = opt_name:gsub('_', '-')
+    return pretyy_opt_name
+end
+
 local BOOLEAN_VALUES = {'0', '1', 'true', 'false'}
 
 local function parse_command_args(args, schema)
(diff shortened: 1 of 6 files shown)


cartridge-cli.lua
```diff
@@ -278,6 +278,110 @@ local function globtopattern(g)
...
+end
+
+local function raw_option_name(arg)
+    assert(is_option_name(arg))
+    return arg:gsub('^%-%-?', '')
+end
+
+local BOOLEAN_VALUES = {'0', '1', 'true', 'false'}
+
+local function parse_command_args(args, schema)
```
olegrok: Could we split "validation" and "parsing" into two separate steps?

PR #144, opened 2020-02-07
Fix parsing pretty boolean flags

dokshina:


olegrok:
Do you think we should validate the following case?
```lua
tarantool> require('cartridge-cli').parse({'--arg', '--a_rg'}, { opts = { arg = 'string' } })
---
- null
- 'Parse error: builtin/internal.argparse.lua:42: Bad input for parameter "arg". Expected
  string, got "true"'
..
```

(Yes, I will agree with you if you say that it's an overkill)

PR #145, opened 2020-02-10
Improve negative tests

dokshina:
Capture errors output to check that command fail reason is right

### DIFF
--- a/test/python/test_docker_pack.py
+++ b/test/python/test_docker_pack.py
@@ -15,6 +15,7 @@ from utils import tarantool_enterprise_is_used
 from utils import recursive_listdir
 from utils import assert_distribution_dir_contents
 from utils import assert_filemodes
+from utils import run_command_and_get_output
 
 
 # #############
@@ -120,8 +121,11 @@ def test_invalid_base_dockerfile(project_without_dependencies, module_tmpdir, tm
         "--from", invalid_dockerfile_path,
         project_without_dependencies.path,
     ]
-    process = subprocess.run(cmd, cwd=module_tmpdir)
-    assert process.returncode == 1
+
+    rc, output = run_command_and_get_output(cmd, cwd=module_tmpdir)
+    assert rc == 1
+    assert 'base dockerfile validation failed' in output
+    assert 'base image must be centos:8' in output
 
 
 def test_using_deprecated_files(deprecated_light_project, tmpdir):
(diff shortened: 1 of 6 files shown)


test/python/test_docker_pack.py
```diff
@@ -228,14 +233,15 @@ def test_base_dockerfile_with_env_vars(project_without_dependencies, module_tmpd
...
     ]
-    process = subprocess.run(cmd, cwd=module_tmpdir)
-    assert process.returncode == 0
+    rc, output = run_command_and_get_output(cmd, cwd=module_tmpdir)
+    assert rc == 0
+    assert 'Detected base Dockerfile {}'.format(dockerfile_with_env_path) in output
 
 
-def test_docker_e2e(docker_image, tmpdir, docker_client):
+def test_e2e(docker_image, tmpdir, docker_client):
```
olegrok: but why?

test/python/test_pack.py
```diff
@@ -354,15 +357,19 @@ def test_packing_with_wrong_filemodes(project_without_dependencies, tmpdir):
...
 
     # add file with invalid (700) mode
-    filepath = os.path.join(project.path, 'wrong-mode-file.lua')
+    filename = 'wrong-mode-file.lua'
+    filepath = os.path.join(project.path, filename)
     with open(filepath, 'w') as f:
         f.write("return 'My filemode is wrong'")
     os.chmod(filepath, 0o700)
 
+    print('filepath:', filepath)
```
olegrok: debug

test/python/utils.py
```diff
@@ -237,3 +237,21 @@ def check_package_files(project, basedir):
...
+        cwd=cwd,
+        stderr=subprocess.STDOUT,
+        stdout=subprocess.PIPE
+    )
+
+    stdout, _ = process.communicate()
+    stdout = stdout.decode('utf-8')
+
+    # This print is here to make running tests with -s flag more verbose
+    print(stdout)
```
olegrok: May be `log`?

test/python/test_docker_pack.py
dokshina: To get rid of `test_docker_pack.py::test_docker_pack` and `test_docker_pack.py::test_docker_e2e`.

I sure that `test_docker_pack.py::test_pack` and `test_docker_pack.py::test_e2e` are better.

test/python/test_pack.py
dokshina: thx

test/python/utils.py
dokshina: Why?

test/python/test_docker_pack.py
olegrok: Ok, agree

test/python/utils.py
olegrok: I think that uncontrolled output into stdout is a mauvais

dokshina: This print is here only to prevent adding it on every `pytest -s` call (I use it to debug some things in `cartridge-cli` itself). I don't see any problems here, there is `cartridge-cli` output anywayю

PR #146, opened 2020-02-11
Fix Tarantool 2.1 repo version compution

dokshina:
Use 2x instead of 2_1

### DIFF
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -2729,6 +2729,8 @@ local function tarantool_repo_version()
     local parts = _TARANTOOL:split('.')
     local version = string.format('%s_%s', parts[1], parts[2])
 
+    -- For Tarantool 2.1 tarantool/2x repo is used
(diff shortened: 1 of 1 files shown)


cartridge-cli.lua
```diff
@@ -2727,7 +2727,13 @@ end
 
 local function tarantool_repo_version()
     local parts = _TARANTOOL:split('.')
-    return string.format('%s_%s', parts[1], parts[2])
+    local version = string.format('%s_%s', parts[1], parts[2])
+
+    if version == '2_1' then
```
olegrok: may be comment?

olegrok: Test is the best variant (of course). But it leads to adding new target to CI

dokshina: Yes, I will add a comment.

I was thinking about unit test, but I realized that changing `_G._TARANTOOL` isn't good idea.
Anyway, I think that comment will help to prevent changing this code.

PR #147, opened 2020-02-12
Add --build-from option

dokshina:
Use --build-from option to specify build image base layers on building in
docker
Use --from option to specify runtime imgae base layers on packing to
docker
Tests

See [coment](https://github.com/tarantool/cartridge-cli/issues/98#issuecomment-584737756) for details.

### DIFF
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -3308,7 +3308,7 @@ function cmd_pack.callback(args)
         end
 
         app_state.build_base_dockerfile_layers = get_dockerfile_base_layers(
-            args.build_base, DEFAULT_BUILD_BASE_DOCKERFILE_LAYERS
+            args.build_from, DEFAULT_BUILD_BASE_DOCKERFILE_LAYERS
         )
     end
 
@@ -3395,7 +3395,7 @@ function cmd_pack.parse(cmd_args)
             download_token = 'string',
             tag = 'string',
             from = 'string',
-            build_base = 'string',
+            build_from = 'string',
             use_docker = 'boolean',
         }
     }
@@ -3432,23 +3432,23 @@ function cmd_pack.parse(cmd_args)
         end
     end
 
-    if args.build_base == nil then
+    if args.build_from == nil then
         local default_build_base_dockerfile_path = fio.pathjoin(
             args.path,
             DEFAULT_BUILD_BASE_DOCKERFILE_NAME
         )
         if fio.path.exists(default_build_base_dockerfile_path) then
-            args.build_base = default_build_base_dockerfile_path
+            args.build_from = default_build_base_dockerfile_path
         end
     end
 
     if args.from == nil then
-        local default_from_dockerfile_path = fio.pathjoin(
+        local default_runtime_base_dockerfile_path = fio.pathjoin(
             args.path,
             DEFAULT_RUNTIME_BASE_DOCKERFILE_NAME
         )
-        if fio.path.exists(default_from_dockerfile_path) then
-            args.from = default_from_dockerfile_path
+        if fio.path.exists(default_runtime_base_dockerfile_path) then
+            args.from = default_runtime_base_dockerfile_path
         end
     end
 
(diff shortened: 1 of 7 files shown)


cartridge-cli.lua
```diff
@@ -3373,6 +3391,7 @@ function cmd_pack.parse(cmd_args)
             download_token = 'string',
             tag = 'string',
             from = 'string',
+            build_base = 'string',
```
olegrok: build_base_image/build_base_dockerfile/...?

templates/cartridge/Dockerfile.build.cartridge
```diff
@@ -0,0 +1,14 @@
...
+# when --use-dcoker option is specified
+
+# The base image must be centos:8
+FROM centos:8
+
+# Here you can install some packages required
+#   for your application build
+
+# RUN set -x \
+#    && curl -sL https://rpm.nodesource.com/setup_8.x | bash - \
```
olegrok: Node 8 is no longer supported (https://blog.risingstack.com/update-nodejs-8-end-of-life-no-support/)

olegrok: 10.x

cartridge-cli.lua
```diff
@@ -1728,19 +1732,23 @@ end
 
 local function construct_build_image_dockerfile()
     -- The application build dockerfile consent theese parts:
-    -- - dockerfile_base_layers: the base image (passed in app_state.dockerfile_base_layers)
+    -- - build_base_dockerfile_layers: the base image
     -- - prepare: installing packages required for build (git gcc make cmake unzip)
     --            and creating tarantool user and directories
     -- - install_tarantool: install Tarantool on image
 
+    if app_state.build_base_dockerfile_layers == nil then
```
olegrok: Should we use some default value in this case?

dokshina: The default value for base dockerfiles is set in `cmd_pack.parse` funcion. 
Here we have just `app_state` with all values set.

cartridge-cli.lua
dokshina: I sure that CLI options should be as short as it's possible.

`--build-base` is short, but `base-build-dockerfile` is more informative. 

Maybe `--build-from` is better (similar to `--from`).

cartridge-cli.lua
olegrok: Hm... Is this check redundant then?
Or it should be an assert (because can happen only in dev)?

cartridge-cli.lua
olegrok: > Maybe --build-from

IMO, yes

cartridge-cli.lua
dokshina: Yes, I agree that it should be an assert.

dokshina:
@olegrok, what do you think about default build and runtime dockerfiles names?

Now, it's `Dockerfile.cartridge` for runtime (`--from`) and `Dockerfile.build.cartridge` for build (`--build-from`).

olegrok:
> @olegrok, what do you think about default build and runtime dockerfiles names?
> 
> Now, it's `Dockerfile.cartridge` for runtime (`--from`) and `Dockerfile.build.cartridge` for build (`--build-from`).

For me it looks good

However, consider the way then where is no any default Dockerfile names and user should pass it explicitly - on the one hand it could be a bit tiresomely, on the another hand - it's completely clear for users.

dokshina:
> on the one hand it could be a bit tiresomely, on the another hand - it's completely clear for users.

I agree with you, but I sure that using `.cartridge` suffix makes using this Dockerfiles clear enough. 
You can't create `Dockerfile.cartridge` for other purpose, and you aren't surprised when it used on `cartridge pack docker`.

cartridge-cli.lua
dokshina: BTW, [here](https://github.com/tarantool/cartridge-cli/pull/148) asserts are caught and logged correclty.

PR #149, opened 2020-02-14
Improve python e2e tests

dokshina:
* Separate e2e tests for pack in a module
* Test running application from tgz packed in docker

### DIFF
--- a/test/python/conftest.py
+++ b/test/python/conftest.py
@@ -20,7 +20,7 @@ def module_tmpdir(request):
     return str(dir)
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture(scope="session")
 def docker_client():
     client = docker.from_env()
     return client
--- a/test/python/conftest.py
+++ b/test/python/conftest.py
@@ -1,6 +1,7 @@
 import py
 import pytest
 import tempfile
+import docker
 
 from project import Project
 from project import remove_dependency
@@ -17,6 +18,12 @@ def module_tmpdir(request):
     dir = py.path.local(tempfile.mkdtemp())
     request.addfinalizer(lambda: dir.remove(rec=1))
     return str(dir)
+
+
+@pytest.fixture(scope="module")
+def docker_client():
+    client = docker.from_env()
+    return client
 
 
 # ################
@@ -86,9 +93,9 @@ def light_project(original_light_project, deprecated_light_project, request):
 #########################
 # Projects with cartridge
 #########################
-@pytest.fixture(scope="module")
-def original_project_with_cartridge(module_tmpdir):
-    project = Project('original-project-with-cartridge', module_tmpdir, 'cartridge')
+@pytest.fixture(scope="function")
+def original_project_with_cartridge(tmpdir):
+    project = Project('original-project-with-cartridge', tmpdir, 'cartridge')
     remove_dependency(project, 'luatest')
 
     add_dependency_submodule(project)
@@ -96,9 +103,9 @@ def original_project_with_cartridge(module_tmpdir):
     return project
 
 
-@pytest.fixture(scope="module")
-def deprecated_project_with_cartridge(module_tmpdir):
-    project = Project('deprecated-project-with-cartridge', module_tmpdir, 'cartridge')
+@pytest.fixture(scope="function")
+def deprecated_project_with_cartridge(tmpdir):
+    project = Project('deprecated-project-with-cartridge', tmpdir, 'cartridge')
     remove_dependency(project, 'luatest')
 
     add_dependency_submodule(project)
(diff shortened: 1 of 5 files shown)


test/python/utils.py
```diff
@@ -255,3 +283,106 @@ def run_command_and_get_output(cmd, cwd=None, env=None):
...
+
+
+def find_image(docker_client, project_name):
+    for image in docker_client.images.list():
+        for t in image.tags:
+            if t.startswith(project_name):
+                return t
+
+
+def wait_for_container_start(container, timeout=10):
```
olegrok: Consider "HEALTHCHECK" section usage then you should wait for "healthy" status

test/python/conftest.py
```diff
@@ -86,19 +93,19 @@ def light_project(original_light_project, deprecated_light_project, request):
 #########################
 # Projects with cartridge
 #########################
-@pytest.fixture(scope="module")
```
olegrok: Are you ready for it? It could significantly slow down testing process

test/python/test_pack.py
```diff
@@ -160,7 +150,7 @@ def rpm_archive_with_custom_units(tmpdir, light_project):
 # #####
 # Tests
 # #####
-def test_tgz_pack(tgz_archive, tmpdir):
```
olegrok: Seems you try to save several bytes of source code)

I don't see significant reasons to rename. Yes it will be displayed as "test_pack::test_tgz" but it's a strange perfectionism

test/python/conftest.py
dokshina: Yes, I sure that `project` fixtures should have `function` scope to prevent accidental changing project files.

This fixture had `module` scope because it was used for `docker_image` fixture with `module` scope (it was used for testing image contents and e2e).

Now, I use light project to test image/archive contents and use `project_with_cartridge` fixture only for e2e tests (one archive per test).

test/python/test_pack.py
dokshina: No, I just think, that `test_pack::test_tgz` looks better than `test_pack::test_pack_tgz` ;(

test/python/utils.py
dokshina: I didn't get it, can you explain?

olegrok: https://docs.docker.com/engine/reference/builder/#healthcheck

test/python/conftest.py
```diff
@@ -19,6 +20,12 @@ def module_tmpdir(request):
     return str(dir)
 
 
+@pytest.fixture(scope="module")
```
olegrok: scope=session?

test/python/conftest.py
olegrok: agree

test/python/test_pack_e2e.py
```diff
@@ -0,0 +1,197 @@
...
+        self.http_port = http_port
+        self.advertise_port = advertise_port
+
+
+# ########
+# Fixtures
+# ########
+@pytest.fixture(scope="function")
+def image_name_for_tests(docker_client, tmpdir, request):
+    if tarantool_enterprise_is_used():
```
olegrok: It could a fixture with scope state. In fact simple boolean value

test/python/test_pack_e2e.py
```diff
@@ -0,0 +1,197 @@
...
+    with open(test_image_dockerfile_path, 'w') as f:
+        f.write('''
+            FROM centos:8
+            RUN curl -s \
+                https://packagecloud.io/install/repositories/tarantool/{}/script.rpm.sh | bash \
+                && yum -y install tarantool tarantool-devel
+        '''.format(tarantool_repo_version()))
+
+    IMAGE_NAME = 'test-image'
+    docker_client.images.build(
```
olegrok: Does pydocker support options as "enable build kit" and something for more aggressive caching?

test/python/utils.py
dokshina: This function waits while Tarantool instance in container starts (it waits `entering the event loop` in logs).

test/python/conftest.py
dokshina: Yes, thanks.

test/python/test_pack_e2e.py
dokshina: No. But it's possible that setting environment variable `DOCKER_BUILDKIT` affects using this feature.

test/python/utils.py
olegrok: I understand. But I think it's quite similar to situation then container goes to "healthy" state

dokshina: So, you suggest adding HEALTHCHECK section to `cartridge pack docker` result container?

I agree, that it's a good idea, but I sure that is should be implemented in a separated PR.

test/python/test_pack_e2e.py
dokshina: I don't like the idea to pass it explicitly in every test where it's used.
I sure that using lazy initialization for `tarantool_version` is enough to use this function without additional optimization.

PR #158, opened 2020-02-24
Prevent root files creation on packing in docker

dokshina:
Now application build in docker creates files owned by root in `<build_dir>/<project>/.rocks` directory. As a result, build directory can't be deleted at the end of packing.

This patch creates user with the same UID on build image and uses it to run application build.
The same approach is used by [packpack](https://github.com/packpack/packpack/blob/master/packpack#L115-L122).


cartridge-cli.lua
```diff
@@ -1320,6 +1320,21 @@ RUN DOWNLOAD_URL=https://tarantool:${"$"}{DOWNLOAD_TOKEN}@download.tarantool.io
 ENV PATH="/usr/share/tarantool/tarantool-enterprise:${"$"}{PATH}"
 ]]
 
+local DOCKERFILE_WRAP_USER = [[
```
olegrok: Please add a comment about following action and its goal

cartridge-cli.lua
```diff
@@ -1758,11 +1782,27 @@ local function construct_build_image_dockerfile()
...
+        return nil, string.format('Failed to get user ID: %s', err)
+    end
+
+    user_id = user_id:strip()
+
+    local wrap_user_part = expand(
+        DOCKERFILE_WRAP_USER,
+        {
+            user_id = user_id,
+            username = os.getenv('USER') or 'myuser',
```
olegrok: Hm... Why "myuser", could we call `whoami` or `id -un` in this case?

test/python/test_pack.py
```diff
@@ -536,3 +536,25 @@ def test_base_build_dockerfile_with_env_vars(project_without_dependencies, pack_
     rc, output = run_command_and_get_output(cmd, cwd=tmpdir)
     assert rc == 0
     assert 'Detected base Dockerfile {}'.format(dockerfile_with_env_path) in output
+
+
+# @pytest.mark.parametrize('pack_format', ['rpm'])
```
olegrok: redundant comment?

cartridge-cli.lua
dokshina: In fact, username It doesn't matter (only UID matters), but calling additional command can cause additional errors.

PR #160, opened 2020-02-29
Small clarification

slemonide:
about why bucket_id is there


artur-barsegyan:
Thank you for your improvements. But if you want to patch `README.md`, you also need to patch `README_RUS.md` because we need to support all of it.

PR #166, opened 2020-03-12
Build in docker reworks

### LINKED ISSUE #165: Make packing with TE more flexible
### Problems

* Bundle download URL isn't correct, is uses `tarantool` user (and requires passing download token as argument).

* Now `pack docker` builds application in the docker container. As a result bundle is always downloaded on image. We should use bundle that is used locally (like in local build) by default.

Packing in docker is used for `pack docker` and `pack {rpm,tgz,deb} --use-docker`.

### Solution

In Tarantool Enterprise case, both of this commands should deliver Tarantool Enterprise bundle on the build image. SDK can be downloaded using download URL or copied from local source.

Now packing in docker for TE requires `--download-token` argument.
I suggest replacing it with `--download-url` (or  `--sdk-download-url`, or `--sdk-url`) and `--sdk-path` arguments.

So:

* by default: current SDK is used (with warning);

* if `--sdk-path` is specified: bundle is copied from localhost to container;

* if `--download-url` is specified: bundle archive is downloaded from this URL (env `TARANTOOL_DOWNLOAD_URL`).

Maybe it would be better to replace default behaviour with a flag (e.g. `--local-sdk`) to make it more clear for the end user.
I sure, that when `pack {rpm,tgz,deb} --use-docker` is used, user doesn't want to use his current SDK (otherwise he doesn't specify `--use-docker` option). 
In this case `--local-sdk` will be useful to make Cartridge CLI behaviour more clear.

### Questions

Now in case of TE the result package (or container) contains only `tarantool` and `tarantoolctl` binaries from SDK. Is is OK? Should we deliver the whole bundle?

dokshina:
Closes #165

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,38 +295,27 @@ across all sections of the YAML file(s) stored in `/etc/tarantool/conf.d/*`.
 
 Specific options:
 
-* `--from` - path to the base image dockerfile;
-
 * `--tag` - resulting image tag;
 
-* `--download-token` (env `TARANTOOL_DOWNLOAD_TOKEN`) - download token for
-  installing Tarantool Enterprise to the resulting image.
+* `--from` - path to the base dockerfile for runtime image 
+  (default to `Dockerfile.cartridge` in the project root);
 
-The base image is `centos:8`. On this image, `cartridge` will install all
-packages required for `cartridge` rocks and for the default `cartridge`
-application (i.e. the one created with `cartridge create`).
+* `--build-from` - path to the base dockerfile for build image
+  (default to `Dockerfile.build.cartridge` in the project root);
 
-If your application requires some other applications, you can specify your own
-base image.
-The base image dockerfile should be specified in the `Dockerfile.cartridge` file
-in the project root.
-Or you can pass a path to another dockerfile via the `--from path/to/dockerfile`
-option.
+* `--sdk-local` - flag indicates if local SDK should be installed on the image;
 
-The base image dockerfile should be started with the `FROM centos:8` line
-(except comments).
+* `--sdk-path` - path to SDK to be installed on the image;
 
-Example Dockerfile:
+* `--sdk-url` (env `TARANTOOL_SDK_DOWNLOAD_URL`, has lower priority) - URL to get 
+  Tarantool Enterprise SDK on the image.
 
-```dockerfile
-FROM centos:8
-RUN yum install -y zip
-```
+**Note**, that one and only one of `--sdk-local`, `--sdk-path` and `--sdk-url` options should be specified for Tarantool Enterprise.
 
-Of course, an opensource Tarantool version will be installed to the image,
-if required.
+###### Image tag
 
 The image is tagged as follows:
+
 * `<name>:<detected_version>`: by default;
 * `<name>:<version>`: if the `--version` parameter is specified;
 * `<tag>`: if the `--tag` parameter is specified.
@@ -334,13 +323,63 @@ The image is tagged as follows:
 `<name>` can be specified in the `--name` parameter, otherwise it will be
 auto-detected from the application rockspec.
 
-For Tarantool Enterprise, you should specify a download token using the
-`--download_token` parameter or the `TARANTOOL_DOWNLOAD_TOKEN` environment
-variable. It's needed to download the SDK to the resulting image.
+###### Tarantool Enterprise SDK
+
+If you use Tarantool Enterprise, you should explicitly specify Tarantool SDK to
+be delivered on the result image.
+If you want to use your current SDK version, just pass `--sdk-local` flag to
+`cartridge pack docker` command.
+You can specify local path to the other SDK using `--sdk-path` option,
+or the URL to download it using `--sdk-url` option
+(or env `TARANTOOL_SDK_DOWNLOAD_URL`).
+
+###### Build and runtime images
+
+In fact, two images are created - build image and runtime image.
+Build image is used to perform application build.
+Then, application files are delivered to the runtime image (that is exactly the
+result of running `cartridge pack docker`).
+
+Both images are created from `centos:8`.
+On build image all packages required for the default  `cartridge` application build
+(`git`, `gcc`, `make`, `cmake`, `unzip`) are installed.
+On the runtime image opensource Tarantool is installed if required.
+
+If your application requires some other applications for build or runtime, you
+can specify base layers for build and runtime images:
+
+* build image: `Dockerfile.build.cartridge` or `--build-from`;
+* runtime image: `Dockerfile.cartridge` or `--from`.
+
+The base image dockerfile should be started with the `FROM centos:8` line
+(except comments).
+
+For example, if your application requires `gcc-c++` for build and `zip` for
+runtime:
+
+`Dockerfile.cartridge.build`:
+
+```dockerfile
+FROM centos:8
+RUN yum install -y gcc-c++
+# Note, that git, gcc, make, cmake, unzip packages
+# will be installed anyway
+```
+
+`Dockerfile.cartridge`:
+
+```dockerfile
+FROM centos:8
+RUN yum install -y zip
+```
+
+###### Building the app
 
 If you want the `docker build` command to be run with custom arguments, you can
 specify them using the `TARANTOOL_DOCKER_BUILD_ARGS` environment variable.
 For example, `TARANTOOL_DOCKER_BUILD_ARGS='--no-cache --quiet'`
+
+###### Using the result image
 
 The application code will be placed in the `/usr/share/tarantool/${app_name}`
 directory. An opensource version of Tarantool will be installed to the image.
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -295,53 +295,92 @@ across all sections of the YAML file(s) stored in `/etc/tarantool/conf.d/*`.
...
 auto-detected from the application rockspec.
 
-For Tarantool Enterprise, you should specify a download token using the
-`--download_token` parameter or the `TARANTOOL_DOWNLOAD_TOKEN` environment
-variable. It's needed to download the SDK to the resulting image.
+###### Tarantool Enterprise SDK
+
+If you use Tarantool Enterprise, you should explicitly specify Tarantool SDK to
+be delivered on the result image.
+If you want to use your current SDK version, just pass `--sdk-local` flag to
```
olegrok: > If you want to use your current SDK version

from your local machine

cartridge-cli.lua
```diff
@@ -2925,6 +2961,7 @@ local function pack_rpm(opts)
 
     body = lead .. buf_pad_to_8_byte_boundary(signature_header) .. body
 
+    info('Write RPM file')
```
olegrok: I'm not sure that user wants to see such details, let's move it to verbose

cartridge-cli.lua
```diff
@@ -3424,15 +3497,54 @@ function cmd_pack.callback(args)
...
+    local sdk_path
     if app_state.tarantool_is_enterprise and app_state.build_in_docker then
-        if app_state.download_token == nil or app_state.download_token == '' then
-            die(
-                'Tarantool download token is required to pack enterprise Tarantool app in docker. ' ..
-                'Please, specify it using --download_token option or TARANTOOL_DOWNLOAD_TOKEN env variable'
-            )
+        -- check that passed one option for SDK
+        local sdk_params_are_right = check_that_only_one_is_true({
+            args.sdk_local,
```
olegrok: Consider a way to validate what you will download or copy.

You should not to copy MacOs binary inside CentOS
You should gracefully die if download url leads to something else except Tarantool EE.

cartridge-cli.lua
dokshina: It confuses a little when you see smth like `Pack in CPIO` and then nothing happens for a few minutes.
I think that messages like this 

```bash
Write application tmpfiles configuration
Create CPIO archive
163006 blocks
Compress it using GZIP
# here nothing happens for a few minutes
```
makes packing process more clear.

dokshina: I sure that we need more log levels, but it should be performed in the separated MR.

olegrok: Ok

cartridge-cli.lua
dokshina: I'm agree with you about dying for bad download URL, but I'm not sure that we should check binaries.

Option `--sdk-local` introduced to prevent using some default scenarios (like copying local SDK). The main idea was to shift responsibility for choosing right bundle to user.

I'm thinking about removing `--sdk-url` option in favor to using only `--sdk-path`. I think that it would be better if user downloads and unpacks SDK by himself.

olegrok: > we should check binaries

Should be enough to call `tarantool --version` and check that it doesn't contain "Darwin" and "FreeBSD" (not sure about this).

olegrok: > I'm thinking about removing --sdk-url option in favor to using only --sdk-path. I think that it would be better if user downloads and unpacks SDK by himself.

Sounds reasonable

test/python/test_pack.py
```diff
@@ -69,6 +76,9 @@ def deb_archive(tmpdir, light_project, request):
     cmd = [os.path.join(basepath, "cartridge"), "pack", "deb", project.path]
 
     if request.param == 'docker':
+        if project.deprecated_flow_is_used:
```
olegrok: I assume that this changes don't connected with current MR - move it to separate merge request, please

cartridge-cli.lua
```diff
@@ -3327,6 +3348,36 @@ local function remove_build_dir()
     end
 end
 
+local function detect_sdk_path(args)
+    -- check that passed one option for SDK
+    local sdk_params_are_right = check_that_only_one_is_true({
```
olegrok: Looks like an overkill - to write separate function to check that array of two booleans contains only one "true" value.

dokshina: It was written when there was 3 parameters.

I wouldn't write it for 2 parameters, but it's already written.

cartridge-cli.lua
```diff
@@ -3427,12 +3498,40 @@ function cmd_pack.callback(args)
...
+        if not fio.path.is_dir(sdk_path) then
+            die('Specified SDK path is not a directory: %s', sdk_path)
+        end
+
+        -- check that SDK directory contains tarantool and tarantoolctl binaries
+        -- and they both are executable
+        for _, binary in ipairs({'tarantool', 'tarantoolctl'}) do
+            local sdk_binary_path = fio.pathjoin(sdk_path, binary)
+            if not fio.path.exists(sdk_binary_path) then
+                die('Specified SDK directory (%s) does not contain %s binary',sdk_path, binary)
```
olegrok: `binary',sdk_path` -> `binary', sdk_path`

test/python/test_pack.py
```diff
@@ -447,28 +457,119 @@ def test_using_both_flows(project_without_dependencies, pack_format, tmpdir):
...
     env = os.environ.copy()
-    del env['TARANTOOL_DOWNLOAD_TOKEN']
+    del env['TARANTOOL_SDK_PATH']
 
-    cmd = [
-        os.path.join(basepath, "cartridge"),
-        "pack", pack_format,
-        "--use-docker",
-        project.path
+    # one and only one of this params should be specified
```
olegrok: Seems it should be a unit test, not integration (but it's only a recommendation)

dokshina: Yes, I've already written unit tests for this case.

test/unit/dockerfile_constructors_test.lua
```diff
@@ -34,20 +34,13 @@ local function check_output(command, ...)
...
-local function get_install_tarantool_enterprise_layers(sdk_version)
+local function get_copy_tarantool_enterprise_layers()
     local layers = remove_leading_spaces([=[
-        ### Install Tarantool Enterprise
-        ARG DOWNLOAD_TOKEN
-
-        RUN DOWNLOAD_URL=https://tarantool:${DOWNLOAD_TOKEN}@download.tarantool.io \
-            && mkdir -p /usr/share/tarantool \
-            && cd /usr/share/tarantool \
-            && curl -O -L ${DOWNLOAD_URL}/enterprise/tarantool-enterprise-bundle-SDK_VERSION.tar.gz \
```
olegrok: Could we remove a "curl" from Dockerfile after that?

test/python/test_pack.py
olegrok: But it has been already implemented https://github.com/tarantool/cartridge-cli/pull/166/files#diff-7baecb9cbb2a69a0b0c1b57742e201aeR17

test/unit/dockerfile_constructors_test.lua
dokshina: What did you mean?

test/python/test_pack.py
dokshina: I will remove this test from integration tests

test/unit/dockerfile_constructors_test.lua
olegrok: Is there a place inside Dockerfile there you install "curl"?

test/python/test_pack.py
dokshina: See #168

test/unit/dockerfile_constructors_test.lua
dokshina: No

olegrok: Ok, my question is not actual

cartridge-cli.lua
```diff
@@ -3427,12 +3498,40 @@ function cmd_pack.callback(args)
...
-        if app_state.download_token == nil or app_state.download_token == '' then
-            die(
-                'Tarantool download token is required to pack enterprise Tarantool app in docker. ' ..
-                'Please, specify it using --download_token option or TARANTOOL_DOWNLOAD_TOKEN env variable'
-            )
+        local sdk_path, err = detect_sdk_path(args)
+        if sdk_path == nil then die(err) end
+
+        -- check that specified path is an existent directory
+        if not fio.path.exists(sdk_path) then
```
olegrok: You can omit this check. In case then directory doesn't exist second check 'is_dir' returns false

dokshina: I think that `Path doesn't exists` error is much better then `Path isn't a directory`.

PR #180, opened 2020-03-20
Compilation into one executable file

dokshina:
Before this patch the Cartridge CLI was delivered as a `cartridge` executable
(that required `cartridge-cli` module and run it's `main` function)
and Lua files placed in `.rocks/share/tarantool` or `/usr/share/tarantool`.
As a result, on running `cartridge` command (from SDK or installed globally),
`cartridge-cli` module was loaded from current `.rocks` directory (and other 
unexpected places according to Tarantool modules loading policy)
and it confused a lot.

This patch changes the `cartridge-cli` build: now it's compiled into the one
`cartridge` executable that is delivered in a result package.


cartridge
```diff
@@ -1,13 +1,88 @@
...
+        end
+        if cli_filepath ~= nil then break end
+    end
+
+    if cli_filepath == nil then
+        return table.concat(errors, '\n')
+    end
+
+    return function()
+        return dofile(cli_filepath)
```
olegrok: Please consider to check several things:

- This file has correct lua sytnax (load(filename) should return correct result)
- This file is really cartridge-cli application

cartridge
```diff
@@ -1,13 +1,88 @@
...
+--
+-- When `require(<modname>)` is called, Lua searches modules this way
+-- (https://www.lua.org/manual/5.1/manual.html#pdf-package.loaders):
+-- - in a package.preload table;
+-- - looks across paths from package.path;
+-- - looks across paths from package.cpath;
+-- - ans so on...
+--
+-- But Tarantool changes this behaviour by inserting custom loaders in this table
+-- (https://github.com/tarantool/tarantool/blob/4e27b0e61747c5c6a8e7b9e53b18fa78f24ae61a/src/lua/init.lua#L215)
```
olegrok: Please file a documentation issue if this actually has not described.

cartridge
dokshina: I think that this loader behaviour should be the same as standard loaders have, so I will study how do standard loaders handle this situation and do the same. Are you OK with it?

Maybe it's enough to simply load `cartridge-cli` module?

olegrok: > Maybe it's enough to simply load cartridge-cli module?

Could you provide a correct and understandable for users message and test such behaviour?
But personally I don't want to see as you drop all predefined loaders.

> I will study how do standard loaders handle this situation and do the same

Agree. I like this way.

cartridge
```diff
@@ -1,17 +1,114 @@
...
+    if dir == '/' then return true end
+
+    if string.startswith(subdir, string.format('%s/', dir)) then
+        return true
+    end
+
+    return false
+end
+
+-- * ------------------------ Log messages ------------------------
```
olegrok: I think you could extract it in something like "console" utils

cartridge
olegrok: https://github.com/tarantool/doc/issues/1195

cartridge
```diff
@@ -1,17 +1,114 @@
...
+--
+-- But Tarantool changes this behaviour by inserting custom loaders in this table
+-- (https://github.com/tarantool/tarantool/blob/4e27b0e61747c5c6a8e7b9e53b18fa78f24ae61a/src/lua/init.lua#L215)
+-- In fact, there is some other loader before `package.preload`, so the
+-- package.path changes don't affect searching module.
+--
+-- But `cartridge-cli` uses it's own rules to load module:
+-- - (1) RPM od DEB package -> /usr/share/tarantool/cartridge-cli.lua
+-- - (2) local development  -> ${script_dir}/cartridge-cli.lua
+-- - (3) SDK                -> ${SDK}/.rocks/share/tarantool/cartridge-cli.lua
```
olegrok: Why is it not the first?

olegrok:
What's about a test?

CMakeLists.txt
```diff
@@ -16,51 +16,29 @@ enable_testing()
...
+  OUTPUT_STRIP_TRAILING_WHITESPACE
+  OUTPUT_VARIABLE TMPDIR
 )
 
-install(
-  DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}
-  DESTINATION ${TARANTOOL_INSTALL_LUADIR}/
-  OPTIONAL
+# Generate cartridge binary in tmpdir
+execute_process(
```
olegrok: Seems we need something similar to https://cmake.org/cmake/help/v3.12/module/FindPython.html (available since cmake 3.12). You require 2.8 and older.

But the problem - you have different python versions on different distros and should try to find python3, python2 or throw an error.

olegrok: try to copy-paste it https://github.com/Kitware/CMake/blob/master/Modules/FindPython/Support.cmake

cartridge-cli/utils.lua
```diff
@@ -0,0 +1,450 @@
...
+-- expand() allows to render a text template, expanding ${statement}
+-- into the calculated value of that statement.
+-- Roughly based on http://lua-users.org/wiki/TextTemplate
+--
+-- First argument is the template string, then arbitrary number of
+-- tables with mappings of variable=value
+function utils.expand(template, ...)
+    assert(type(template)=='string', 'expecting string')
+    local searchlist = {...}
+    local estring,evar
```
olegrok: `estring,evar` -> `estring, evar`

cartridge-cli/utils.lua
```diff
@@ -0,0 +1,450 @@
...
+    end
+
+    local function get(tab,index)
+        for _,symt in ipairs(searchlist) do
+            local ts = type(symt)
+            local value
+            if     ts == 'function' then value = symt(index)
+            elseif ts == 'table'
+            or ts == 'userdata' then value = symt[index]
+            else error'search item must be a function, table or userdata' end
```
olegrok: I think that this part of code should be rewritten using our styleguide

generate_binary.py
```diff
@@ -0,0 +1,114 @@
+#! /usr/bin/env python
```
olegrok: I'm not sure that it should called "binary".

May be "package"... or "executable"

dokshina: Maybe `compile_executable`?

CMakeLists.txt
dokshina: I think, I can require cmake 3.

cartridge-cli/utils.lua
dokshina: This function was copy-pasted from some resource (there is a comment and a link).
Maybe it should be "as is".

Anyway, it requires many edits, it can be performed in the next PR (for example, the one that will continue restructuring `cartridge-cli`).

olegrok: Ok. Agree

generate_binary.py
olegrok: Yes, it's better IMO

CMakeLists.txt
olegrok: > I think, I can require cmake 3.

No, because you supports CentOS 7

> It will be used on Gitlab CI, I can install any version I need there.

Yes, but my concern is about users who clone the repo and build/install cartridge-cli from source

CMakeLists.txt
```diff
@@ -16,51 +16,29 @@ enable_testing()
...
+  OUTPUT_VARIABLE TMPDIR
 )
 
-install(
-  DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}
-  DESTINATION ${TARANTOOL_INSTALL_LUADIR}/
-  OPTIONAL
+# Generate cartridge executable in tmpdir
+execute_process(
+  COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/compile_executable.lua
```
olegrok: You have a `findTarantool.cmake` script in `cmake` directory. I think you should find tarantool and run this script using it.

compile_executable.lua
```diff
@@ -0,0 +1,210 @@
...
+    {'source', 'string'},
+    {'dest', 'string'},
+    {'module', 'string'},
+})
+
+assert(args.source ~= nil, "{lease, specify source directory (--source)")
+assert(args.dest ~= nil, "{lease, specify destination file name (--dest)")
+assert(args.module ~= nil, "{lease, specify module name (--module)")
+
+compile_executable(args.source, args.dest, args.module)
```
olegrok: One day we should carefully test this script

compile_executable.lua
```diff
@@ -0,0 +1,210 @@
...
+
```
olegrok: Seems that it's a redundant line

olegrok:
Please, check my comments and let's push it ASAP

CMakeLists.txt
dokshina: It's already called https://github.com/tarantool/cartridge-cli/pull/180/files#diff-af3b638bc2a3e6c650974192a53c7291L12

compile_executable.lua
dokshina: Yes, but now I want to add a simple test that `tarantoolctl rocks make` creates working module.

cartridge-cli/utils.lua
dokshina: The same as below - in a separated patch.

CMakeLists.txt
olegrok: I mean you should use `COMMAND ${TARANTOOL_INSTALL_BINDIR}/tarantool ${CMAKE_CURRENT_SOURCE_DIR}/compile_executable.lua` or smth similar.

Main idea - to specify tarantool binary directly as you've already found it.

PR #185, opened 2020-03-26
Use compiled executable in tests

dokshina:
Compile cartridge executable and use it as a fixture in all tests

### DIFF
--- a/test/helper.lua
+++ b/test/helper.lua
@@ -1,5 +1,6 @@
 jit.off()
-package.path = package.path .. ";./?;?"
+
+local fio = require('fio')
 
 -- box.NULL, custom and cdata errors aware assert
 function assert(val, message, ...) -- luacheck: no global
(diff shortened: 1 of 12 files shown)


test/python/project.py
```diff
@@ -51,13 +51,13 @@ def is_removed_on_pack(filename):
 # Class Project
 ###############
 class Project:
-    def __init__(self, name, basepath, template='cartridge'):
+    def __init__(self, cartridge_cmd, name, basepath, template='cartridge'):
```
olegrok: Could we eliminate passing this fixture in each test if you've already used it in "Project" class? Seems we could export it as Project.cmd or smth similar.

dokshina: I'm not sure that `cartridge_cmd` is a part of Project.

Passing this fixture in each tests is clear and doesn't confuse.

olegrok: IMO If you have a "Project" you should have a way to run it

dokshina: Yes, but `cartridge_cmd` doesn't belong to a project.

olegrok: But you've already used it to build. I don't see reasons why it should stop us from using to run.

dokshina: I sure that `project.cartridge_cmd` will confuse.
And I don't see any problems in passing this fixture, this command is the main object of testing, why shouldn't we pass it explicitly?

olegrok: ok

test/helper.lua
```diff
@@ -1,5 +1,6 @@
 jit.off()
```
olegrok: wow

Why?

PR #197, opened 2020-04-10
It adds SOURCERPM tag

Lebensgefahr:
Would you be so kind to add some additional RPM tags to align RPMs packages produced by cartridge-cli with the native builded RPMs. This will prevent errors in rpm processing of third party software like mc, genpkglist (altlinux repo tools) etc. Packages builded with the native rpmbuild are not causing this kind of errors 
Why don't you use rpmbuild for rpm packages building? For example it strips debuginfo symbols from ELF files you are not stipping, it can use xz or gzip for different size files you can't.
Missing SOURCERPM tag causes non error exit code of genpkglist so it is not possible to create the repository with packages builded by cartridge-cli.
Bear in mind that missing of some tags in rpm spec file causes non error exit code of rpmbuild. You can use rpmlint to check produced rpm.
And please add the BuildArch for the package name.

### DIFF
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -1832,6 +1832,7 @@ local HEADER_TAG_TABLE = {
     SIZE = 1009,
     OS = 1021,
     ARCH = 1022,
+    SOURCERPM = 1044,
     PAYLOADFORMAT = 1124,
     LICENSE = 1014,
     GROUP = 1016,
@@ -2265,6 +2266,7 @@ local function pack_rpm(opts)
         {'GROUP', 'STRING', 'None'},
         {'OS', 'STRING', 'linux'},
         {'ARCH', 'STRING', 'x86_64'},
+        {'SOURCERPM', 'STRING', ''}
         {'PAYLOADCOMPRESSOR', 'STRING', 'gzip'},
         {'PAYLOADFLAGS', 'STRING', '5'},
         {'PREIN', 'STRING', create_user_script_rpm},
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -2266,7 +2266,7 @@ local function pack_rpm(opts)
         {'GROUP', 'STRING', 'None'},
         {'OS', 'STRING', 'linux'},
         {'ARCH', 'STRING', 'x86_64'},
-        {'SOURCERPM', 'STRING', ''}
+        {'SOURCERPM', 'STRING', ''},
         {'PAYLOADCOMPRESSOR', 'STRING', 'gzip'},
         {'PAYLOADFLAGS', 'STRING', '5'},
         {'PREIN', 'STRING', create_user_script_rpm},


dokshina:
Thanks for your patch and feedback!

We don't use rpmbuild for packing because one of `cartridge-cli` use-cases is using it in production environment without access to the Internet and root privileges to install rpmbuild.
But you still can use rpmbuild to pack your application. 

I agree that we should add missed tags to prevent `genpkglist` error.
Is `SOURCERPM` the only problem? Or there are other flags required by `genpkglist`?

BTW, `rpmlint` doesn't show `SOURCERPM flag missed` error.

Lebensgefahr:
> ild for packing be
Is it not possible to use binaries with its linked libraries compiled for the target arch in the production environment?
For genpkglist it is enough to add this missed tag. But it will be better to have exactly the same tags as rpmbuild builded package has to prevent issues in the future.
I know about rpmlint, but it shows a lot of other errors.

dokshina:
What binaries do you mean?

Please, fix syntax error in your patch:

`cartridge-cli.lua:2270:9: expected '}' (to close '{' on line 2258) near '{'`

You can use [tarantool/luacheck](https://github.com/tarantool/luacheck) to check if everything is OK.

dokshina:
I've figured out that installing RPM with SOURCERPM tag causes non-zero exit code on centos.
So, this PR cannot be merged "as is".

Lebensgefahr:
> I've figured out that installing RPM with SOURCERPM tag causes non-zero exit code on centos.
> So, this PR cannot be merged "as is".

Are you sure that the reason of the non-zero code is SOURCERPM tag? I have checked installation of the cartridge generated package with SOURCERPM tag and there is not any error except missed dependencies which I have resolved with --nodeps.

dokshina:
Yes, I've checked one more time:

(after rebasing on master to use `--suffix` option):

```bash
# in the cartridge-cli project root
./cartridge create --name myapp
./cartridge pack rpm myapp --version 1.0.0-0 --suffix after
# the same for cartridge-cli master
git checkout master
./cartridge pack rpm myapp --version 1.0.0-0 --suffix before
```

It produces `myapp-1.0.0-0-after.rpm` and `myapp-1.0.0-0-before.rpm` files.
Then, I copy these files on centos:7 docker image and try to install:

```
[root@6ebd2ff1a641 /]# yum install -y /opt/myapp-1.0.0-0-after.rpm
Loaded plugins: fastestmirror, ovl
Examining /opt/myapp-1.0.0-0-after.rpm: myapp-1.0.0-0.src
Cannot add package /opt/myapp-1.0.0-0-after.rpm to transaction. Not a compatible architecture: src
```

It fails for `myapp-1.0.0-0-after.rpm`, but works for `yum install -y /opt/myapp-1.0.0-0-before.rpm`.

How did you create packages?

Lebensgefahr:
I've got the same error, earlier I used rpm to install produced package. Sorry for spending your time.

Issue #187, opened 2020-03-30
Elaborate deploying stateboard.lua


rosik:
Latest cartridge versions introduce new entity - state provider for stateful failover (aka `stateboard.lua`). We should elaborate deploying it as conveniently as the other cluster instances.

PR #207, opened 2020-04-15
Deliver app stateboard in RPM/DEB

dokshina:
This is the first part of #187 (see also https://github.com/tarantool/cartridge/issues/714)

Add `stateboard.init.lua` to template
Use `cartridge == scm-1`
Add stateboard unit file in RPM/DEB systemd directory
Test that this file is created

### DIFF
--- a/README.md
+++ b/README.md
@@ -566,6 +566,21 @@ systemctl start myapp@instance-1
 systemctl start myapp@instance-2
 ```
 
+If you use stateful failover, you need to start application stateboard.
+Add `myapp-stateboard` section to `/etc/tarantool/conf.d/myapp.yml`:
+
+```yaml
+myapp-stateboard:
+  listen: localhost:3310
+  password: passwd
+```
+
+Then, start stateboard service:
+
+```bash
+systemctl start myapp-stateboard
+```
+
 #### Package details
 
 The installed package name will be `<name>` no matter what the artifact name is.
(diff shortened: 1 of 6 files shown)


cartridge-cli.lua
```diff
@@ -681,7 +683,43 @@ TimeoutStopSec=10s
...
+Environment=TARANTOOL_APP_NAME=${stateboard_name}
+Environment=TARANTOOL_WORKDIR=${stateboard_workdir}
+Environment=TARANTOOL_CFG=/etc/tarantool/conf.d/
+Environment=TARANTOOL_PID_FILE=/var/run/tarantool/${stateboard_name}.pid
+Environment=TARANTOOL_CONSOLE_SOCK=/var/run/tarantool/${stateboard_name}.control
+
+LimitCORE=infinity
+# Disable OOM killer
+OOMScoreAdjust=-1000
+# Increase fd limit for Vinyl
```
olegrok, line 712: We don't use vinyl for stateboard

README.md
```diff
@@ -566,6 +566,21 @@ systemctl start myapp@instance-1
 systemctl start myapp@instance-2
 ```
 
+If you use stateful failover, you need to start application stateboard.
```
olegrok, line 569: I think that `stateboard` should be wrapped in braces `"`. I think you should be quite experienced user to understand/know what stateboard is.

Also may be we should paraphrase `If you use stateful failover` to `If you plan/want to use stateful failover`.

PR #208, opened 2020-04-16
Prettify `cartridge start/stop` logs

### LINKED ISSUE #213: Prettify start/stop logs
Now `start` and `stop` commands use `assert` for failing on errors.
Since we use coloured logs for other commands, we can use it for `start` and `stop` commands as well.

dokshina:
Closes #213

### DIFF
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -171,6 +171,7 @@ local ERROR_COLOR_CODE = '\x1B[31m' -- red
 local WARN_COLOR_CODE = '\x1B[33m' -- yellow
 local INFO_COLOR_CODE = '\x1B[36m' -- cyan
 local DEBUG_COLOR_CODE = '\x1B[35m' -- magenta
+local OK_COLOR_CODE = '\x1B[32m' -- green
 
 -- Map of `log_level_letter => color_code`.
 local COLOR_CODE_BY_LOG_LEVEL = fun.iter({
@@ -3468,7 +3469,7 @@ local function read_configuration(path)
 end
 
 -- Read configuration at `path` and fetch instance names.
-local function get_configured_isntances(path, app_name)
+local function get_configured_instances(path, app_name)
     local config, err = read_configuration(path)
     if config == nil then return nil, err end
 
@@ -3515,7 +3516,7 @@ local function start_one(args)
 end
 
 local function start_all(args)
-    local instance_names, err = get_configured_isntances(args.cfg, args.app_name)
+    local instance_names, err = get_configured_instances(args.cfg, args.app_name)
     if instance_names == nil then return nil, err end
 
     for _, instance_name in pairs(instance_names) do
@@ -3777,7 +3778,6 @@ local function stop_one(args)
     assert(args.instance_name ~= nil)
 
     cmd_start.finalize_args(args)
-    info('Stopping %s...', args.instance_name)
 
     local pid_file = args.pid_file
     if fio.stat(pid_file) == nil then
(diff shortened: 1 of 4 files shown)


cartridge-cli.lua
```diff
@@ -472,33 +466,31 @@ local function detect_git_version(source_dir)
...
-
-local function detect_name(source_dir)
-    local rockspec_filename, err = find_rockspec(source_dir)
-    if rockspec_filename == nil then return nil, err end
 
-    local rockspec_filepath = fio.pathjoin(source_dir, rockspec_filename)
-    local rockspec, err = utils.load_variables_from_file(rockspec_filepath)
-    if rockspec == nil then
-        return nil, string.format('Failed to load rockspec %s: %s', rockspec_filepath, err)
+    local rockspecs = fio.glob(fio.pathjoin(dir, '*-scm-1.rockspec'))
```
olegrok, line 474: why scm only?

olegrok: `string.endswith(file, '.rockspec')` was before. I could use `app-1.1.0-1.rockspec` - it is not scm

cartridge-cli.lua
```diff
@@ -472,33 +466,31 @@ local function detect_git_version(source_dir)
...
+    if #rockspecs == 1 then
+        return rockspecs[1]
     end
 
-    local name = rockspec.package
-    if name == nil then
-        return nil, string.format("Rockspec %s doesn't contain required field 'package'", rockspec_filepath)
+    if #rockspecs > 1 then
+        return nil, string.format('Found multiple rockspecs in %s', dir)
```
olegrok, line 480: Could I specify directly rockspec/version that I need?

dokshina: You can specify appname explicitly. In fact, you have to do it, if appname can't be detected from rockspec.

cartridge-cli.lua
dokshina: I agree, I will use `*.rockspec`pattern and fetch package name from `package` sprcified in rockspec.

cartridge-cli.lua
```diff
@@ -3735,10 +3805,46 @@ function cmd_stop.callback(args)
...
     if fio.stat(pid_file) then
-        assert(fio.unlink(pid_file))
+        if not fio.unlink(pid_file) then
+            warn('Failed to remove pid file (%s)', pid_file)
+        end
     end
     if fio.stat(args.console_sock) then
-        assert(fio.unlink(args.console_sock))
+        if not fio.unlink(args.console_sock) then
+            warn('Failed to remove console sock (%s)', args.console_sock)
```
olegrok, line 3814: `sock -> socket`

cartridge-cli.lua
```diff
@@ -3735,10 +3805,46 @@ function cmd_stop.callback(args)
...
+end
+
+local function stop_all(args)
+    local instance_names = get_configured_isntances(args.cfg, args.app_name)
+    for _, instance_name in pairs(instance_names) do
+        local instance_args = table.copy(args)
+        instance_args.instance_name = instance_name
+        local ok, err = stop_one(instance_args)
+        if not ok then
+            return nil, string.format('Failed to stop %s: %s', instance_name, err)
```
olegrok, line 3835: I'm not sure that we should do it a such way.

We should call "stop" for each instance and in case of errors we should accumulate error messages. E.g. we can't stop first instance, but we stop second and third and print error.
```
[instance_name_1] failed to stop: error_message1
[instance_name_4] failed to stop: error_message4
```

cartridge-cli.lua
```diff
@@ -3505,82 +3590,88 @@ function Process:initialize()
...
-    execve(arg[-1], {self.script}, self.env) -- stops execution
-end
-
 function Process:build_notify_socket()
-    local sock = assert(socket('AF_UNIX', 'SOCK_DGRAM', 0), 'Can not create socket')
+    local sock = socket('AF_UNIX', 'SOCK_DGRAM', 0)
+    if sock == nil then return nil, 'Cannot create notify socket' end
+
     local basename = self.app_name .. '.' .. self.instance_name .. '-notify.sock'
     local sock_name = fio.pathjoin(self.run_dir, basename)
```
olegrok, line 3609: Please, verify socket length - https://github.com/tarantool/tarantool/issues/4634

dokshina: For notify socket: https://github.com/tarantool/cartridge-cli/issues/214
For instance socket: https://github.com/tarantool/cartridge/issues/790

cartridge-cli.lua
```diff
@@ -192,30 +192,25 @@ end
...
 local function info(fmt, ...) -- luacheck: no unused
     local msg = string.format(fmt, ...)
-    print_and_flush(colored_msg(msg, INFO_COLOR_CODE))
+    log.info(colored_msg(msg, INFO_COLOR_CODE))
 end
 
 local function debug(fmt, ...) -- luacheck: no unused
     local msg = string.format(fmt, ...)
-    print_and_flush(colored_msg(msg, DEBUG_COLOR_CODE))
+    log.info(colored_msg(msg, DEBUG_COLOR_CODE))
```
olegrok, line 213: debug -> info? But it's ok because by default we can't log in debug mode without box.cfg

PR #219, opened 2020-04-20
Check stateboard init script

dokshina:


test/python/test_pack.py
```diff
@@ -725,3 +733,50 @@ def test_builddir_is_removed(cartridge_cmd, project_without_dependencies, pack_f
...
+    os.makedirs(extract_dir)
+
+    if pack_format == 'rpm':
+        extract_rpm(archive_path, extract_dir)
+    elif pack_format == 'deb':
+        extract_deb(archive_path, extract_dir)
+        with tarfile.open(name=os.path.join(extract_dir, 'data.tar.xz')) as data_arch:
+            data_arch.extractall(path=extract_dir)
+
+    # check that stateboard unit file wasn't delivered
```
olegrok, line 775: Are there some tests where stateboard is delivered?

dokshina: Yes, see `check_systemd_dir` function in `utils.py`

PR #220, opened 2020-04-20
Start stateboard only

dokshina:


cartridge-cli.lua
```diff
@@ -3591,25 +3600,6 @@ end
...
-    for _, instance_name in pairs(instance_names) do
-        local instance_args = table.copy(args)
-        instance_args.instance_name = instance_name
-
-        local process_args = get_process_args(instance_args)
-        table.insert(processes, process_args)
-    end
-
     -- stateboard
     if args.stateboard then
```
olegrok, line 3604: `--with-stateboard`

dokshina: `args.stateboard = args.stateboard or args.stateboard_only`, this flag will be set to prevent missing `args.stateboard or args.stateboard_only` everywhere in the code.

olegrok: This has nothing to do with my remark. However, as well as my remark to your patch.

PR #224, opened 2020-04-21
Warn on starting without build

dokshina:
Warn if `.rocks` directory does not exists on `cartridge start`


cartridge-cli.lua
```diff
@@ -3746,6 +3748,12 @@ function cmd_start.callback(args)
         end
     end
 
+    if not fio.path.exists(fio.pathjoin(args.app_dir, '.rocks')) then
```
olegrok, line 3751: is_dir?

PR #228, opened 2020-04-23
Do not save archives data in lua memory

### LINKED ISSUE #230: `not enough memory` error
Running `cartridge pack` causes `not enough memory` error.

dokshina:
Closes #230 

Archive data was stored in a Lua string. 
Now all archives data is stored in a files.

### DIFF
--- a/cartridge-cli.lua
+++ b/cartridge-cli.lua
@@ -1766,16 +1766,14 @@ local function pack_tgz()
     if not ok then return false, err end
 
     info('Create archive')
-    local data, err = check_output(
-        "cd %s && %s -cvzf - %s",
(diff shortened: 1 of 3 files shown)


cartridge-cli.lua
```diff
@@ -2103,6 +2101,46 @@ local function filter_out_known_files(files)
...
+    if md5sum == nil then
+        return nil, 'md5sum binary not found'
+    end
+
+    local output = check_output('%s %s', md5sum, fullpath)
+    if output == nil then
+        return nil, string.format('Failed to get %s MD5 using md5sum', fullpath)
+    end
+
+    local md5_hex = output:strip():split()[1]
```
olegrok, line 2120: split(1)

PR #232, opened 2020-04-24
Fix error handling in RPM pack

dokshina:


cartridge-cli.lua
```diff
@@ -451,7 +456,10 @@ local function detect_git_version(source_dir)
         return nil
     end
 
-    local raw_version, err = check_output('cd "%s" && git describe --tags --long', source_dir)
+    local raw_version, err = check_output(
```
olegrok, line 459: check_output/call/... should call escape internally. You can't escape every argument by hands. In future it's easy to miss it.

dokshina: In future we will use Go.

cartridge-cli.lua
```diff
@@ -147,6 +147,11 @@ local function globtopattern(g)
     return p
 end
 
+local function escape_path(filepath)
```
olegrok, line 150: Add a comment please why we need it

PR #241, opened 2020-05-12
Rewrite in Go: Create

dokshina:


commands/version.go
```diff
@@ -0,0 +1,24 @@
...
+	rootCmd.AddCommand(versionCmd)
+}
+
+var versionCmd = &cobra.Command{
+	Use:   "version",
+	Short: "Print the version number of Cartridge CLI",
+	Long:  `All software has versions. This is Cartridge CLI's`,
+	Args:  cobra.MaximumNArgs(0),
+	Run: func(cmd *cobra.Command, args []string) {
+		initLogger()
```
olegrok, line 19: I've found 2 mentions of "initLogger" function (excluding definition).

But seems that we should have only one entry point for that.

commands/cartridge.go
```diff
@@ -0,0 +1,48 @@
...
+
+	log "github.com/sirupsen/logrus"
+	"github.com/tarantool/cartridge-cli/project"
+
+	"github.com/spf13/cobra"
+)
+
+var (
+	projectCtx project.ProjectCtx
+	rootCmd    = &cobra.Command{
```
olegrok, line 14: Broken spaces between "rootCmd" and "="

commands/cartridge.go
```diff
@@ -0,0 +1,48 @@
...
+
+func init() {
+	rootCmd.PersistentFlags().BoolVar(&verboseLogLevel, "verbose", false, "Verbose output")
+
+	initLogger()
+}
+
+func Execute() {
+	if err := rootCmd.Execute(); err != nil {
+		log.Fatalf(err.Error())
```
olegrok, line 30: I think one day we should do as it was done it 1.x versions - pretty error message with message that it's an error and place where it could be reported.

commands/create.go
```diff
@@ -0,0 +1,101 @@
...
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/tarantool/cartridge-cli/common"
+	"github.com/tarantool/cartridge-cli/create"
+	"github.com/tarantool/cartridge-cli/project"
+	"github.com/tarantool/cartridge-cli/templates"
```
olegrok, line 12: Why templates dir is on top-level. I assume that we use templates only in "create". It should be placed inside "create" directory.

commands/create.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// set current directory as a default path
+	if projectCtx.BasePath == "" {
+		projectCtx.Path, err = os.Getwd()
+		if err != nil {
+			return err
+		}
+	}
+
+	// check parent path
```
olegrok, line 77: Seems it's not "normalization", but "validation"

olegrok: I propose to move it in separate function.

commands/version.go
```diff
@@ -0,0 +1,24 @@
...
+
+		// TODO: improve getting version
+		fmt.Println("Cartridge CLI v2.0.0")
```
olegrok, line 22: As preparation step you could move "2.0.0" into global variable that could be redefined  at compile time.

common/utils.go
```diff
@@ -0,0 +1,20 @@
+package common
+
+import "fmt"
+
+func Prompt(text, defaultValue string) string {
```
olegrok, line 5: Consider this:
https://github.com/c-bata/go-prompt
https://github.com/stromland/cobra-prompt

create/create.go
```diff
@@ -0,0 +1,139 @@
...
+	log.Infof("Creating an application %q...", projectCtx.Name)
+
+	// check that application doesn't exist
+	if _, err := os.Stat(projectCtx.Path); err == nil {
+		return fmt.Errorf("Application already exists in %s", projectCtx.Path)
+	}
+
+	var err error
+
+	err = os.Mkdir(projectCtx.Path, 0755)
```
olegrok, line 30: Above you use one-liner "if". Here you don't. It's a bit inconsistent. I think we should support uniformity where it's possible

create/create.go
```diff
@@ -0,0 +1,139 @@
...
+	log.Infof("Application %q created successfully", projectCtx.Name)
+
+	return nil
+}
+
+func initGitRepo(projectCtx project.ProjectCtx) error {
+	var err error
+
+	// check that git is installed
+	if _, err = exec.LookPath("git"); err != nil {
```
olegrok, line 59: Does "git" that you use library supports such checks?

create/create.go
```diff
@@ -0,0 +1,139 @@
...
+	userSignature, err := repo.DefaultSignature()
+	if err != nil {
+		return fmt.Errorf(`failed to get default git user: %s.
+Please, call
+
+    git config --global user.name "Your Name"
+    git config --global user.email you@example.com
+
+to set user`,
+			err)
```
olegrok, line 81: Do something this tabs here.

Also you could move such message into separate variable.

templates/build_files.go
```diff
@@ -0,0 +1,90 @@
...
+# rm -rf doc
+`,
+	}
+
+	buildDockerfile := fileTemplate{
+		Path: "Dockerfile.build.cartridge",
+		Mode: 0644,
+		Content: `# Simple Dockerfile
+# Used by "pack" command as a base for build image
+# when --use-dcoker option is specified
```
olegrok, line 48: `dcoker ` -> `docker`.

Also please fix it in master. I think you simply copy-pasted this text

templates/dev_files.go
```diff
@@ -0,0 +1,83 @@
...
+			Mode: 0755,
+		},
+	}
+}
+
+func getCartridgeDevFiles() []fileTemplate {
+	depsScript := fileTemplate{
+		Path: "deps.sh",
+		Mode: 0755,
+		Content: `#!/bin/sh
```
olegrok, line 21: I propose to stop inline such texts and move it into variables (and may be files).

templates/templates.go
```diff
@@ -0,0 +1,159 @@
...
+		return fmt.Errorf("Failed to create directory %s: %s", d.Path, err)
+	}
+
+	return nil
+}
+
+func getTemplatedStr(text string, obj interface{}) (string, error) {
+	tmpl, err := template.New("path").Parse(text)
+	if err != nil {
+		return "", err
```
olegrok, line 149: May be we should return a pointer to string.

`"", err` looks ugly.

commands/version.go
dokshina: Yes, thx.

commands/cartridge.go
dokshina: This is the result of running `go fmt`. I've added a space between definitions and now everything looks fine.

commands/cartridge.go
dokshina: Yes, I sure that we need to handle internal errors and format them. 
But here is handled runtime errors, such as non-existent directory and so on.

commands/create.go
dokshina: Yes, I'm planning to rewrite this code.

create/create.go
dokshina: I didn't find it in a documentation, but maybe it's possible.

templates/build_files.go
dokshina: I don't want to fix it in a master - this isn't a reason to tag a new `1.8` version.

common/utils.go
dokshina: I'm not sure that we need to use something complex to perform just one prompt.

commands/create.go
```diff
@@ -0,0 +1,111 @@
...
+
+	var err error
+
+	// fill create-specific context
+	if err := fillCreateCtx(&projectCtx); err != nil {
+		return err
+	}
+
+	// get project path
+	basePath := cmd.Flags().Arg(0)
```
olegrok, line 57: why not explicit cwd?

commands/create.go
```diff
@@ -0,0 +1,111 @@
...
+	projectCtx.Path, err = getProjectPath(basePath)
+	if err != nil {
+		return err
+	}
+
+	// fill project-specific context
+	if err := project.FillCtx(&projectCtx); err != nil {
```
olegrok, line 64: I think we should do it not in create but at upper level

commands/create.go
```diff
@@ -0,0 +1,111 @@
...
+		projectCtx.Name = common.Prompt("Enter project name", "myapp")
+	}
+
+	return nil
+}
+
+func getProjectPath(basePath string) (string, error) {
+	var err error
+
+	if basePath == "" {
```
olegrok, line 88: But you've already done it here https://github.com/tarantool/cartridge-cli/pull/241/files#diff-d979e9548d0d641ce071ee98608da5faR57?

commands/create.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	basePath, err = filepath.Abs(basePath)
+	if err != nil {
+		return "", fmt.Errorf("Failed to normalize args: %s", err)
+	}
+
+	// check base path
+	fileInfo, err := os.Stat(basePath)
+	if os.IsNotExist(err) {
```
olegrok, line 102: And you ignore another error types

commands/create.go
dokshina: Because the current version has this argument.

commands/create.go
```diff
@@ -0,0 +1,111 @@
...
+func fillCreateCtx(projectCtx *project.ProjectCtx) error {
+	// prompt name if not specified
+	if projectCtx.Name == "" {
+		projectCtx.Name = common.Prompt("Enter project name", "myapp")
+	}
+
+	return nil
+}
+
+func getProjectPath(basePath string) (string, error) {
```
olegrok, line 85: `nil, error` or `"", error`?

commands/create.go
dokshina: Where? This function should be called after filling command-specific context.

For example, `create` command gets a project name by flag, `pack` describes it from rockspec (if it isn't specified by flag). This is command-specific context filling, only after that we can compute stateboard name.

create/create.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/tarantool/cartridge-cli/create/templates"
+	"github.com/tarantool/cartridge-cli/project"
+)
+
+func CreateProject(projectCtx *project.ProjectCtx) error {
+	log.Infof("Creating an application %q...", projectCtx.Name)
+
+	// check that application doesn't exist
+	if _, err := os.Stat(projectCtx.Path); err == nil {
```
olegrok, line 17: That if you e.g. don't have enough permission to open this dir? It returns an error but you ignore it

commands/create.go
dokshina: No, that was getting command argument.

create/git.go
```diff
@@ -0,0 +1,66 @@
...
+			return fmt.Errorf("Failed to initialize git repo")
+		}
+	}
+
+	log.Debug("Initialized git repo")
+
+	// add files to index
+	addCmd := exec.Command("git", "add", "-A")
+	if err := common.RunCommand(addCmd, projectCtx.Path, false); err != nil {
+		if err != nil {
```
olegrok, line 38: You've already checked that error is not nil. At previous line.

dokshina: Yes, and then I copy pasted it about 6 times...

commands/create.go
dokshina: Decided to use `"", error` in the whole project.

commands/version.go
dokshina: Yes, I will improve getting version in a separated PR.

PR #244, opened 2020-05-13
Rewrite in Go: Build

dokshina:


build_project/local.go
```diff
@@ -0,0 +1,66 @@
...
+		if err != nil {
+			return fmt.Errorf("Failed to run pre-build hook: %s", err)
+		}
+	}
+
+	// tarantoolctl rocks make
+	if rockspec, err := common.FindRockspec(projectCtx.BuildDir); err != nil {
+		return err
+	} else if rockspec != "" {
+		rocksMakeCmd := exec.Command("tarantoolctl", "rocks", "make")
```
olegrok, line 30: offtop: it's  quite ironic. We rewrite it in go to have independent binary file. Which in fact strictly depends on "tarantoolctl" which depends on "tarantool".

olegrok: Seems that before run build (in fact each command) we should validate that such tools are available.

tarantoolctl, git, etc... I think user should know about dependency problem before we will do anything

common/utils.go
```diff
@@ -33,3 +36,61 @@ func RunCommand(cmd *exec.Cmd, dir string, showOutput bool) error {
...
+	fileInfo, err := os.Stat(path)
+	if err != nil {
+		return false, err
+	}
+
+	perm := fileInfo.Mode().Perm()
+	return perm&0100 != 0, nil
+}
+
+func FindRockspec(path string) (string, error) {
```
olegrok, line 80: pointer to string as we discussed before?

test/python/project.py
```diff
@@ -166,10 +166,14 @@ def add_dependency_submodule(project):
     # add submodule to rocks content
     project.rocks_content.add('.rocks/share/tarantool/rocks/{}'.format(SUBMODULE_NAME))
 
+    print(os.listdir(project.path))
```
olegrok, line 169: Debug

build_project/local.go
```diff
@@ -0,0 +1,66 @@
...
+	"path/filepath"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/tarantool/cartridge-cli/common"
+	"github.com/tarantool/cartridge-cli/project"
+)
+
+func buildProjectLocally(projectCtx project.ProjectCtx) error {
+	// pre-build
```
olegrok, line 16: But where is post-build hooks?

build_project/local.go
dokshina: Yes, but `tarantool` isn't required for creating a project, or building in docker....

build_project/local.go
dokshina: Post build hook runs only on packing (it removes some files that shouldn't be delivered in a result artifact, but we shouldn't remove anything on building in the current directory).

build_project/local.go
olegrok: Yes. It doesn't contradict with the fact that we should check tarantool existence as early as possible when it needed.

.gitlab-ci.yml
```diff
@@ -105,8 +105,8 @@ test_enterprise-1.10:
     - source tmp/sdk-1.10/env.sh
     - tarantool -V
     - make ci_prepare
-    - make tmp/cache-image.tar
-    - docker load -i tmp/cache-image.tar
+    # - make tmp/cache-image.tar
```
olegrok, line 108: I suggest drop it or uncomment if your tests fail on the intermediate state it's ok

build_project/build.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	log.Infof("Application build succeeded")
+
+	return nil
+}
+
+func checkCtx(projectCtx *project.ProjectCtx) error {
+	if projectCtx.Path == "" {
+		return fmt.Errorf("missed project path")
```
olegrok, line 59: Some errors start with a capital letter, others don't. I think it should be uniformly.

build_project/local.go
```diff
@@ -0,0 +1,55 @@
...
+	rocksMakeCmd := exec.Command("tarantoolctl", "rocks", "make")
+	err := common.RunCommand(rocksMakeCmd, projectCtx.BuildDir, !projectCtx.Quiet)
+	if err != nil {
+		return fmt.Errorf("Failed to install rocks: %s", err)
+	}
+
+	return nil
+}
+
+func checkLocalBuildRecommendedBinaries() {
```
olegrok, line 40: tarantool/tarantoolctl?

.gitlab-ci.yml
dokshina: Nothing fails, this code is redundant since there is no build in docker implemented. It just takes a lot of time in CI.

build_project/local.go
```diff
@@ -0,0 +1,55 @@
...
+	var recommendedBinaries = []string{
```
olegrok, line 41: So, I think we should check also "required" binaries here.

dokshina: Yes, I'm thinking about doing that in `project.FillCtx` function because it's called for every command.

commands/build.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	log "github.com/sirupsen/logrus"
+	build "github.com/tarantool/cartridge-cli/build_project"
+	"github.com/tarantool/cartridge-cli/project"
+
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	rootCmd.AddCommand(buildCmd)
```
olegrok, line 12: Is it described in some guidelines?

Alternative variant is export command description and add it directly in main

commands/create.go
```diff
@@ -66,7 +66,7 @@ func runCreateCommand(cmd *cobra.Command, args []string) error {
 	}
 
 	// create project
-	if err := create.CreateProject(&projectCtx); err != nil {
```
olegrok, line 69: IMO `create.CreateProject` looks better than `create.Create`

commands/build.go
dokshina: In the cobra README: https://github.com/spf13/cobra#create-additional-commands

test/python/test_build.py
```diff
@@ -94,3 +65,71 @@ def test_files_with_bad_symbols(cartridge_cmd, project_without_dependencies, tmp
     ]
     process = subprocess.run(cmd, cwd=project.path)
     assert process.returncode == 0, 'Building project failed'
+
+
+def test_app_without_rockspec(cartridge_cmd, project_without_dependencies):
```
olegrok, line 70: Could you push this tests to upstream?

.gitlab-ci.yml
olegrok: Ok. Stay it as it is.

build_project/local.go
olegrok: Do it somewhere. But we should add some validation before run every command. It makes us a bit user-friendly

commands/build.go
olegrok: Thanks for explanation

test/python/test_build.py
dokshina: No, since now application build works without rockspec and hooks can be non-executable. It's an incompatible changes.

commands/create.go
dokshina: Decided to use `<command>.Run`.

build_project/local.go
dokshina: Added `project.CheckTarantoolBinaries()`, on build it causes an erro, on create just warning.

project/project.go
```diff
@@ -59,3 +60,20 @@ func FillCtx(projectCtx *ProjectCtx) error {
...
+func CheckTarantoolBinaries() error {
+	var requiredBinaries = []string{
+		"tarantool",
+		"tarantoolctl",
+	}
+
+	// check recommended binaries
+	for _, binary := range requiredBinaries {
+		if _, err := exec.LookPath(binary); err != nil {
+			return fmt.Errorf("missed %s binary", binary)
```
olegrok, line 74: `missed ` -> `Missed `

PR #246, opened 2020-05-15
Add a spinner on quiet run

dokshina:


common/execute.go
```diff
@@ -6,19 +6,56 @@ import (
...
+	if !showOutput {
+		s = spinner.New(spinner.CharSets[9], 100*time.Millisecond)
+		s.Start()
+	}
+
+	wg.Add(1)
+	go startAndWaitCommand(cmd, &wg, &err)
+	wg.Wait()
+
+	if s != nil {
```
olegrok, line 54: I don't thing that such check is really needed

common/execute.go
```diff
@@ -6,19 +6,56 @@ import (
...
 
 	if showOutput {
 		cmd.Stdout = os.Stdout
 		cmd.Stderr = os.Stderr
 	}
 
-	return cmd.Run()
+	var s *spinner.Spinner
+
+	if !showOutput {
```
olegrok, line 45: That about complete generalization of such things

Potentially you could have several running processes simultaneously. And I think that all commands could be running in separate goroutine.

And addition of such thing allows you run scheduler that run several tasks. In your case "run commands" and "spinner". I think such approach makes your code better. Without ugly pointers and several IFs

common/execute.go
```diff
@@ -6,19 +6,75 @@ import (
...
+var (
+	spinnerPicture    = spinner.CharSets[9]
+	spinnerUpdateTime = 100 * time.Millisecond
 )
 
+const (
+	ready = 1
+)
+
+// startAndWaitCommand starts and waits command
```
olegrok, line 24: Are you serious?

common/execute.go
```diff
@@ -6,19 +6,75 @@ import (
...
+
+	if *err = cmd.Start(); *err != nil {
+		return
+	}
+
+	if *err = cmd.Wait(); *err != nil {
+		return
+	}
+
+	err = nil
```
olegrok, line 38: Do you really need it?

PR #242, opened 2020-05-12
Completely rewrite in Go

dokshina:
Closes #106

This PR is separated to the next parts:

- [x] Tests
  - [x] Gitlab CI
  - [x] Rewrite unit tests in Go
  - [x] e2e tests
    - [x] write e2e tests using python docker SDK #273 
  - [x] Rewrite integration tests in Pytest
    - [x] start,stop, status tests #270 
    - [x] version test #270 
- [x] Reworks
  - [x] Store command output in a file #277 
  - [x] Allow to use `centos:7` as a base image #287 
  - [x] Read config from .cartridge.yml #279 
  - [x] improve internal errors handling #283 
- [x] Command `version`
  - [x] implement command #241 
  - [x] improve version compution #268 
- [x] Command `create`:
  - [x] implement command #241 
- [x] Command `build`
  - [x] implement command #244 
  - [x] build in docker #252 
- [x] Command `pack`
  - [x] tgz #247
  - [x] rpm #251
  - [x] deb #248
  - [x] docker #261 
  - [x] additional options for build in docker #266 
- [x] Command `start` #264
- [x] Wait on notify socket #275 
- [x] Command `stop`  #264
- [x] Command `status`  #264
- [x] Prettify logs #286 
- [x] Update the doc #291 
- [x] Set up GitLab CI to build packages #285 
- [x] Set up GitLab CI to push packages #297 

Incompatible changes:

- deprecated build flow is removed;
- build requires rockspec in the application directory;
- pre- and post-build hooks should be executable
- unit templates variables format changed
- removed TARANTOOL_DOCKER_BUILD_ARGS
- added `--data-dir` option for start instead of specifying `workdir` in a project config.
- removed `--apps-dir` option and changed other options default values
- Renamed CARTRIDGE_BUILDDIR to CARTRIDGE_TEMPDIR
- `cartridge start` expects only instance names, or no arguments to start instansec from conf

PR #247, opened 2020-05-15
Rewrite in go: Pack TGZ

dokshina:


Makefile
```diff
@@ -13,6 +13,7 @@ tmp/cache-image.tar:
 	docker save -o tmp/cache-image.tar cache-image
 
 .PHONY: lint
+lint:
```
olegrok, line 16: I think such changes could be extracted and pushed directly to matser.

build_project/build.go
```diff
@@ -22,26 +22,24 @@ func Run(projectCtx *project.ProjectCtx) error {
...
+		return fmt.Errorf("Unable to build application in %s: it's not a directory", projectCtx.BuildDir)
+	}
+
+	log.Infof("Building application in %s...", projectCtx.BuildDir)
+
 	// check that application directory contains rockspec
-	if rockspec, err := common.FindRockspec(projectCtx.Path); err != nil {
+	if rockspecPath, err := common.FindRockspec(projectCtx.Path); err != nil {
 		return err
-	} else if rockspec == "" {
```
olegrok, line 44: I suggest you to consider absence of rockspec as an error. Then additional `else if == ""` is not needed.

build_project/post_build.go
```diff
@@ -0,0 +1,40 @@
...
+	log "github.com/sirupsen/logrus"
+	"github.com/tarantool/cartridge-cli/common"
+	"github.com/tarantool/cartridge-cli/project"
+)
+
+func PostRun(projectCtx *project.ProjectCtx) error {
+	// post-build
+	postBuildHookPath := filepath.Join(projectCtx.BuildDir, postBuildHookName)
+
+	if _, err := os.Stat(postBuildHookPath); !os.IsNotExist(err) {
```
olegrok, line 17: But it could be another error type. Permission denied or smth else.

I think we should run hook only if stat doesn't return error.

commands/cartridge.go
```diff
@@ -45,4 +46,8 @@ func setLogLevel() {
 	if projectCtx.Verbose {
 		log.SetLevel(log.DebugLevel)
 	}
+
+	if projectCtx.Debug {
```
olegrok, line 50: Hm... Is it possible to specify `--verbose --debug`?

I think if you want to support several log levels you should add `--log-level=debug/verbose`

commands/pack.go
```diff
@@ -0,0 +1,84 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+		err := runPackCommand(cmd, args)
+		if err != nil {
+			log.Fatalf(err.Error())
+		}
+	},
+}
+
+func runPackCommand(cmd *cobra.Command, args []string) error {
+	setLogLevel()
```
olegrok, line 39: log level is general property for the whole application. Why do you still place it at start of thus/each command?

commands/pack.go
```diff
@@ -0,0 +1,84 @@
...
+
+	var err error
+
+	projectCtx.PackType = cmd.Flags().Arg(0)
```
olegrok, line 43: Any validation for type/path/...?

common/utils.go
```diff
@@ -53,6 +64,28 @@ func TarantoolIsEnterprise(tarantoolDir string) (bool, error) {
...
+
+	tarantool := filepath.Join(tarantoolDir, "tarantool")
+	versionCmd := exec.Command(tarantool, "--version")
+
+	tarantoolVersion, err := GetOutput(versionCmd, nil)
+	if err != nil {
+		return "", err
+	}
+
+	r := regexp.MustCompile(`\d+\.\d+\.\d+-\d+-\w+`)
```
olegrok, line 79: If you plan to call this function more the one time at application execution, it could be reasonable to compile regexp into global var (not for every call). But feel free to ignore. Anyway it shouldn't lead to dramatic performance degradation for user application.

So you also could make variable name more expressive.

common/utils.go
```diff
@@ -83,3 +116,136 @@ func FindRockspec(path string) (string, error) {
...
+}
+
+// IsGitProject checks if specified path is a git project
+func IsGitProject(path string) bool {
+	fileInfo, err := os.Stat(filepath.Join(path, ".git"))
+	return err == nil && fileInfo.IsDir()
+}
+
+// HasPerm checks if specified file has permissions
+func HasPerm(fileInfo os.FileInfo, perm os.FileMode) bool {
```
olegrok, line 187: HasPermissions?

common/utils.go
```diff
@@ -83,3 +116,136 @@ func FindRockspec(path string) (string, error) {
...
+	return fileInfo.Mode()&perm == perm
```
olegrok, line 188: `fileInfo.Mode()&perm` -> `fileInfo.Mode() & perm`?

pack/app_dir.go
```diff
@@ -0,0 +1,269 @@
...
+		}
+	} else {
+		tarantoolVersionLine := fmt.Sprintf("TARANTOOL=%s", projectCtx.TarantoolVersion)
+		versionFileLines = append(versionFileLines, tarantoolVersionLine)
+	}
+
+	// rocks versions
+	rocksVersionsMap, err := getRocksVersions(appDirPath, projectCtx)
+	if err != nil {
+		log.Warnf("can't process rocks manifest file. Dependency information can't be "+
```
olegrok, line 178: `can't` -> `Can't`

pack/app_dir.go
```diff
@@ -0,0 +1,269 @@
...
+							"Found multiple versions for %s dependency in rocks manifest",
+							depName,
+						)
+					}
+					rocksVersionsMap[depName] = depVersion
+				})
+			}
+		})
+
+	} else if !os.IsNotExist(err) {
```
olegrok, line 240: And handle other errors in else section

project/project.go
```diff
@@ -6,6 +6,8 @@ import (
 	"os/exec"
 	"path/filepath"
 
+	lua "github.com/yuin/gopher-lua"
```
olegrok, line 9: I suggest to place "lua" logic into separate helper and require "lua" package only in such place.

Makefile
dokshina: I'm not sure that removing `luacheck` should be in a master. I've broke Makefile in the `completely-rewrite-in-go`.

build_project/build.go
dokshina: I'm not sure that `common.FindRockspec` should make decision if rockspec should exists or not.

Makefile
olegrok: I don't see that you remove "luacheck" here

commands/pack.go
dokshina: Because it depends on parsing command arguments result.

dokshina: Maybe it's possible to add smth like `pre-run` for all commands, I will research.

commands/pack.go
dokshina: In the `pack` directly.

common/utils.go
dokshina: `go` formatter thinks that it should be `fileInfo.Mode()&perm`...

pack/app_dir.go
dokshina: Again...

Makefile
dokshina: I've removed it in the patch that broke Makefile.

common/utils.go
dokshina: `perm` word is used in the `os` module doc, I've decided that it's OK.

olegrok: Ok

commands/cartridge.go
dokshina: I'm sure that using `--debug` is more convenient than `--log-level debug`.

Now there are three log level flags:

* `--verbose` prints additional messages (`log.DebugLevel`), it makes running commands more **verbose for user**;
* `--debug` is used for debug (`log.TraceLevel`). It doesn't remove tmp build directory and makes running commands more **verbose for developer**;
* `--quiet` is used to silence build command output (for example, `tarantoolctl rocks make` output), that makes commands more **convenient to use**.

For example, I use `--quiet --debug` to keep tmp build directory and silence build commands on packing debug.

If `--verbose --debug` is specified, the `--debug` level will be used because it includes `--verbose`.

pack/app_dir.go
dokshina: But stop, I handle all other errors here and ignore manifest file absence.

commands/cartridge.go
olegrok: `--verbose --debug`
I'm not sure that this combination is appropriate.

pack/app_dir.go
olegrok: Oh, yes, my bad.

commands/cartridge.go
dokshina: It can be a mistake, but I don't see anything bad here - this two modes don't exclude each other.

project/project.go
dokshina: Moved into `common`.

commands/pack.go
dokshina: Moved into root command `PersistentPreRun`.

commands/cartridge.go
olegrok: It contains an implicit logic. One option affect another. I'm not sure that's good idea to allow enable both at one time.

build_project/post_build.go
```diff
@@ -0,0 +1,42 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/tarantool/cartridge-cli/common"
+	"github.com/tarantool/cartridge-cli/project"
+)
+
+func PostRun(projectCtx *project.ProjectCtx) error {
```
olegrok, line 13: PostBuildRun?

dokshina: It's called as `build.PostRun`. Now there are `build.Run` and `build.PostRun`. Isn't it OK?

commands/build.go
```diff
@@ -38,6 +36,8 @@ func runBuildCommand(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	projectCtx.BuildDir = projectCtx.Path
```
olegrok, line 39: Could we specify it near `projectCtx.Path = cmd.Flags().Arg(0)`
Looks like:
```
var1 = ...
some_magic_with_var_1_maybe()
var2 = var1
```

Was `var1` changed or not? It's not clear.

commands/cartridge.go
olegrok: Decided do it later

common/lua.go
```diff
@@ -0,0 +1,85 @@
...
+
+	log "github.com/sirupsen/logrus"
+	lua "github.com/yuin/gopher-lua"
+)
+
+const (
+	rocksManifestPath = ".rocks/share/tarantool/rocks/manifest"
+)
+
+// LuaReadStringVar reads string variable from specified Lua file
```
olegrok, line 16: `global string variables`?

olegrok: Ok... varName is specified. `variable`

commands/cartridge.go
dokshina: Added it to the checklist in #242

common/lua.go
```diff
@@ -0,0 +1,85 @@
...
+
+		depsLTable.ForEach(func(depNameL lua.LValue, depInfoL lua.LValue) {
+			depName := depNameL.String()
+
+			depInfoLTable, ok := depInfoL.(*lua.LTable)
+			if !ok {
+				log.Warnf("Failed to get %s dependency info", depName)
+			} else {
+				depInfoLTable.ForEach(func(depVersionL lua.LValue, _ lua.LValue) {
+					depVersion := depVersionL.String()
```
olegrok, line 68: May be "number"?

Or "String" casts to string?

pack/pack.go
```diff
@@ -0,0 +1,117 @@
...
+		return fmt.Errorf("Unsupported distribution type: %s", projectCtx.PackType)
+	}
+
+	if _, err := os.Stat(projectCtx.Path); err != nil {
+		return fmt.Errorf("Failed to use path %s: %s", projectCtx.Path, err)
+	}
+
+	checkPackRecommendedBinaries()
+
+	projectCtx.BuildID = common.RandomString(10)
```
olegrok, line 47: Could we simply use UUID?

common/lua.go
dokshina: As I got from doc, it's something like Lua `tostring`.

olegrok: good

build_project/post_build.go
olegrok: May be ok...

pack/pack.go
dokshina: Decided not to do it.

commands/build.go
dokshina: It's filled in `project.FillCtx` function. It should be obvious that it changes context.

PR #251, opened 2020-05-22
Rewrite in Go: Pack RPM

dokshina:


src/common/utils.go
```diff
@@ -296,8 +300,44 @@ func WriteTgzArchive(srcDirPath string, destFilePath string) error {
...
+	destFile, err := os.Create(destFilePath)
+	if err != nil {
+		return fmt.Errorf("Failed to create result GZIP file %s: %s", destFilePath, err)
+	}
+
+	defer destFile.Close()
+
+	// dest file GZIP writer
+	gzipWriter, err := gzip.NewWriterLevel(destFile, gzip.BestCompression)
+	defer gzipWriter.Flush()
```
olegrok, line 325: Should be after error check

src/common/utils.go
```diff
@@ -308,3 +348,108 @@ func GetNextMajorVersion(version string) (string, error) {
...
+	if _, err := io.Copy(hasher, file); err != nil {
+		return "", err
+	}
+
+	return fmt.Sprintf("%x", hasher.Sum(nil)), nil
+}
+
+// FileMD5Hex computes MD5 for a given file.
+// The result is returned in a hex form
+func FileMD5Hex(path string) (string, error) {
```
olegrok, line 388: This function I think should use `FileMD5`. They actually do similar things.

src/pack/rpm.go
```diff
@@ -0,0 +1,68 @@
...
+		panic(err)
+	}
+
+	// app dir
+	appDirPath := filepath.Join(projectCtx.PackageFilesDir, "/usr/share/tarantool/", projectCtx.Name)
+	if err := initAppDir(appDirPath, projectCtx); err != nil {
+		return err
+	}
+
+	// systemd dir
```
olegrok, line 26: I think a comments are redundant - the name `initSystemdDir `, `initTmpfilesDir ` say it.

src/pack/rpm.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	log.Infof("Created result package: %s", projectCtx.ResPackagePath)
+
+	return nil
+}
+
+func checkPackRpmCtx(projectCtx *project.ProjectCtx) error {
+	if projectCtx.Version == "" {
+		return fmt.Errorf("Missed project version")
```
olegrok, line 48: It doesn't look like a statement. I think "Project version is missed". But may be I'm wrong - ask somebody more competent.

olegrok: It's relevant for all messages below.

src/rpm/common.go
```diff
@@ -0,0 +1,281 @@
...
+
+		packed.Count = len(byteArray)
+		for _, byteValue := range byteArray {
+			if _, err := io.Copy(packed.Data, packValues(byteValue)); err != nil {
+				return nil, err
+			}
+		}
+	case rpmTypeStringArray: // STRING_ARRAY
+		// value should be strings array
+		stringsArray, ok := tag.Value.([]string)
```
olegrok, line 67: Seems if you are in "rpmTypeStringArray" you have a guarantee that it could be casted to `[]string`. `!ok` check is redundant.

olegrok: BTW could it be rewritten as `tag.Value.(rpmTypeStringArray)`?

src/rpm/common.go
```diff
@@ -0,0 +1,281 @@
...
+
+		packed.Count = len(int32Values)
+
+		for _, value := range int32Values {
+			if _, err := io.Copy(packed.Data, packValues(value)); err != nil {
+				return nil, err
+			}
+		}
+
+	case rpmTypeInt64: // INT64
```
olegrok, line 139: Are all this types really used?

src/rpm/const.go
```diff
@@ -0,0 +1,124 @@
...
+	headerSignatures = 62
+	headerImmutable  = 63
+
+	hashAlgoSHA256 = 8
+
+	// XXX
+	fileFlag = 1 << 4
+	dirFlag  = 0
+
+	rpmTypeNull        = 0
```
olegrok, line 19: May be iota could/should be used here

src/rpm/const.go
```diff
@@ -0,0 +1,124 @@
...
+	rpmSensePrereq       = 0x40
+	rpmSenseInterp       = 0x100
+	rpmSenseScriptPre    = 0x200
+	rpmSenseScriptPost   = 0x400
+	rpmSenseScriptPreun  = 0x800
+	rpmSenseScriptPostun = 0x1000
+)
+
+var (
+	headerMagic   = []byte{0x8e, 0xad, 0xe8}
```
olegrok, line 86: Why not `const`?

src/rpm/header.go
```diff
@@ -0,0 +1,180 @@
...
+
+		filesInfo.BaseNames = append(filesInfo.BaseNames, filepath.Base(relPath))
+		filesInfo.FileMtimes = append(filesInfo.FileMtimes, int32(fileInfo.ModTime().Unix()))
+
+		filesInfo.FileUserNames = append(filesInfo.FileUserNames, defaultFileUser)
+		filesInfo.FileGroupNames = append(filesInfo.FileGroupNames, defaultFileGroup)
+		filesInfo.FileLangs = append(filesInfo.FileLangs, defaultFileLang)
+		filesInfo.FileLinkTos = append(filesInfo.FileLinkTos, defaultFileLinkTo)
+
+		sysFileInfo, ok := fileInfo.Sys().(*syscall.Stat_t)
```
olegrok, line 156: `syscall.Stat_t` does it really needed?

Aren't any "high-level" functions for that?

src/rpm/header.go
```diff
@@ -0,0 +1,180 @@
...
+		filesInfo.FileModes = append(filesInfo.FileModes, int16(sysFileInfo.Mode))
+		filesInfo.FileInodes = append(filesInfo.FileInodes, int32(sysFileInfo.Ino))
+		filesInfo.FileDevices = append(filesInfo.FileDevices, int32(sysFileInfo.Dev))
+		filesInfo.FileRdevs = append(filesInfo.FileRdevs, int16(sysFileInfo.Rdev))
+	}
+
+	return filesInfo, nil
+}
+
+func addDirAngGetIndex(dirNames *[]string, fileDir string) int {
```
olegrok, line 171: `Ang`? May be `And`?

src/rpm/const.go
dokshina: I don't use iota because this constants aren't defined by me, it defined by RPM format. 
I think it's more convenient to define such constants without using iota.

src/rpm/const.go
dokshina: Because constant can't be slice.

src/rpm/header.go
dokshina: I didn't find another way to get file inode.

src/rpm/common.go
dokshina: No, but I think it's better ho have a code that works for all types now.

src/pack/rpm.go
dokshina: Yes, I changed it to `Version is missed` etc.

src/rpm/common.go
dokshina: No (discussed)

src/common/execute.go
```diff
@@ -58,11 +58,17 @@ func RunCommand(cmd *exec.Cmd, dir string, showOutput bool) error {
...
 
+	var stdoutBuf bytes.Buffer
+	var stderrBuf bytes.Buffer
+
 	cmd.Dir = dir
 	if showOutput {
 		cmd.Stdout = os.Stdout
 		cmd.Stderr = os.Stderr
 	} else {
+		cmd.Stdout = &stdoutBuf
```
olegrok, line 69: Do you think that it's a good idea? Write a text (in any amount) to buffer? Output could be huge.

src/common/execute.go
```diff
@@ -72,6 +78,18 @@ func RunCommand(cmd *exec.Cmd, dir string, showOutput bool) error {
 
 	wg.Wait()
 
+	if err != nil {
+		if !showOutput {
+			return fmt.Errorf(
+				"Failed to run \n%s\n\nStderr: %s\n\nStdout: %s\n\n%s",
```
olegrok, line 84: In future I hope you will write it to the file.

src/common/utils.go
```diff
@@ -296,8 +300,43 @@ func WriteTgzArchive(srcDirPath string, destFilePath string) error {
...
+	// src file reader
+	srcFileReader, err := os.Open(srcFilePath)
+	if err != nil {
+		return fmt.Errorf("Failed to open source file %s: %s", srcFilePath, err)
+	}
+
+	defer srcFileReader.Close()
+
+	// dest file writer
+	destFile, err := os.Create(destFilePath)
```
olegrok, line 316: If file with such name already exists does it produce an error?

src/rpm/common.go
```diff
@@ -0,0 +1,298 @@
...
+		}
+
+		packed.Count = 1
+
+		bytedString := []byte(stringValue)
+		bytedString = append(bytedString, 0)
+		if _, err := io.Copy(packed.Data, packValues(bytedString)); err != nil {
+			return nil, err
+		}
+
```
olegrok, line 106: Since this line you change code style and add an empty line before each following "case".
Could you do it for the lines above? Or remove it below?

src/rpm/common.go
olegrok: The main problem of such approach - use have unused and untested code...

src/rpm/common.go
```diff
@@ -0,0 +1,298 @@
...
+	regionTagIndex := getPackedTagIndex(resData.Len(), regionTagID, rpmTypeBin, 16)
+
+	// resIndex is regionTagIndex + tagsIndex
+	if err := common.ConcatBuffers(resIndex, regionTagIndex, tagsIndex); err != nil {
+		return nil, err
+	}
+
+	// regionTag data
+	tagsNum := len(tagSet) + 1
+	regionTagData := getPackedTagIndex(-tagsNum*16, regionTagID, rpmTypeBin, 16)
```
olegrok, line 240: nit: `-tagsNum*16` -> `-tagsNum * 16`... codestyle...

src/rpm/common.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	return res, nil
+}
+
+func alignData(data *bytes.Buffer, boundaries int) {
+	dataLen := data.Len()
+
+	if dataLen%boundaries != 0 {
+		alignedDataLen := (dataLen/boundaries + 1) * boundaries
```
olegrok, line 263: nit: `dataLen/boundaries` -> `dataLen / boundaries`

src/rpm/const.go
```diff
@@ -0,0 +1,124 @@
...
+	versionMagic  = 0x01
+	reservedMagic = 0
+
+	knownFiles = map[string]struct{}{
```
olegrok, line 90: Will it work for Ubuntu, Centos, AltLinux,...? It should be carefully tested for the different platforms.

Also the name "knownFiles" tells me nothing

src/rpm/common_test.go
```diff
@@ -0,0 +1,181 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func hex(b *bytes.Buffer) string {
+	return fmt.Sprintf("%x", b.Bytes())
+}
+
+func TestPackTag(t *testing.T) {
```
olegrok, line 15: To be honest I don't understand a way how to obtain "tested" values...
May be you could add a some schema how every value is calculated... or maybe a link to doc/spec.

src/rpm/header.go
olegrok: Not sure, check please - https://stackoverflow.com/a/28339521

src/rpm/const.go
dokshina: Yes, it already works for all this systems. A part of this dirs is created by prein script (`"usr/share/tarantool` etc.).

I agree that naming can be better, what about `systemDirs` or `existingDirs`?

src/common/utils.go
dokshina: If file already exists, it will be truncated. I'm not sure that this function should make a decision about file existent.

olegrok: May be some flags should be passed

src/rpm/const.go
olegrok: `systemDirs `

src/common/execute.go
dokshina: Decided to store only last N bytes in such buffer.

I suggest doing in in a separated patch, added to checklist of #242

src/common/execute.go
dokshina: See comment above.

src/rpm/common.go
dokshina: Go formatter thinks it's OK.

src/rpm/common.go
dokshina: Go formatter thinks it's OK.

src/common/utils.go
dokshina: I don't want to make this function more difficult.

BTW, when you run `cartridge pack tgz` two times without changing a version, it rewrites existing TGZ archive with a new one.

src/rpm/common_test.go
dokshina: I've added a doc in `tagset.go` file.

src/rpm/header.go
dokshina: But it uses `syscall.Stat_t` too...

src/common/utils.go
olegrok: Ok - rewrite ok

src/pack/rpm.go
```diff
@@ -0,0 +1,69 @@
...
+
+	if projectCtx.ResPackagePath == "" {
+		return fmt.Errorf("ResPackagePath is missed")
+	}
+
+	if projectCtx.TmpDir == "" {
+		return fmt.Errorf("TmpDir is missed")
+	}
+
+	if projectCtx.PackageFilesDir == "" {
```
olegrok, line 64: May be we should check existence of some directories.
Before packaging anything.

src/rpm/header.go
olegrok: without ugly `(*syscall.Stat_t)` cast

olegrok: Hmmm... But ok... I've missed something

src/rpm/rpm.go
```diff
@@ -0,0 +1,142 @@
...
+
+	"github.com/tarantool/cartridge-cli/src/common"
+	"github.com/tarantool/cartridge-cli/src/project"
+)
+
+// Many thanks to @knazarov, who wrote packing in RPM in Lua a long time ago
+// This code can be found here
+// https://github.com/tarantool/cartridge-cli/blob/cafd75a5c8ddfdb93ef8290d6e4ebd5d83e4c46e/cartridge-cli.lua#L1814
+
+//  RPM file is a binary format, consisting of metadata in the form of
```
olegrok, line 17: `RPM file is a binary format`

File is not format.

RPM file in binary representation... Or smth similar...

src/pack/rpm.go
dokshina: This directory is initialized by `pack` module, so it's an internal error if it doesn't exists.

src/rpm/tagset.go
```diff
@@ -0,0 +1,384 @@
...
+}
+
+func (tagSet *rpmTagSetType) addTags(tags ...rpmTagType) {
+	*tagSet = append(*tagSet, tags...)
+}
+
+// Bless @knazarov for writing this code once,
+// I just rewrote it in Go.
+
+// See http://ftp.rpm.org/max-rpm/s1-rpm-file-format-rpm-file-format.html
```
olegrok, line 33: Seems it's better to use multiline comments according Tarantool Developers guide:
```go
/*
 *
 *
*/

https://www.tarantool.io/en/doc/2.3/dev_guide/c_style_guide/#commenting-style
It's for C, but anyway
```

src/rpm/rpm.go
dokshina: It seems to be copy-pasted from https://blog.packagecloud.io/eng/2014/10/20/yum-createrepo-generate-incorrect-metadata/

src/rpm/tagset.go
dokshina: Changed

src/pack/rpm.go
```diff
@@ -0,0 +1,69 @@
...
+	if projectCtx.Release == "" {
+		return fmt.Errorf("Release is missed")
+	}
+
+	if projectCtx.TarantoolVersion == "" {
+		return fmt.Errorf("TarantoolVersion is missed")
+	}
+
+	if projectCtx.ResPackagePath == "" {
+		return fmt.Errorf("ResPackagePath is missed")
```
olegrok, line 57: Does user know what is "ResPackagePath"?

dokshina: No, context check is an internal check that produces internal error **for developer**.

src/rpm/rpm.go
dokshina: fixed

PR #252, opened 2020-05-26
Rewrite in Go: Build in Docker

dokshina:


src/commands/pack.go
```diff
@@ -85,7 +95,25 @@ instantiated unit file
 Used for rpm and deb types
 `
 
-	stateboardUnitTemplateFlagDoc = `Path to the template for stateboard systemd unit file
+	stateboardUnitTemplateFlagDoc = `Path to the template for
+stateboard systemd unit file
 Used for rpm and deb types
+`
+
+	useDockerDoc = `forces to build the application in Docker`
```
olegrok, line 103: `forces ` -> `Forces `

src/commands/pack.go
```diff
@@ -85,7 +95,25 @@ instantiated unit file
...
+Used on build in docker
+Default to Dockerfile.build.cartridge
+`
+
+	sdkLocalDoc = `SDK from the local machine should be
+delivered in the result artifact
+Used for building in docker with Tarantool Enterprise
+`
+
+	sdkPathDoc = `path to the SDK to be delivered in the result artifact
```
olegrok, line 115: `path ` -> `Path `

src/common/utils.go
```diff
@@ -308,3 +307,36 @@ func GetNextMajorVersion(version string) (string, error) {
 
 	return strconv.Itoa(major + 1), nil
 }
+
+func GetTarantoolRepoVersion(version string) string {
```
olegrok, line 311: Does it consider that 2.1 is 2_x? However check it please, may be it has been already fixed.

src/docker/run.go
```diff
@@ -0,0 +1,132 @@
...
+	WorkingDir string
+	Cmd        []string
+
+	Volumes map[string]string
+
+	Quiet bool
+	Debug bool
+}
+
+func waitStartOutput(out io.ReadCloser, quiet bool) error {
```
olegrok, line 31: I think you should provide an interface that get function and flag and enable spinner if flag is true.

It's not good when you meet waitgroups/channels creations to run spinner over all project.

`run_in_background(myFunction, enableSpinner)`.

src/docker/run.go
```diff
@@ -0,0 +1,132 @@
...
+
+	defer func() {
+		if opts.Debug {
+			log.Warnf("Build container %s is not removed due to debug mode", containerID)
+			return
+		}
+
+		log.Infof("Remove build container...")
+		err := cli.ContainerRemove(ctx, containerID, types.ContainerRemoveOptions{
+			// RemoveVolumes: true,
```
olegrok, line 101: drop it if unused

src/docker/run.go
```diff
@@ -0,0 +1,132 @@
...
+		Cmd:        opts.Cmd,
+		WorkingDir: opts.WorkingDir,
+		Tty:        true,
+	}
+
+	hostConfig := container.HostConfig{
+		Binds: binds,
+	}
+
+	// create container
```
olegrok, line 85: Too obvious for "ContainerCreate" function

src/docker/run.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+	}()
+
+	// start container
+	if err := cli.ContainerStart(ctx, containerID, types.ContainerStartOptions{}); err != nil {
+		return fmt.Errorf("Failed to start container: %s", err)
+	}
+
+	out, err := cli.ContainerLogs(ctx, containerID, types.ContainerLogsOptions{
+		ShowStdout: true,
```
olegrok, line 114: Does it respect "quiet" mode?

src/docker/run.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	if err := waitStartOutput(out, opts.Quiet); err != nil {
+		return err
+	}
+
+	if code, err := cli.ContainerWait(ctx, containerID); err != nil {
+		return fmt.Errorf("Failed to run container: %s", err)
+	} else if code != 0 {
+		return fmt.Errorf("Failed to run command on container")
```
olegrok, line 128: Put error code here please

olegrok: If it's internal "docker" error code

src/project/project.go
```diff
@@ -24,17 +24,25 @@ type ProjectCtx struct {
...
 	TarantoolVersion      string
 	TarantoolIsEnterprise bool
 	WithStateboard        bool
 
+	BuildInDocker       bool
+	BuildFrom           string
+	BuildDockerfileBase string
+	SDKLocal            bool
+	SDKPath             string
+	BuildSDKDirame      string
```
olegrok, line 44: `BuildSDKDirame ` -> `BuildSDKDirname`?

src/docker/run.go
dokshina: Yes, it collects all logs, but shows them only when quiet mode isn't enabled.

src/docker/run.go
dokshina: Yes, I sure that all `spinner` magic should be placed in some common module.

Are you OK with doing it in a separated PR? I've added it to checklist of #242

olegrok:
Of course

src/docker/run.go
olegrok: Of course

src/common/utils.go
dokshina: Fixed to replace `2_1` to `2x`.

PR #257, opened 2020-05-28
Fix version patterns

dokshina:


test/python/test_pack.py
```diff
@@ -413,6 +413,31 @@ def test_result_filename_suffix(cartridge_cmd, project_without_dependencies, tmp
         assert expected_filename in os.listdir(tmpdir)
 
 
+@pytest.mark.parametrize('pack_format', ['tgz'])
+def test_invalid_version(cartridge_cmd, project_without_dependencies, tmpdir, pack_format):
+    project = project_without_dependencies
+
+    bad_versions = [
```
olegrok, line 420: You don't test "problem" situation when git stash has numbers at start and matched "%d" pattern. I think you could ass something like "123xxx"

cartridge-cli.lua
```diff
@@ -467,11 +467,13 @@ local function detect_git_version(source_dir)
         return nil
     end
 
+    raw_version = raw_version:gsub('\n$', '')
```
olegrok, line 470: Add a comment please

olegrok: why it is needed.

cartridge-cli.lua
```diff
@@ -420,12 +420,12 @@ end
 
 local function normalize_version(str)
     local patterns = {
-        "(%d+)%.(%d+)%.(%d+)-(%d+)-(%g+)",
-        "(%d+)%.(%d+)%.(%d+)-(%d+)",
-        "(%d+)%.(%d+)%.(%d+)-(%g+)",
-        "(%d+)%.(%d+)%.(%d+)",
-        "(%d+)%.(%d+)",
-        "(%d+)"
+        "^(%d+)%.(%d+)%.(%d+)%-(%d+)%-(g%g+)$",
```
olegrok, line 423: to be completely fair, git hash has constant length - (e.g. `g1b9031f2`) g + 8 symbols after

PR #261, opened 2020-05-29
Rewrite in Go: Pack Docker

dokshina:


Makefile
```diff
@@ -30,6 +30,7 @@ pytest:
 unit:
 	go test -v ./src/rpm
```
olegrok, line 31: Are there something like "run all"?

May be simple `go test -v`?

src/project/project.go
```diff
@@ -154,3 +159,15 @@ func detectName(path string) (string, error) {
 
 	return name, nil
 }
+
+func GetAppDir(projectCtx *ProjectCtx) string {
+	return filepath.Join("/usr/share/tarantool", projectCtx.Name)
```
olegrok, line 164: nit: seems it's a bit inconsistent `<path>/` and `<path>`.

I suggest you to add slash at the end of `/usr/share/tarantool` or drop for paths below.

test/python/test_pack.py
```diff
@@ -555,22 +554,6 @@ def test_project_without_build_dockerfile(cartridge_cmd, project_without_depende
     assert process.returncode == 0
 
 
-@pytest.mark.skip()
```
olegrok, line 558: Why do you drop tests?

dokshina: This test duplicates the same one from `test_pack_docker`

Makefile
dokshina: Yes, changed to `go test -v ./src/...`.

PR #263, opened 2020-06-01
Rewrite in Go: Runtime paths reworks

dokshina:


cli/pack/systemd_dir.go
```diff
@@ -125,32 +122,64 @@ func getSystemdTemplate(projectCtx *project.ProjectCtx) (templates.Template, err
...
 }
 
+func getSystemdCtx(projectCtx *project.ProjectCtx) *systemdCtx {
+	var ctx systemdCtx
+
+	ctx.Name = projectCtx.Name
+	ctx.StateboardName = projectCtx.StateboardName
+
+	ctx.DefaultWorkDir = project.GetInstanceWorkDir(projectCtx, "default")
+	ctx.InstanceWorkDir = project.GetInstanceWorkDir(projectCtx, "%i")
```
olegrok, line 139: What does "%i" mean here?

cli/pack/systemd_dir.go
```diff
@@ -125,32 +122,64 @@ func getSystemdTemplate(projectCtx *project.ProjectCtx) (templates.Template, err
...
+
+	ctx.ConfDir = projectCtx.ConfDir
+
+	ctx.AppEntrypointPath = project.GetAppEntrypointPath(projectCtx)
+	ctx.StateboardEntrypointPath = project.GetStateboardEntrypointPath(projectCtx)
+
+	if projectCtx.TarantoolIsEnterprise {
+		ctx.Tarantool = filepath.Join(projectCtx.AppDir, "tarantool")
+	} else {
+		ctx.Tarantool = "/usr/bin" // TODO
```
olegrok, line 158: Seems $PATH should be checked and $TARANTOOL_BIN/$TARANTOOL_DIR

olegrok: See https://github.com/tarantool/modulekit/blob/ckit/FindTarantool.cmake

cli/pack/systemd_dir.go
dokshina: It's `%i` from intantiated systemd unit. When you start `myapp@storage` instance, the `%i` is replaced with `storage`.

dokshina: I can use this value as a named constant and add a link to the systemd units doc.

cli/pack/systemd_dir.go
dokshina: This is the PATH where Tarantool is placed on the target platform. It's `/usr/bin/tarantool` for centos, alt and ubuntu, but it can be set to current system Tarantool path. 

I agree that we need some reworks here, but I suggest to perform it later (e.g. when we start to use `runtime.txt`).

cli/pack/systemd_dir.go
olegrok: Well, thanks for explanation

PR #264, opened 2020-06-03
Rewrite in Go: Start, Stop, Status

dokshina:


cli/project/files.go
```diff
@@ -47,18 +47,30 @@ func GetStateboardPidFile(projectCtx *ProjectCtx) string {
 
 func GetInstanceConsoleSock(projectCtx *ProjectCtx, instanceName string) string {
```
olegrok, line 48: Do you check that its length is less than 108 symbols?

cli/project/project.go
```diff
@@ -53,9 +56,9 @@ type ProjectCtx struct {
 	Entrypoint           string
 	StateboardEntrypoint string
 	AppDir               string
-	ConfDir              string
+	ConfPath             string
 	RunDir               string
-	WorkDir              string
+	WorkDirBase          string
```
olegrok, line 61: What does "base" mean here?

cli/running/common.go
```diff
@@ -0,0 +1,100 @@
...
+	"strings"
+
+	"github.com/tarantool/cartridge-cli/cli/common"
+	"gopkg.in/yaml.v2"
+
+	"github.com/tarantool/cartridge-cli/cli/project"
+)
+
+const (
+	DefaultLocalRunDir   = "tmp/run"
```
olegrok, line 16: Please use https://golang.org/pkg/os/#TempDir instead of `tmp`.

cli/pack/app_dir.go
```diff
@@ -35,7 +35,16 @@ func initAppDir(appDirPath string, projectCtx *project.ProjectCtx) error {
 				panic(err)
 			}
 
-			return relPath == ".rocks" || strings.HasPrefix(relPath, ".rocks/")
+			if relPath == ".rocks" || strings.HasPrefix(relPath, ".rocks/") {
+				return true
+			}
+
+			if _, err := os.Open(src); err != nil {
+				log.Warnf("Failed to copy: %s", err)
```
olegrok, line 43: `Failed to copy` what?

cli/running/common.go
```diff
@@ -0,0 +1,206 @@
...
+			process := NewInstanceProcess(projectCtx, instance)
+			processes.Add(process)
+		}
+	}
+
+	return &processes, nil
+}
+
+func formatEnv(key, value string) string {
+	return fmt.Sprintf("%s=%s", key, value)
```
olegrok, line 143: I don't sure that it's really needed but such approach doesn't support several words values?
E.g. `KEY=VALUE1 VALUE2` is in fact `KEY=VALUE1`, should be `KEY="VALUE1 VALUE2"`

cli/running/process.go
```diff
@@ -0,0 +1,261 @@
...
+	var err error
+
+	ctx := context.Background()
+	process.cmd = exec.CommandContext(ctx, "tarantool", process.entrypoint)
+
+	process.cmd.Env = append(os.Environ(), process.env...)
+	process.cmd.Dir = process.workDir
+
+	// create run dir
+	if err := os.MkdirAll(process.runDir, 0755); err != nil {
```
olegrok, line 135: Is executable bit really needed here? Why not `0644`?

cli/running/writer.go
```diff
@@ -0,0 +1,102 @@
...
+
+		// prefix
+		if _, err := os.Stdout.Write([]byte(w.prefix)); err != nil {
+			return n, err
+		}
+
+		// colorize line by log level
+		matches := logLineRgx.FindStringSubmatch(string(lineBytes))
+		if matches != nil {
+			logLevel := matches[1]
```
olegrok, line 64: and `len(matches) > 0`?

cli/pack/app_dir.go
dokshina: All `os` errors contain filepath, so it will be printed.

cli/project/files.go
dokshina: In the old cli we checked this only for notify socket that is used directly by `cartridge-cli`. For console sock this check will be performed by `cartridge`. It looks like a good approach.

cli/running/common.go
dokshina: I've just checked it - everything is OK with passing multi-words values. Using `KEY="VALUE1 VALUE2"` leads to setting `KEY` equal to `'"VALUE1 VALUE2"'` (with quotes).

olegrok: Well, could you add a test case for it?

dokshina: Decided that we shouldn't test `os/exec`, but if someone add quotes here, tests will fail.

cli/commands/stop.go
```diff
@@ -0,0 +1,54 @@
...
+
+	stopCmd.Flags().StringVar(&projectCtx.RunDir, "run-dir", "", runDirFlagDoc)
+	stopCmd.Flags().StringVar(&projectCtx.ConfPath, "cfg", "", cfgFlagDoc)
+
+	stopCmd.Flags().BoolVar(&projectCtx.WithStateboard, "stateboard", false, stateboardFlagDoc)
+	stopCmd.Flags().BoolVar(&projectCtx.StateboardOnly, "stateboard-only", false, stateboardOnlyFlagDoc)
+}
+
+var stopCmd = &cobra.Command{
+	Use: "stop [INSTANCE_NAME...] ",
```
olegrok, line 22: Is a space between `]` and `"` really needed.

PR #269, opened 2020-06-11
new metrics role in template project

ZolotarevIgor-mail-ru:
According to: https://github.com/tarantool/metrics/issues/73

New metrics role allows users to manage metrics with configuration files and enables default metrics on start of the app.


cartridge-cli/templates/cartridge/files/app_files.lua
```diff
@@ -14,6 +14,7 @@ local app_files = {
                 'lua >= 5.1',
                 'checks == 3.0.1-1',
                 'cartridge == 2.1.2-1',
+                'metrics == scm-1,'
```
olegrok, line 17: Let's await when feature will be completely released.

PR #270, opened 2020-06-11
Tests reworks

dokshina:
* Integration tests for `version`, `start`, `stop` and `status` 
  command are rewritten in pytest
* getting-started-app tests are rewritten in pytest
* python tests are restructured: integration tests are placed
  in `test/integration`, examples tests in `test/examples`


test/utils.py
```diff
@@ -0,0 +1,734 @@
+#!/usr/bin/python3
```
olegrok, line 1: It's actually redundant

test/integration/test_version.py
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/python3
```
olegrok, line 1: It's actually redundant

test/integration/test_running.py
```diff
@@ -0,0 +1,499 @@
+import os
```
olegrok, line 1: There is no shebang in such file and it's OK.

test/examples/test_getting_started.py
```diff
@@ -0,0 +1,173 @@
...
+    # get customer by wrong id
+    url = 'http://localhost:{}/storage/customers/{}'.format(router_http_port, CUSTOMER_ID+1)
+    r = requests.get(url, json=customer)
+    assert r.status_code == requests.status_codes.codes.NOT_FOUND
+    resp = r.json()
+    assert 'info' in resp
+    assert resp['info'] == 'Customer not found'
+
+    # update customer balance
+    # XXX: now I have no idea how to perform this call
```
olegrok, line 168: File an issue for each TODO

test/examples/test_getting_started.py
```diff
@@ -0,0 +1,173 @@
...
+    # it requires account_id field, but now I don't even
+    # understand how to add an account to customer
+
+    # stop
+    cli.terminate()
```
olegrok, line 173: In case of failed test you will never terminate it. `request.addfinalizer(lambda: cli.terminate())`

dokshina: See `Cli` constructor.

olegrok: Well, then I don't understand why we should call cli.terminate twice?

dokshina: I've removed `cli.terminate` here, but generally I think that it's OK to call it twice (`terminate` defined in constructor just prevents running cli process after test fail, but here is the test emulating user session).

test/examples/test_getting_started.py
dokshina: See #271

PR #275, opened 2020-06-18
Rewrite in Go: Use notify socket

dokshina:


test/utils.py
```diff
@@ -106,9 +130,11 @@ def __init__(self, cartridge_cmd):
 
         self._processes = []
         self._instance_pids = set()
+        self._subprocess = None
 
     def start(self, project, instances=[], daemonized=False, stateboard=False, stateboard_only=False,
-              cfg=None, script=None, run_dir=None, data_dir=None, log_dir=None):
+              cfg=None, script=None, run_dir=None, data_dir=None, log_dir=None,
+              capture_output=False, exp_rc=0):
```
olegrok, line 137: `exp_rc ` I understand that it means "return code"... But may be it's better to give directly "expected_return_code" name for this parameter.

Also seems this function accepts too many parameters. I can advice a reconsider it's structure for simplification.

cli/running/processes_set.go
```diff
@@ -11,10 +11,10 @@ import (
 type ProcessesSet []*Process
 
 const (
-	procOk procResType = iota + 10
-	procSkipped
-	procFailed
-	procExited
+	procResOk procResType = iota + 10
```
olegrok, line 14: Why `+10`?

dokshina: I've removed it.

test/utils.py
dokshina: Maybe sometimes...

PR #281, opened 2020-06-22
Rewrite in Go: Fill ctx reworks

dokshina:


cli/running/running.go
```diff
@@ -2,11 +2,48 @@ package running
...
+	}
+
+	if projectCtx.Name, err = project.DetectName(projectCtx.AppDir); err != nil {
+		return fmt.Errorf("Failed to detect application name: %s", err)
+	}
+
+	projectCtx.StateboardName = project.GetStateboardName(projectCtx)
+
+	if projectCtx.StateboardOnly {
+		projectCtx.WithStateboard = true
```
olegrok, line 33: Can we make `return nil` right after this line?

dokshina: I think it's a good approach to check specified instance ID's anyway and warn if they was specified with `--stateboard-only` flag.

PR #293, opened 2020-06-26
Rewrite in Go: Prettify log messages

dokshina:


cli/create/git.go
```diff
@@ -19,40 +19,32 @@ func initGitRepo(projectCtx *project.ProjectCtx) error {
...
 	initCmd := exec.Command("git", "init")
 	if err := common.RunCommand(initCmd, projectCtx.Path, false); err != nil {
-		return fmt.Errorf("Failed to initialize git repo")
+		return fmt.Errorf("Failed to initialize git repository")
 	}
 
-	log.Debug("Initialized git repo")
-
-	// add files to index
+	log.Debug("Add files to index")
```
olegrok, line 31: May be `Add files to git index`? Or add git prefix for such messages

cli/pack/app_dir.go
```diff
@@ -24,7 +24,7 @@ const (
 func initAppDir(appDirPath string, projectCtx *project.ProjectCtx) error {
 	var err error
 
-	log.Infof("Create application dir: %s", appDirPath)
+	log.Infof("Initialize application dir")
```
olegrok, line 27: Why did you drop `appDirPath `?

cli/pack/deb.go
```diff
@@ -63,21 +64,21 @@ func packDeb(projectCtx *project.ProjectCtx) error {
 	}
 
 	// control.tar.gz
+	log.Debugf("Create deb control dir archive")
```
olegrok, line 67: `dir ` -> `directory`?

cli/pack/deb_control_dir.go
```diff
@@ -42,7 +42,7 @@ var (
 )
 
 func initControlDir(destDirPath string, projectCtx *project.ProjectCtx) error {
-	log.Debugf("Create DEB control dir: %s", destDirPath)
+	log.Debugf("Create DEB control dir")
```
olegrok, line 45: Why did you drop destDirPath here?

cli/pack/app_dir.go
dokshina: Because I think that long tmp directory name in logs is a bad idea. In fact, user doesn't care where application dir is placed in a tmp directory. Path to used tmp directory is logged at the beginning.

cli/create/git.go
dokshina: There is `Initialize application git repository` info level message above, I think it's enough. 

I've changed this message to `Add files to git index` anyway.

cli/pack/deb_control_dir.go
dokshina: Se comment above

PR #296, opened 2020-06-26
template: perform ddl in apply_config

printercu:
because `if opts.is_master` branch is never invoked in role.init during rolling update

### DIFF
--- a/examples/getting-started-app/app/roles/storage.lua
+++ b/examples/getting-started-app/app/roles/storage.lua
@@ -136,6 +136,12 @@ local exported_functions = {
 }
 
 local function init(opts)
+    for name, func in pairs(exported_functions) do
+        rawset(_G, name, func)
+    end
+end
+
+local function apply_config(config, opts)
     if opts.is_master then
         init_spaces()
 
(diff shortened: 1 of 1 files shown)


dokshina:
Please, fix luacheck errors.

printercu:
Fixed. Please, squash commits when merging.

examples/getting-started-app/app/roles/storage.lua
```diff
@@ -135,27 +135,27 @@ local exported_functions = {
...
             box.schema.func.create(name, {if_not_exists = true})
             box.schema.role.grant('public', 'execute', 'function', name, {if_not_exists = true})
         end
     end
-
-    for name, func in pairs(exported_functions) do
-        rawset(_G, name, func)
-    end
-
-    return true
```
dokshina, line 152: In the Cartridge application template both `init` and `apply_config` functions return `true`. 
This role is an example based on template, I think it should use the same convention.

The same is about function args, I suggest using `local function init(opts) -- luacheck: no unused args`, as it's defined in the template.

printercu: Return values of this functions are not used

dokshina: OK, I will remove return values from template, but renaming `opts` -> `options` confuses user. Example application is an edited template application, I don't see any reason to rename this parameter.

dokshina:
See #428

PR #308, opened 2020-07-06
Log command

### LINKED ISSUE #278: [2pt] We need `cartridge log` command
Since `cartridge start -d` redirects logs to a file, we can reveal instance(s) logs using `tail -f`. 

I suggest to implement `cartridge log` command:

`cartridge log [INSTANCE_NAME...] [flags]` show instance(s) logs. 
If no ` [INSTANCE_NAME...]` are specified, instances are discovered from cfg file (just like `start/stop/status` do).

Options:

* `-f, --follow` output appended data as the file grows;
* `-n, --lines=NUM` output the last NUM lines.

I think, it would be good to show logs with colorized prefixes (just like start in interactive mode does).

dokshina:
Closes #278

### DIFF
--- a/cli/commands/log.go
+++ b/cli/commands/log.go
@@ -24,6 +24,7 @@ func init() {
 	logCmd.Flags().StringVar(&ctx.Project.Name, "name", "", nameFlagDoc)
 
 	logCmd.Flags().StringVar(&ctx.Running.RunDir, "run-dir", "", runDirFlagDoc)
+	logCmd.Flags().StringVar(&ctx.Running.LogDir, "log-dir", "", logDirFlagDoc)
 	logCmd.Flags().StringVar(&ctx.Running.ConfPath, "cfg", "", cfgFlagDoc)
 
 	logCmd.Flags().BoolVar(&ctx.Running.WithStateboard, "stateboard", false, stateboardFlagDoc)
--- a/cli/commands/log.go
+++ b/cli/commands/log.go
@@ -1,0 +1,63 @@
+package commands
+
+import (
+	"fmt"
+
+	"github.com/apex/log"
+	"github.com/spf13/cobra"
+
+	"github.com/tarantool/cartridge-cli/cli/running"
+)
+
+var (
+	linesFlagDoc = fmt.Sprintf(`Output the last NUM lines.
+Defaults to %d`, defaultLinesLog)
+)
+
+const (
+	defaultLinesLog = 15
+)
+
+func init() {
+	rootCmd.AddCommand(logCmd)
+
+	logCmd.Flags().StringVar(&ctx.Project.Name, "name", "", nameFlagDoc)
+
+	logCmd.Flags().StringVar(&ctx.Running.RunDir, "run-dir", "", runDirFlagDoc)
+	logCmd.Flags().StringVar(&ctx.Running.ConfPath, "cfg", "", cfgFlagDoc)
+
+	logCmd.Flags().BoolVar(&ctx.Running.WithStateboard, "stateboard", false, stateboardFlagDoc)
+	logCmd.Flags().BoolVar(&ctx.Running.StateboardOnly, "stateboard-only", false, stateboardOnlyFlagDoc)
+
+	logCmd.Flags().BoolVarP(&ctx.Running.LogFollow, "follow", "f", false, followFlagDoc)
+	logCmd.Flags().IntVarP(&ctx.Running.LogLines, "lines", "n", defaultLinesLog, followFlagDoc)
+}
+
+var logCmd = &cobra.Command{
+	Use:   "log [INSTANCE_NAME...]",
+	Short: "Get logs of instance(s)",
+	Long:  fmt.Sprintf("Get logs of instance(s)n\n%s", runningCommonDoc),
+	Run: func(cmd *cobra.Command, args []string) {
+		err := runLogCmd(cmd, args)
+		if err != nil {
+			log.Fatalf(err.Error())
+		}
+	},
+}
+
+func runLogCmd(cmd *cobra.Command, args []string) error {
+	if err := running.FillCtx(&ctx, args); err != nil {
+		return err
+	}
+
+	if err := running.Log(&ctx); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+const (
+	followFlagDoc = `Output appended data as the file grows
+`
+)
--- a/cli/commands/log.go
+++ b/cli/commands/log.go
@@ -2,20 +2,22 @@ package commands
 
 import (
 	"fmt"
+	"strconv"
 
 	"github.com/apex/log"
 	"github.com/spf13/cobra"
 
+	"github.com/tarantool/cartridge-cli/cli/project"
 	"github.com/tarantool/cartridge-cli/cli/running"
 )
 
 var (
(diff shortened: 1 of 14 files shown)


cli/commands/log.go
```diff
@@ -0,0 +1,67 @@
...
+	logCmd.Flags().BoolVar(&projectCtx.StateboardOnly, "stateboard-only", false, stateboardOnlyFlagDoc)
+
+	logCmd.Flags().BoolVarP(&followLog, "follow", "f", false, followFlagDoc)
+	logCmd.Flags().IntVarP(&linesLog, "lines", "n", defaultLinesLog, followFlagDoc)
+}
+
+var logCmd = &cobra.Command{
+	Use:   "log [INSTANCE_NAME...]",
+	Short: "Get logs of instance(s)",
+	Long:  fmt.Sprintf("Get logs of instance(s)n\n%s", runningCommonDoc),
```
olegrok, line 43: `instance(s)n` -> `instance(s)`?

cli/common/utils.go
```diff
@@ -143,3 +145,91 @@ func ParseYmlFile(path string) (map[string]interface{}, error) {
...
+	buf := make([]byte, bufSize)
+
+	var filePos int64 = fileSize - bufSize
+	var lastNewLinePos int64 = 0
+	var newLinesN int = 0
+
+	// check last symbol of the last line
+
+	if err := readFromPos(f, fileSize-1, &buf); err != nil {
+		return 0, fmt.Errorf("%s", err)
```
olegrok, line 193: `return 0, err`

cli/common/utils.go
```diff
@@ -143,3 +145,91 @@ func ParseYmlFile(path string) (map[string]interface{}, error) {
...
+	}
+	if buf[0] != '\n' {
+		newLinesN++
+	}
+
+	lastPart := false
+
+Loop:
+	for {
+		if filePos < 0 {
```
olegrok, line 203: Is it possible?

cli/common/utils.go
```diff
@@ -143,3 +145,91 @@ func ParseYmlFile(path string) (map[string]interface{}, error) {
...
+			filePos = 0
+			lastPart = true
+
+			buf = make([]byte, fileSize%bufSize)
+		}
+
+		if err := readFromPos(f, filePos, &buf); err != nil {
+			return 0, fmt.Errorf("%s", err)
```
olegrok, line 211: `return 0, err`

cli/common/utils.go
```diff
@@ -143,3 +145,91 @@ func ParseYmlFile(path string) (map[string]interface{}, error) {
...
+		}
+
+		for i := len(buf) - 1; i >= 0; i-- {
```
olegrok, line 214: Seems it could be moved into separate function

cli/running/processes_set.go
```diff
@@ -231,3 +231,61 @@ func (set *ProcessesSet) Status() error {
...
+		time.Sleep(100 * time.Millisecond)
+	}
+
+	var errors []error
+
+	// wait for all processes result
+	for i := 0; i < len(*set); i++ {
+		select {
+		case res := <-resCh:
+			log.Infof(getResStr(&res))
```
olegrok, line 272: Is there simple `log.Info` without `f`?

test/project.py
```diff
@@ -255,3 +252,51 @@ def patch_init_to_send_statuses(project, statuses):
 
     with open(os.path.join(project.path, 'stateboard.init.lua'), 'w') as f:
         f.write(patched_init)
+
+
+def patch_init_to_log_lines(project, lines):
```
olegrok, line 257: Write simple description for it.

cli/running/processes_set.go
dokshina: Yes, in this case function just returns.

cli/common/utils.go
dokshina: This is the main part of this function, I'm not sure that is should be separated.

test/project.py
dokshina: Added description.

cli/commands/log.go
```diff
@@ -0,0 +1,77 @@
...
+
+	if err := running.Log(&projectCtx, followLog, linesLog); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+const (
+	followFlagDoc = `Output appended data as the log grows
```
olegrok, line 75: `follow` description from `man journalctl`.
```
Show only the most recent journal entries, and continuously print new entries as they are appended to the journal.
```

cli/common/utils.go
```diff
@@ -16,6 +16,8 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+var bufSize int64 = 10000
```
olegrok, line 19: Why is it not const?

cli/common/utils.go
```diff
@@ -143,3 +145,95 @@ func ParseYmlFile(path string) (map[string]interface{}, error) {
...
+		return 0, nil
+	}
+
+	buf := make([]byte, bufSize)
+
+	var filePos int64 = fileSize - bufSize
+	var lastNewLinePos int64 = 0
+	var newLinesN int = 0
+
+	// check last symbol of the last line
```
olegrok, line 194: Why is this case separated from others?

cli/common/utils.go
```diff
@@ -143,3 +145,95 @@ func ParseYmlFile(path string) (map[string]interface{}, error) {
...
+
+	lastPart := false
+
+Loop:
+	for {
+		if filePos < 0 {
+			filePos = 0
+			lastPart = true
+
+			buf = make([]byte, fileSize%bufSize)
```
olegrok, line 211: I suggest you could avoid recreation of buffer for each iteration. Create it once and reuse.

cli/common/utils.go
```diff
@@ -143,3 +145,95 @@ func ParseYmlFile(path string) (map[string]interface{}, error) {
...
+
+	if _, err := f.Read(*buf); err != nil {
+		return fmt.Errorf("Failed to read: %s", err)
+	}
+
+	return nil
+}
+
+// GetLastNLinesBegin return the position of last n lines begin
+func GetLastNLinesBegin(filepath string, n int) (int64, error) {
```
olegrok, line 162: BTW, have you measure performance of this implementation on big files? Say, 1GB and more.

cli/common/utils.go
dokshina: To be changed in test.

cli/common/utils.go
dokshina: To process `one\ntwo` and `one\ntwo\n` cases the same way.

cli/common/utils.go
dokshina: Buffer is recreated once at the last part. It's size can be less than `bufSize` and `len(buf)` returns `bufSize`, no matter how many bytes was read.

cli/commands/log.go
dokshina: I was thinking about `tail` command writing this feature. Why do you think that current description is bad?

olegrok: > Why do you think that current description is bad?

I don't. But I suggest you an alternative.

cli/common/utils.go
dokshina: I've removed the buffer recreation.

PR #311, opened 2020-07-08
`--timeout` option for `start` command

### LINKED ISSUE #276: [2pt] Start timeout
I suggest adding `--timeout` option  to specify the time to wait for instance is started.

Now on `cartridge start -d` CLI waits for instance to send `READY=1` to notify socket.
And it can wait forever...

I sure that we need to stop it after timeout (by default, it's a `1m`).
But user can say `cartridge start -d --timeout 10m` that mean `I want to wait 10 minutes until my instance is started`.

dokshina:
Closes #276

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -361,7 +361,7 @@ The following options (``[flags]``) are supported:
   Defaults to ``./instances.yml`` (or to the value of the "cfg"
   parameter in the Cartridge `configuration file <Overriding default options_>`_).
 
-* ``--daemonize / -d`` starts the instance in background.
+* ``--daemonize, -d`` starts the instance in background.
   With this option, Tarantool also waits until the application's main script is
   finished.
   For example, it is useful if the ``init.lua`` requires time-consuming startup
@@ -377,6 +377,11 @@ The following options (``[flags]``) are supported:
 
 * ``--name string`` defines the application name.
   By default, it is taken from the application rockspec.
+
+* ``--timeout duration`` Time to wait for instance(s) start in background.
+  Should be specified in the form ``72h3m0.5s``.
+  The default timeout is ``1m0s``.
+  Timeout ``0s`` means no timeout (wait for instance start forever).
 
 ^^^^^^^^^^^^^^^^^^^^^^
 Environment variables
--- a/README.rst
+++ b/README.rst
@@ -378,10 +378,11 @@ The following options (``[flags]``) are supported:
 * ``--name string`` defines the application name.
   By default, it is taken from the application rockspec.
 
-* ``--timeout duration`` Time to wait for instance(s) start in background.
-  Should be specified in the form ``72h3m0.5s``.
-  The default timeout is ``1m0s``.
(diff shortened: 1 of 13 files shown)


cli/commands/start.go
```diff
@@ -2,13 +2,28 @@ package commands
...
 	"github.com/tarantool/cartridge-cli/cli/running"
 )
 
+const (
+	defaultStartTimeout = 1 * time.Minute
+)
+
+var (
+	daemonize      bool
+	timeout        time.Duration
```
olegrok, line 20: `type Duration int64` that means that `timeout` is signed.

Could we pass some negative value?

dokshina: This option accepts value in the `00h4m0.5s` form, the negative value will be rejected by cobra (or, as I understood, by the `time.Duration` constructor).

dokshina: I've added error on negative timeout.

README.rst
```diff
@@ -378,6 +378,12 @@ The following options (``[flags]``) are supported:
 * ``--name string`` defines the application name.
   By default, it is taken from the application rockspec.
 
+* ``--timeout string`` Time to wait for instance(s) start in background.
+  Can be specified in seconds or in the duration form (``72h3m0.5s``).
```
olegrok, line 382: `72h3m0.5s`

`0.5s`

Is it really possible to specify a half of second?

cli/commands/common.go
```diff
@@ -0,0 +1,39 @@
...
+		durationStr = fmt.Sprintf("%ds", seconds)
+	}
+
+	duration, err := time.ParseDuration(durationStr)
+	if err != nil {
+		return 0, err
+	}
+
+	if duration < 0 {
+		return 0, fmt.Errorf("Negative duration is specified")
```
olegrok, line 35: I suggest "negative timeout"

olegrok:
Is it possible/reasonable to implement simple wrapper (as for spinner) that will run some command with timeout. It allows to generalize `timeout` logic not only for start. But for stop and another commands that could be implemented in future.

README.rst
dokshina: Yes, it's parsed to `500 * time.Milisecond`. (but I will check id additionally)

cli/commands/common.go
dokshina: But this function is called `getDuration`.

dokshina:
Yes, but there are some problems with running spinner in different goroutines. I'm planning to implement it in a big separated patch.

cli/commands/common.go
olegrok: But user specifies it as "timeout" and don't know about "durations"

olegrok: But ok. May be you are right

olegrok:
> Yes, but there are some problems with running spinner in different goroutines. I'm planning to implement it in a big separated patch.

Please file an issue.

README.rst
dokshina: Yes, it's parsed to `500ms`. I've checked.

Issue #282, opened 2020-06-22
[8pt] Start reworks


dokshina:
I don't like some `start,stop,status` (I will call that's all `running` commands)  features in the old CLI, so I suggest to change it in a new version (since it's a next major version that can be incompatible).

## Intro

In the old cli running commands had these options (with default values):

* `--run-dir` - directory where pid and sock files are stored (`/var/run/tarantool`);
* `--apps-path` - path to apps directory when running in multi-app env (`/usr/share/tarantool`).
   App directory is `<apps-path>/<app-name>`;
* `--cfg` -  cartridge instances config file (`instances.yml`);
* `--stateboard`, `--stateboard-only` flags indicates if stateboard instance should be managed with or instead of app instances ;
* `--script` - application entrypoint (`init.lua`).

All these options has default values that can be overwritten  by `TARANTOOL_*` env variables or values from `./.cartridge.yml` or `~/.cartridge.yml`.

New version of CLI doesn't support `apps-path` option.

### Problem

Application can be started from it's directory (local) or from arbitrary place (system).
In the system case we discover application as `<apps-path>/<app-name>`.

Now, all default options except `--cfg` are defined for system start. 
If current directory contains rockspec than current application is running, else
cli is looking for it in `/usr/share/tarantool/<app-name>`. 
In this case, if app dir doesn't contain rockspec, you get error that your app isn't
found in `/usr/share/tarantool/<app-name>` and it's a bit confused.

The second case is a `cfg` file - now it's `./instances.yml` in the current directory.
It's OK for local running, but if you wand to manage apps from any place, it leads to frustration...
For system running we need to use `/etc/tarantool/conf.d` as a default value. 

The same about `.cartridge.yml` file.

### Solution

I sure that user should specify if he wants `system` running or `local`.

I suggest to introduce `--system` flag, that is `false` by default.
Local running is used *only for running from app directory*.

System running requires app name specifying. 
(No `start router`, only `start --name myapp router` or `start myapp router s1-master`)

Since CLI can start two or more instances (`cartridge start router s1-master ...`), 
we should deny to specify different apps instances in one command.

Depends on mode, options has different default values:

**run-dir**:
  * local: `./tmp/run`
  * system: `/var/run/tarantool`

**cfg**:
  * local: `./instances.yml`
  * system: `/etc/tarantool/conf.d`

**apps-path**:
  * local: Is not allowed - local running it's for starting one app in it's directory
  * system: `/usr/share/tarantool`

[...]

PR #314, opened 2020-07-14
Refactor command flags

dokshina:
This is the preparation step for #282 

* Make *Cmd variables non-global to avoid copy-paste errors
* Move all flags usage to the separated file
* Add function for flags configuring
* Add functions for adding common flags
* Move default values to the separated file
* Support -v and --version

### DIFF
--- a/cli/commands/const.go
+++ b/cli/commands/const.go
@@ -1,0 +1,14 @@
+package commands
+
+import "time"
+
+// DEFAULT VALUES
+const (
+	defaultStartTimeout = 1 * time.Minute
+	defaultLogLines     = 15
+)
+
+// ENV
+const (
+	cartridgeTmpDirEnv = "CARTRIDGE_TEMPDIR"
+)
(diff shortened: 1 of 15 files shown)


cli/commands/validate.go
```diff
@@ -0,0 +1,58 @@
+package commands
```
olegrok, line 1: I believe that `validatePack` should live in `pack/validate.go` but not in common place.

dokshina: Yes, I agree.

Fixed.

cli/commands/const.go
```diff
@@ -0,0 +1,14 @@
+package commands
+
+import "time"
+
+// DEFAULT VALUES
+const (
+	defaultStartTimeout = 1 * time.Minute
```
olegrok, line 7: It's not "common" values. I think it could be easily separated to start/const.go and log/const.go.

But you can argue with me.

dokshina: I sure that this default values are about parsing commands and should be declared in the `commands` package.

PR #315, opened 2020-07-14
Global running

### LINKED ISSUE #282: [8pt] Start reworks

dokshina:
Closes #282


cli/commands/usage.go
```diff
@@ -127,7 +140,7 @@ Ignored if "--stateboard-only" is specified
 `
 
 	stateboardOnlyUsage = `Manage only application stateboard
-If specified, "INSTANCE_NAME..." are ignored.
```
olegrok, line 130: Why did you do it?

cli/project/files.go
```diff
@@ -5,36 +5,51 @@ import (
...
 	"path/filepath"
 
+	"github.com/apex/log"
 	"github.com/tarantool/cartridge-cli/cli/common"
 	"github.com/tarantool/cartridge-cli/cli/context"
 )
 
 const (
-	cartridgeLocalConf = ".cartridge.yml"
+	defaultHomeDir = "/home"
```
olegrok, line 14: Why not `$HOME`?

olegrok: See here - https://github.com/tarantool/cartridge-cli/pull/315/files#diff-660f4520ad6b4abeaabb003310e50b43R177

But you don't check access rights for this directory

cli/project/files.go
```diff
@@ -152,47 +167,114 @@ func getPath(conf map[string]interface{}, opts PathOpts) (string, error) {
...
+// * ./.cartridge.yml for local running
+// * ~/.cartridge.yml for global running
 //
 // The priority of sources is:
 // * user-specified flags
-// * value from .cartridge.yml
+// * value from .cartridge.yml (if useConf is true)
 // * default values (defined here in const section)
-func SetLocalRunningPaths(ctx *context.Ctx) error {
+func SetRunningPaths(ctx *context.Ctx, useConf bool) error {
```
olegrok, line 223: Seems it could be splitted to `SetLocalRunningPaths ` and `SetGlobalRunningPaths `. If here is quite big

olegrok: Common part could be moved into separate function

dokshina: `useConf ` variable isn't about local/global, it's about running/pack (in pack rpm and deb the same global paths are used).

The main difference of `SetLocalRunningPaths` and `SetGlobalRunningPaths` is the default values and `App(s)Dir` variable. And I sure that in the one functions it's quite simple to understand the difference and support this code.

cli/commands/usage.go
dokshina: Because I decided to remove trailing dots.

olegrok: Is it only one place?

Could you please move such refactoring to separate patch if there are other places?

dokshina: I've done it in the previous patch, but missed this place.

cli/project/files.go
dokshina: Fixed:

* `os.UserHomeDir()` is used now (BTW, it returns `${HOME}`);
* if getting home directory fails, cli returns an error.

cli/commands/usage.go
```diff
@@ -84,14 +84,27 @@ result artifact
...
+	runningCommonUsage = `Manage instance(s) of current application
 
-Application name is described from rockspec in the current directory.
+There are two modes of running instances: local and global.
+If --global flag is specified, then application from <apps-dir>/<name>
+is started and global paths (run-dir, data-dir, log-dir, cfg)
+are used, otherwise application from the current directory is started
+and all application directories and files are placed in the current directory.
+
+I case of local running, application name is taken from rockspec in
```
olegrok, line 95: typo: `In case`

olegrok: `rockspec in` -> `rockspec from current directory`?

cli/commands/usage.go
```diff
@@ -84,14 +84,27 @@ result artifact
...
+the current directory. For global running name should be passed using
+--name option.
+
+Some flags default options can be overridden in .cartridge.yml config file.
```
olegrok, line 99: `config file` -> `configuration file`?

dokshina:
We've decided that this feature is unused.

PR #324, opened 2020-07-30
--force option for stop command

### LINKED ISSUE #312: [2pt] `--force` option for `stop` command
Now `cartridge stop` gracefully sends `SIGTERM` signal. But sometimes we want to kill instance(s), and it would be great to say `cartridge stop --force` instead of sending `SIGKILL` manually.

Thx @olegrok for this great idea

dokshina:
Sometimes we need to kill instances instead of gracefully sending SIGTERM.

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Closes #312

### DIFF
--- a/cli/commands/usage.go
+++ b/cli/commands/usage.go
@@ -86,7 +86,7 @@ result artifact
 const (
 	runningCommonUsage = `Running instance(s) of current application
 
-Application name is described from rockspec in the current directory.
+Application name is taken from rockspec in the current directory.
 
 If INSTANCE_NAMEs aren't specified, then all instances described in
 config file (see --cfg) are used.
@@ -131,6 +131,9 @@ If specified, "INSTANCE_NAME..." are ignored.
 `
 
 	logFollowUsage = `Output appended data as the log grows
+`
+
+	stopForceUsage = `Force instance(s) stop (sends SIGKILL)
 `
 )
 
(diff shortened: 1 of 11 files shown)


cli/commands/usage.go
```diff
@@ -86,7 +86,7 @@ result artifact
 const (
 	runningCommonUsage = `Running instance(s) of current application
 
-Application name is described from rockspec in the current directory.
+Application name is taken from rockspec in the current directory.
```
olegrok, line 89: May be `retrieved`?

cli/running/running.go
```diff
@@ -112,7 +112,7 @@ func Stop(ctx *context.Ctx) error {
 		return fmt.Errorf("No instances specified")
 	}
 
-	if err := processes.Stop(); err != nil {
+	if err := processes.Stop(ctx.Running.StopForced); err != nil {
```
olegrok, line 115: I think it could be whole `ctx.Running` not only single flag. In future we may introduce more flags and function will have too many arguments. Keep it in mind don't change it now.

dokshina: Yes, maybe sometimes we should keep all parameters in the options map or smth like this.

But `force` isn't an attribute of processes set, it's a `Stop` option. Just like `daemonize` for `Start` or `follow, n` options for `Log`.

cli/commands/usage.go
dokshina: In the README `taken` word is used.

PR #325, opened 2020-07-31
Clean command

### LINKED ISSUE #316: [2pt] Command to wipe a data dir of the specified instance
I need to delete data of one instance, not of the entire cluster, sometimes. I can do it with bash, but it's not so convenient as the command execution.

dokshina:
Sometimes we want to cleanup all instances files (working directories, logs, sockets).
We introduce `cartridge clean` command to do it easily.

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Closes #316

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -126,6 +126,7 @@ The following commands are supported:
 * ``stop`` — stop a Tarantool instance(s);
 * ``status`` — get current instance(s) status;
 * ``log`` — get logs of instance(s);
+* ``clean`` - clean instance(s) files;
 * ``pack`` — pack the application into a distributable bundle.
 
 The following global flags are supported:
@@ -478,9 +479,9 @@ are supported:
 
 .. // Please, update the doc in cli/commands on updating this section
 
-***********
+*******
 ``log``
-***********
+*******
 
 To get logs of the instance running in background, use the ``log`` command:
 
@@ -499,6 +500,40 @@ The following `options <Options_>`_ from the ``start`` command
 are supported:
 
 * ``--log-dir DIR``
+* ``--run-dir DIR``
+* ``--cfg FILE``
+* ``--stateboard``
+* ``--stateboard-only``
+
+.. // Please, update the doc in cli/commands on updating this section
+
+.. _cartridge-cli-packing-an-application:
+
+*********
+``clean``
+*********
+
+To remove instance(s) files (log, workdir, console and notify sockets),
+use the ``clean`` command:
+
+.. code-block:: bash
+
+    cartridge clean [INSTANCE_NAME...] [flags]
+
+By default, `cartridge clean` for running instances causes an error.
+But it can be forced with ``--force`` flag.
+PID-files aren't remove because they are used by `Cartridge CLI` itself
+to check instance process status.
+
+The following options (``[flags]``) are supported:
+
+* ``-f, --force`` remove files even if instances are running.
+
+The following `options <Options_>`_ from the ``start`` command
+are supported:
+
+* ``--log-dir DIR``
+* ``--data-dir DIR``
 * ``--run-dir DIR``
 * ``--cfg FILE``
 * ``--stateboard``
--- a/README.rst
+++ b/README.rst
@@ -521,7 +521,7 @@ use the ``clean`` command:
     cartridge clean [INSTANCE_NAME...] [flags]
 
 `cartridge clean` for running instance(s) causes an error.
-PID-files aren't remove because they are used by `Cartridge CLI` itself
+PID-files aren't removed because they are used by `Cartridge CLI` itself
 to check instance process status.
 
 The following `options <Options_>`_ from the ``start`` command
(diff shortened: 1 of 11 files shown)


README.rst
```diff
@@ -508,6 +509,40 @@ are supported:
...
+
+To remove instance(s) files (log, workdir, console and notify sockets),
+use the ``clean`` command:
+
+.. code-block:: bash
+
+    cartridge clean [INSTANCE_NAME...] [flags]
+
+By default, `cartridge clean` for running instances causes an error.
+But it can be forced with ``--force`` flag.
```
olegrok, line 524: I'm against it. We doesn't have to drop any data for running instances

dokshina: Yes, you are right. Removed.

cli/running/process.go
```diff
@@ -420,3 +421,58 @@ func (process *Process) Log(follow bool, n int) error {
...
+		process.consoleSock,
+		process.notifySockPath,
+		// PID file isn't deleted
+		// since it's used by Cartridge CLI to start and stop instance
+	}
+
+	var existedFiles []string
+	var skipped = true
+
+	for _, path := range pathsToDelete {
```
olegrok, line 442: I think it should be splitted into two parts: validation and execution. At first you check possibility to delete files and only then remove it physically. It should prevent partial dropping if you don't have enough right to drop files for e.g. the third instance and delete two first but don't try to clean for fourth and fifth.

test/utils.py
```diff
@@ -278,6 +278,40 @@ def get_logs(self, project, instances=[], n=None, log_dir=None, run_dir=None, cf
...
         return logs
 
+    def clean(self, project, instances=[], log_dir=None, run_dir=None, cfg=None, data_dir=None,
+              stateboard=False, stateboard_only=False, force=False, exp_rc=0):
+        cmd = [self._cartridge_cmd, 'clean']
+        if stateboard:
+            cmd.append('--stateboard')
+        if stateboard_only:
+            cmd.append('--stateboard-only')
+        if force:
```
olegrok, line 288: Should be removed

cli/commands/usage.go
```diff
@@ -135,6 +134,9 @@ If specified, "INSTANCE_NAME..." are ignored.
 
 	stopForceUsage = `Force instance(s) stop (sends SIGKILL)
 `
+
+	cleanForceUsage = `Remove files even if instances are running
```
olegrok, line 138: should be removed

cli/running/running.go
```diff
@@ -170,3 +170,29 @@ func Log(ctx *context.Ctx) error {
...
 	return nil
 }
+
+func Clean(ctx *context.Ctx) error {
+	var err error
+
+	if !ctx.Running.StateboardOnly && len(ctx.Running.Instances) == 0 {
+		ctx.Running.Instances, err = collectInstancesFromConf(ctx)
+		if err != nil {
+			return fmt.Errorf("Failed to get configured instances from conf: %s", err)
```
olegrok, line 180: `conf` -> `config`

cli/running/process.go
dokshina: Now files deletion process isn't stopped, if fails for one file. It just collects errors and returns an error if failed for some files.

cli/running/process.go
```diff
@@ -420,3 +421,63 @@ func (process *Process) Log(follow bool, n int) error {
...
+		ProcessID: process.ID,
+	}
+
+	pathsToDelete := []string{
+		process.logFile,
+		process.workDir,
+		process.consoleSock,
+		process.notifySockPath,
+		// PID file isn't deleted
+		// since it's used by Cartridge CLI to start and stop instance
```
olegrok, line 436: It could be mentioned in readme as well

cli/running/process.go
```diff
@@ -420,3 +421,63 @@ func (process *Process) Log(follow bool, n int) error {
...
+		if _, err := os.Stat(path); os.IsNotExist(err) {
+			nonExistedFiles = append(nonExistedFiles, path)
+			continue
+		} else if err != nil {
+			errors = append(errors, err.Error())
+			continue
+		} else if err := os.RemoveAll(path); err != nil {
+			errors = append(errors, err.Error())
+			continue
+		}
```
olegrok, line 453:
```
else {
    skipped = false
}
```

And drop three `continue`

README.rst
```diff
@@ -508,6 +509,35 @@ are supported:
...
+
+To remove instance(s) files (log, workdir, console and notify sockets),
+use the ``clean`` command:
+
+.. code-block:: bash
+
+    cartridge clean [INSTANCE_NAME...] [flags]
+
+`cartridge clean` for running instance(s) causes an error.
+PID-files aren't remove because they are used by `Cartridge CLI` itself
```
olegrok, line 524: `remove ` -> `removed`

cli/running/process.go
dokshina: This is for developer to remember why PID files aren't deleted.

olegrok: I think it could be useful to know for user

PR #331, opened 2020-08-12
Refactor results

dokshina:


cli/common/result.go
```diff
@@ -0,0 +1,62 @@
+package common
```
olegrok, line 1: It would be great to rename `Res` to `Result`.

dokshina: Everything for you

PR #336, opened 2020-08-13
Add bash autocomplete

### LINKED ISSUE #335: [2pt] Autocomletion
I think it's time to add completion scripts. Moreover, `cobra` generates its from the box.

dokshina:
I didn't forget about

- [x] Changelog
- [x] Documentation

Closes #335


cli/commands/completion.go
```diff
@@ -0,0 +1,85 @@
...
+	genCmd.Flags().StringVar(&bashCompFilePath, "bash", defaultBashCompFilePath, "Bash completion file path")
+	genCmd.Flags().StringVar(&zshCompFilePath, "zsh", defaultZshCompFilePath, "Zsh completion file path")
+}
+
+// cutFlagsDesc cuts command usage on first '\n'
+// it's needed to make zsh comletion for flags prettier
+func cutFlagsDesc(cmd *cobra.Command) {
+	flags := cmd.Flags()
+	flags.VisitAll(func(f *pflag.Flag) {
+		f.Usage = strings.SplitN(f.Usage, "\n", 2)[0]
```
olegrok, line 55: Why "2"? Not "1"?

cli/commands/completion.go
```diff
@@ -0,0 +1,85 @@
...
+
+	rootCmd.AddCommand(genCmd)
+
+	genCmd.Flags().StringVar(&bashCompFilePath, "bash", defaultBashCompFilePath, "Bash completion file path")
+	genCmd.Flags().StringVar(&zshCompFilePath, "zsh", defaultZshCompFilePath, "Zsh completion file path")
+}
+
+// cutFlagsDesc cuts command usage on first '\n'
+// it's needed to make zsh comletion for flags prettier
+func cutFlagsDesc(cmd *cobra.Command) {
```
olegrok, line 52: Desc -> Description

cli/commands/completion.go
```diff
@@ -0,0 +1,85 @@
...
+		return fmt.Errorf("Failed to create zsh completion directory: %s", err)
+	}
+
+	// gen completions
+	if err := cmd.Root().GenBashCompletionFile(bashCompFilePath); err != nil {
+		return fmt.Errorf("failed to generate bash completion: %s", err)
+	}
+
+	if err := cmd.Root().GenZshCompletionFile(zshCompFilePath); err != nil {
+		return fmt.Errorf("failed to generate zsh completion: %s", err)
```
olegrok, line 81: Should it really be an error or we could allow users to work our users e.g. on read-only systems?

cli/running/process.go
```diff
@@ -176,7 +176,6 @@ func (process *Process) Start(daemonize bool) error {
 	process.cmd = exec.CommandContext(ctx, "tarantool", process.entrypoint)
 
 	process.cmd.Env = append(os.Environ(), process.env...)
-	process.cmd.Dir = process.workDir
```
olegrok, line 179: Why was it removed?

magefile.common.go
```diff
@@ -32,3 +35,53 @@ func downloadFile(url string, dest string) error {
 
 	return nil
 }
+
+func copyFile(destPath, srcPath string) error {
```
olegrok, line 39: I hope you don't really need because you have https://github.com/otiai10/copy/blob/master/copy.go

cli/commands/completion.go
dokshina: Because 2 is a number of parts, not splits.

cli/running/process.go
dokshina: It was a part of #337, I removed it.

magefile.common.go
dokshina: Yes, you are right.

cli/commands/completion.go
dokshina: I think that command that generates completion should fail if it fails to generate completion.

(It's not the `__complete` command that is called explicitly, this command generates scripts)

CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Now instance process is ran from the application directory
 
+### Added
+
+- `cartridge gen` command to generate shell completion scripts
```
olegrok, line 16: gen completion

Issue #327, opened 2020-08-04
[8pt] Repair: Base


dokshina:
Sometimes we have to repair cluster using brute force - by manually editing `.config.yml`.
It could be useful to automate this flow.

We need to support these actions:

* change instance advertise URI;
* remove node from cluster;
* change replicaset master.

dokshina:
I suggest adding new command `repair` that includes sub-commands:

### Repair 

```
cartridge repair [command]

Available Commands:
  remove  Remove instance from cluster config
  uri  Change instance URI
  set-master  Change replicaset master # or leader???
```

Each subcommand has `--name`, `--data-dir` and `--dry-run` options:

```
--name string  Application name
--data-dir string  Directory where instances' data is stored
                             "<data-dir>/<app-name>.<instance-name>"
                             Defaults to /var/lib/tarantool
--dry-run  Run command in dry-run mode
                  Show changes but don't apply them
```

The main idea of all `repair` commands is to patch all application instances config:

* topology config is stored in the `<work-dir>/<config>/<topology.yml>`
* each instance working directory is `<data-dir>/<app-name>.<instance-name>`

So, `cartridge repair` walks across `<data-dir>/<app-name>.*` directories and patches `<data-dir>/<app-name>.*/<config>/<topology.yml>` file.

### Repair URI

```
cartridge repair uri FROM TO [flags]
```

Changes advertise URI of one instance. Fails if there is no instances with specified URI:

```diff
```

PR #339, opened 2020-08-18
Repair: Set URI

dokshina:
- [x] Tests
- [x] Documentation (see #343)
- [x] Changelog

First part of #327

### DIFF
--- a/cli/common/files.go
+++ b/cli/common/files.go
@@ -84,7 +84,7 @@ func FileLinesScanner(file *os.File) *bufio.Scanner {
 	return scanner
 }
 
-// GetFileContent returns file content bytes
+// GetFileContent returns file content as a bytes slice
 func GetFileContentBytes(path string) ([]byte, error) {
 	file, err := os.Open(path)
 	if err != nil {
@@ -101,7 +101,7 @@ func GetFileContentBytes(path string) ([]byte, error) {
 
 }
 
-// GetFileContent returns file content
+// GetFileContent returns file content as a string
 func GetFileContent(path string) (string, error) {
 	fileContentBytes, err := GetFileContentBytes(path)
 	if err != nil {
--- a/cli/common/files.go
+++ b/cli/common/files.go
@@ -84,20 +84,31 @@ func FileLinesScanner(file *os.File) *bufio.Scanner {
 	return scanner
 }
 
-// GetFileContent returns file content
-func GetFileContent(path string) (string, error) {
+// GetFileContent returns file content bytes
+func GetFileContentBytes(path string) ([]byte, error) {
 	file, err := os.Open(path)
 	if err != nil {
-		return "", err
+		return nil, err
 	}
 	defer file.Close()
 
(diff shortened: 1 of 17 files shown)


cli/common/files.go
```diff
@@ -84,20 +84,31 @@ func FileLinesScanner(file *os.File) *bufio.Scanner {
 	return scanner
 }
 
-// GetFileContent returns file content
-func GetFileContent(path string) (string, error) {
+// GetFileContent returns file content bytes
```
olegrok, line 87: _as bytes slice_

cli/common/files.go
```diff
@@ -84,20 +84,31 @@ func FileLinesScanner(file *os.File) *bufio.Scanner {
...
 	fileContent, err := ioutil.ReadAll(file)
+	if err != nil {
+		return nil, err
+	}
+
+	return fileContent, nil
+
+}
+
+// GetFileContent returns file content
```
olegrok, line 104: _as a string_

cli/repair/common.go
```diff
@@ -0,0 +1,177 @@
...
+		return nil, fmt.Errorf("%s is not a directory", ctx.Running.DataDir)
+	}
+
+	if ctx.Cli.Verbose {
+		log.Infof("Data directory is set to: %s", ctx.Running.DataDir)
+	}
+
+	workDirs, err := ioutil.ReadDir(ctx.Running.DataDir)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to list data directory: %s", err)
```
olegrok, line 36: `list data dir` -> `list a data dir`?

cli/repair/common.go
```diff
@@ -0,0 +1,177 @@
...
+			changelog = append(changelog, "Topology config changes:")
+			changelog = append(changelog, configDiff...)
+		} else {
+			changelog = append(changelog, "Topology config wasn't changed")
+		}
+
+		changelog = append(changelog, "") // an empty line to separate instances changes
+	}
+
+	if !ctx.Repair.DryRun {
```
olegrok, line 134: Maybe running without "dry-run" option is more common case than without it.

Consider

```go
if ctx.Repair.DryRun {
    return changelog, nil
}
-- actions to run without dry-run
```

But it's mostly up to you

cli/repair/repair.go
```diff
@@ -0,0 +1,78 @@
...
+
+			resCh <- res
+		}(workDirPath, workDirName, resCh)
+	}
+
+	var errors []error
+
+	for i := 0; i < len(appWorkDirNames); i++ {
+		select {
+		case res := <-resCh:
```
olegrok, line 56: Please add some timeout

cli/repair/common.go
dokshina: Yes, I agree.

cli/repair/common.go
dokshina: Maybe `the`?

dokshina: Changed to `the`.

PR #353, opened 2020-08-24
Repair: Fix fail on non bootstrapped instance

dokshina:
When instance is started, but not bootstrapped yet, data directory exists, but cluster-wide configuration don't.
It's a valid case, so it shouldn't cause any error (only if there is no one bootstrapped instance).

I didn't forget about

- [x] Tests
- [ ] Changelog
- [ ] Documentation

### DIFF
--- a/cli/repair/app_configs.go
+++ b/cli/repair/app_configs.go
@@ -31,6 +31,11 @@ func getAppConfigs(instanceNames []string, ctx *context.Ctx) (AppConfigs, error)
 		if err != nil {
 			return appConfigs, fmt.Errorf("Failed to get cluster-wide config path: %s", err)
 		}
+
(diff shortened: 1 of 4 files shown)


cli/repair/app_configs.go
```diff
@@ -31,6 +31,11 @@ func getAppConfigs(instanceNames []string, ctx *context.Ctx) (AppConfigs, error)
 		if err != nil {
 			return appConfigs, fmt.Errorf("Failed to get cluster-wide config path: %s", err)
 		}
+
+		if topologyConfPath == "" {
```
olegrok, line 35: Add comment please what does it mean

olegrok (approved):

Issue #249, opened 2020-05-20
Implement "watch" mode for "cartridge start"


knazarov:
When I develop code, I would like "cartridge start" to automatically watch for changes on filesystem and reload the code.

The primitive approach may be to check the code for errors by doing a "load", and then doing a complete restart in the same session if the code loads OK.

More advanced approach may be to do hot code reload by reloading modules in `package.loaded`.

artur-barsegyan:
Depends on https://github.com/tarantool/cartridge/issues/1100

PR #358, opened 2020-08-26
Repair: Reload Config

dokshina:
- [x] Tests
- [ ] Changelog
- [ ] Documentation

### DIFF
--- a/cli/commands/common.go
+++ b/cli/commands/common.go
@@ -59,12 +59,12 @@ func addCommonRunningPathsFlags(cmd *cobra.Command) {
 
 func addCommonRepairFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&ctx.Project.Name, "name", "", "Application name")
-	cmd.Flags().StringVar(&ctx.Running.DataDir, "data-dir", "", repairDataDirUsage)
 	cmd.Flags().BoolVar(&ctx.Repair.DryRun, "dry-run", false, dryRunUsage)
 	cmd.Flags().BoolVarP(&ctx.Repair.Force, "force", "f", false, repairForceUsage)
+	cmd.Flags().StringVar(&ctx.Running.DataDir, "data-dir", "", repairDataDirUsage)
 }
 
 func addCommonRepairPatchFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&ctx.Running.RunDir, "run-dir", "", repairRunDirUsage)
-	cmd.Flags().BoolVar(&ctx.Repair.NoReload, "no-reload", false, repairNoReloadUsage)
+	cmd.Flags().BoolVar(&ctx.Repair.Reload, "reload", false, repairReloadUsage)
 }
--- a/cli/commands/common.go
+++ b/cli/commands/common.go
@@ -63,3 +63,8 @@ func addCommonRepairFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&ctx.Repair.DryRun, "dry-run", false, dryRunUsage)
 	cmd.Flags().BoolVarP(&ctx.Repair.Force, "force", "f", false, repairForceUsage)
 }
+
+func addCommonRepairPatchFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&ctx.Running.RunDir, "run-dir", "", repairRunDirUsage)
+	cmd.Flags().BoolVar(&ctx.Repair.NoReload, "no-reload", false, repairNoReloadUsage)
+}
--- a/cli/common/socket.go
+++ b/cli/common/socket.go
@@ -56,11 +56,8 @@ type TarantoolEvalRes struct {
 func EvalTarantoolConn(conn net.Conn, funcBody string) (interface{}, error) {
 	evalFuncTmpl := `
 	local ok, res, err = pcall(function()
-		local function f()
-			require('fiber').self().storage.console = nil
-			{{ .FunctionBody }}
-		end
-		return f()
+		require('fiber').self().storage.console = nil
+		{{ .FunctionBody }}
 	end)
 
 	if res == nil then res = box.NULL end
--- a/cli/common/socket.go
+++ b/cli/common/socket.go
@@ -31,6 +31,19 @@ func ReadFromConn(conn net.Conn) ([]byte, error) {
 	return data, nil
 }
 
+func WriteToConn(conn net.Conn, data string) error {
+	writer := bufio.NewWriter(conn)
+	if _, err := writer.WriteString(data); err != nil {
+		return fmt.Errorf("Failed to send to socket: %s", err)
+	}
+
+	if err := writer.Flush(); err != nil {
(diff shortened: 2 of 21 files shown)


cli/commands/common.go
```diff
@@ -63,3 +63,8 @@ func addCommonRepairFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&ctx.Repair.DryRun, "dry-run", false, dryRunUsage)
 	cmd.Flags().BoolVarP(&ctx.Repair.Force, "force", "f", false, repairForceUsage)
 }
+
+func addCommonRepairPatchFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&ctx.Running.RunDir, "run-dir", "", repairRunDirUsage)
+	cmd.Flags().BoolVar(&ctx.Repair.NoReload, "no-reload", false, repairNoReloadUsage)
```
olegrok, line 69: I'm not sure that reload by default it's ok. I'd prefer to disable it by default.

cli/common/socket.go
```diff
@@ -0,0 +1,140 @@
...
+// to be correctly processed
+func EvalTarantoolConn(conn net.Conn, funcBody string) (interface{}, error) {
+	evalFuncTmpl := `
+	local ok, res, err = pcall(function()
+		local function f()
+			require('fiber').self().storage.console = nil
+			{{ .FunctionBody }}
+		end
+		return f()
+	end)
```
olegrok, line 64: Why not:
```lua
local ok, res, err = pcall(function()
    require('fiber').self().storage.console = nil
    {{ .FunctionBody }}
end)
```
?

cli/common/socket.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	if !data.Success {
+		return nil, fmt.Errorf("Failed to eval: %s", data.ErrStr)
+	}
+
+	return data.Data, nil
+}
+
+func processEvalTarantoolRes(resBytes []byte) (*TarantoolEvalRes, error) {
```
olegrok, line 116: Does this hack work with `\set output lua`?

Seems you need to explicitly enable yaml output... for Tarantool 2.2+

```
type 'help' for interactive help
tarantool> \set output lua
true;
tarantool> return {success = '123', error = 'No'}
{error = "No", success = "123"};
```

cli/repair/patch.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	if _, err := confFile.Write(newConfContent); err != nil {
+		return nil, fmt.Errorf("Failed to write a new config: %s", err)
+	}
+
+	return resMessages, nil
+}
+
+func reloadConf(topologyConfPath string, instanceName string, ctx *context.Ctx) ([]common.ResultMessage, error) {
```
olegrok, line 82: I hope we could move this function to separate file.

May be in future we will implement something like `cartridge reload` or it could be useful for https://github.com/tarantool/cartridge-cli/issues/249

cli/repair/patch.go
```diff
@@ -0,0 +1,174 @@
...
+				roles_configured_state, state
+			)
+		end
+
+		local current_uuid = box.info().uuid
+		if cfg:get_readonly().topology.servers[current_uuid] == nil then
+			return false
+		end
+
+		cfg:lock()
```
olegrok, line 139: AFAIK it should be done internally by cartridge. You don't really need it. Also you lock config here and don't unlock it.

cli/repair/patch.go
```diff
@@ -0,0 +1,174 @@
...
+
+		if state ~= roles_configured_state then
+			return nil, string.format(
+				'Failed to desire %s config state. Stuck in %s',
+				roles_configured_state, state
+			)
+		end
+
+		local current_uuid = box.info().uuid
+		if cfg:get_readonly().topology.servers[current_uuid] == nil then
```
olegrok, line 135: Seems this check could be performed before `with_state`

cli/common/socket.go
olegrok: Not relevant. yaml output enabled by default for connection

cli/repair/patch.go
olegrok: I've briefly looked at cartridge code. It seems to be ok.

cli/repair/patch.go
dokshina: Yes, but it seems that this function should be modified for this (or only some part will be specified).
I suggest moving this function later, when we understood, where it can be used,

olegrok: Well, ok

cli/repair/patch.go
```diff
@@ -0,0 +1,174 @@
...
+
+	// read greeting
+	if _, err := common.ReadFromConn(conn); err != nil {
+		return resMessages, fmt.Errorf("Failed to read greeting: %s", err)
+	}
+
+	// eval
+	evalFuncTmpl := `
+		local ClusterwideConfig = require('cartridge.clusterwide-config')
+		local cfg, err = ClusterwideConfig.load('{{ .ConfigPath }}')
```
olegrok, line 107: I think load should be performed only after `state` checks

cli/repair/repair.go
```diff
@@ -42,6 +42,8 @@ func SetLeader(ctx *context.Ctx) error {
 func Run(processConfFunc ProcessConfFuncType, ctx *context.Ctx, patchConf bool) error {
 	log.Debugf("Data directory is set to: %s", ctx.Running.DataDir)
 
+	// XXX: reload don't work for cartridge < 2.0
```
olegrok, line 45: Please provide "Not supported for ..." error message for such case.

dokshina: I'm writing the code that checks it right now =)

cli/commands/common.go
dokshina: Yes, I will change this behaviour to reload only with flag.

olegrok (approved):

cli/repair/patch.go
```diff
@@ -0,0 +1,169 @@
...
+
+	// eval
+	evalFuncTmpl := `
+		local ClusterwideConfig = require('cartridge.clusterwide-config')
+		local confapplier = require('cartridge.confapplier')
+
+		local roles_configured_state = 'RolesConfigured'
+		local connecting_fullmesh_state = 'ConnectingFullmesh'
+
+		local state = confapplier.wish_state(roles_configured_state, {{ .WishStateTimeout }})
```
rosik, line 107: It's better to wish_state right before apply_config.

cli/repair/patch.go
```diff
@@ -0,0 +1,169 @@
...
+
+		if state == connecting_fullmesh_state then
+			return nil, string.format(
+				'Failed to desire %s config state. Stuck in %s. ' ..
```
rosik, line 111: I think 'Failed to reach ..' is more appropriate

dokshina: Fixed in #361

cli/repair/patch.go
dokshina: Fixed in #361

PR #378, opened 2020-10-12
Cartridge Admin

dokshina:
I didn't forget about

- [x] Tests
- [x] Changelog #382 
- [x] Documentation #381 

Closes #322

PR #379, opened 2020-10-13
Refactor test projects files

dokshina:
In some tests application files (such as `init.lua`) are rewritten completely.
I think it's time to store such files in a separated directory and provide a function to easily replace files in test projects.
It will be used in #378 to rewrite admin role for different tests.

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -102,6 +102,11 @@ func Lint() error {
 		return err
 	}
 
+	fmt.Println("Running luacheck for test projects files...")
+	if err := sh.RunV(".rocks/bin/luacheck", "test/files"); err != nil {
+		return err
+	}
+
 	return nil
 }
 
--- a/test/files/init_no_cartridge.lua
+++ b/test/files/init_no_cartridge.lua
@@ -11,10 +11,10 @@ local tnt_version = string.split(_TARANTOOL, '.')
 local tnt_major = tonumber(tnt_version[1])
 local tnt_minor = tonumber(tnt_version[2])
 if tnt_major < 2 or (tnt_major == 2 and tnt_minor < 2) then
-  local notify_socket = os.getenv('NOTIFY_SOCKET')
-  if notify_socket then
-      local socket = require('socket')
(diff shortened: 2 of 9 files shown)


test/project.py
```diff
@@ -389,3 +388,9 @@ def patch_cartridge_version(project, new_version):
 
     with open(os.path.join(project.path, 'init.lua'), 'a') as f:
         f.write(new_version_code)
+
+
+def rewrite_project_file(project, project_filepath, filepath):
```
olegrok, line 393: It's not clear that does it mean - seems arguments should have more verbose names.

Why do you rewrite content of file... Why can't you drop previous file and copy new one?

magefile.go
```diff
@@ -102,6 +102,11 @@ func Lint() error {
 		return err
 	}
 
+	fmt.Println("Running luacheck for test projects files...")
+	if err := sh.RunV(".rocks/bin/luacheck", "test/files"); err != nil {
```
olegrok, line 106: What error does user see if there is not luacheck installed?

test/files/init_no_cartridge.lua
```diff
@@ -0,0 +1,20 @@
...
+require('log').info('I am starting...')
+
+-- Copied from cartridge.cfg to provide support for NOTIFY_SOCKET in old tarantool
+local tnt_version = string.split(_TARANTOOL, '.')
+local tnt_major = tonumber(tnt_version[1])
+local tnt_minor = tonumber(tnt_version[2])
+if tnt_major < 2 or (tnt_major == 2 and tnt_minor < 2) then
+  local notify_socket = os.getenv('NOTIFY_SOCKET')
+  if notify_socket then
+      local socket = require('socket')
```
olegrok, line 16: There are 4 spaces here but there are only 2 spaces at the previous line

dokshina: Oh, copy-paste...

magefile.go
dokshina: The same as errors where `python3` or `go` isn't installed.

test/project.py
dokshina: Refactored.

olegrok (approved):

PR #391, opened 2020-10-29
Allow to specify Tarantool version on packing

### LINKED ISSUE #88: Ability to choose a version of Tarantool for packaging in a docker image
Currently, executing the `cartridge pack docker` command creates a docker image with the same version of Tarantool as on the host machine.

I propose to give the opportunity to choose which version Tarantool will be installed in the final image.

I think it would be convenient to do this with an additional flag, and if it is not provided then you can use the latest available version.

dokshina:
I didn't forget about

- [ ] Tests
- [ ] Changelog
- [ ] Documentation

See #88

### DIFF
--- a/cli/build/build.go
+++ b/cli/build/build.go
@@ -86,7 +86,7 @@ func checkCtx(ctx *context.Ctx) error {
 			return fmt.Errorf("TmpDir is missed")
 		}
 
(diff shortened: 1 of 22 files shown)


dokshina:
We decided to delay developing this feature.

PR #407, opened 2020-11-23
Move local run directory to tempdir

### LINKED ISSUE #406: [4pt] Max Socket PATH problem
It's a very popular problem when `cartridge start` fails with confusion error because inctance socket path is too long.

Possible solution is to put run directory into `~/.cartridge/tmp` dir (it's used for packing now).

Each application run directory is placed in `~/.cartridge/tmp/<hash(app-path)>`. 
`~/.cartridge/tmp` can be rewritten via `CARTRIDGE_TEMPDIR` env variable.

dokshina:
CARTRIDGE_TEMPDIR was used before on packing
Now, each application run directory is placed in
~/.cartridge/tmp/run-<project-ID>.
Project ID is a hash by project path.
This approach allows to avoid max socket path problem

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Closes #406

### DIFF
--- a/test/integration/running/test_clean.py
+++ b/test/integration/running/test_clean.py
@@ -1,36 +1,47 @@
 import os
 
-from utils import get_instance_id, get_stateboard_name
-from utils import DEFAULT_CFG, DEFAULT_DATA_DIR, DEFAULT_LOG_DIR, DEFAULT_RUN_DIR
 from utils import write_conf
 from utils import check_instances_running
 from utils import check_instances_stopped
 
 
-CARTRIDGE_CONF = '.cartridge.yml'
-
 FILES_TO_BE_DELETED = ['log', 'workdir', 'console-sock', 'notify-sock', 'pid']
 
 
-def get_instance_files(project, instance_id, log_dir=DEFAULT_LOG_DIR, data_dir=DEFAULT_DATA_DIR,
-                       run_dir=DEFAULT_RUN_DIR):
+def get_instance_files(project, instance_name, log_dir=None, data_dir=None, run_dir=None):
+    run_dir = project.get_run_dir(run_dir)
+
     return {
-        'log': os.path.join(project.path, log_dir, '%s.log' % instance_id),
-        'workdir': os.path.join(project.path, data_dir, instance_id),
-        'console-sock': os.path.join(project.path, run_dir, '%s.control' % instance_id),
-        'pid': os.path.join(project.path, run_dir, '%s.pid' % instance_id),
-        'notify-sock': os.path.join(project.path, run_dir, '%s.notify' % instance_id),
+        'log': project.get_log_dir(instance_name, log_dir),
+        'workdir': project.get_workdir(instance_name, data_dir),
+        'console-sock': project.get_console_sock(instance_name, run_dir),
+        'pid': project.get_pidfile(instance_name, run_dir),
+        'notify-sock': project.get_notify_sock(instance_name, run_dir),
     }
 
 
-def create_instances_files(project, instance_names=[],
-                           log_dir=DEFAULT_LOG_DIR, data_dir=DEFAULT_DATA_DIR, run_dir=DEFAULT_RUN_DIR):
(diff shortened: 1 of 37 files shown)


test/project.py
```diff
@@ -73,6 +86,10 @@ def __init__(self, cartridge_cmd, name, basepath, template='cartridge', create_f
         else:
             self.path = create_func(basepath)
 
+        home = str(Path.home())
```
mRrvz, line 89: This is str call not needed here. os.path.join can work with pathlib module.

test/integration/running/test_clean.py
```diff
@@ -1,36 +1,47 @@
...
-def get_instance_files(project, instance_id, log_dir=DEFAULT_LOG_DIR, data_dir=DEFAULT_DATA_DIR,
-                       run_dir=DEFAULT_RUN_DIR):
+def get_instance_files(project, instance_name, log_dir=None, data_dir=None, run_dir=None):
+    run_dir = project.get_run_dir(run_dir)
+
+    return {
+        'log': project.get_log_dir(instance_name, log_dir),
+        'workdir': project.get_workdir(instance_name, data_dir),
+        'console-sock': project.get_console_sock(instance_name, run_dir),
+        'pid': project.get_pidfile(instance_name, run_dir),
```
mRrvz, line 18: pid and pidfile is slightly different things

test/integration/running/test_clean.py
```diff
@@ -1,36 +1,47 @@
...
     return {
-        'log': os.path.join(project.path, log_dir, '%s.log' % instance_id),
-        'workdir': os.path.join(project.path, data_dir, instance_id),
-        'console-sock': os.path.join(project.path, run_dir, '%s.control' % instance_id),
-        'pid': os.path.join(project.path, run_dir, '%s.pid' % instance_id),
-        'notify-sock': os.path.join(project.path, run_dir, '%s.notify' % instance_id),
+        'log': project.get_sb_log_dir(log_dir),
+        'workdir': project.get_sb_workdir(data_dir),
+        'console-sock': project.get_sb_console_sock(run_dir),
+        'pid': project.get_sb_pidfile(run_dir),
```
mRrvz, line 30: same comment above

test/integration/running/test_clean.py
```diff
@@ -1,36 +1,47 @@
...
+        'notify-sock': project.get_sb_notify_sock(run_dir),
     }
 
 
-def create_instances_files(project, instance_names=[],
-                           log_dir=DEFAULT_LOG_DIR, data_dir=DEFAULT_DATA_DIR, run_dir=DEFAULT_RUN_DIR):
-    instance_ids = [get_instance_id(project.name, instance_name) for instance_name in instance_names]
-    instance_ids.append(get_stateboard_name(project.name))
+def create_instances_files(project, instance_names=[], stateboard=False, log_dir=None, data_dir=None, run_dir=None):
+    all_files = [
```
mRrvz, line 36: Passing a large number of arguments in this form is not very convenient and is difficult to scale. I suggest using ``** kwargs``  with the ``.setdefault`` method.

test/integration/running/test_clean.py
```diff
@@ -1,36 +1,47 @@
...
-from utils import get_instance_id, get_stateboard_name
-from utils import DEFAULT_CFG, DEFAULT_DATA_DIR, DEFAULT_LOG_DIR, DEFAULT_RUN_DIR
 from utils import write_conf
 from utils import check_instances_running
 from utils import check_instances_stopped
 
 
-CARTRIDGE_CONF = '.cartridge.yml'
-
 FILES_TO_BE_DELETED = ['log', 'workdir', 'console-sock', 'notify-sock', 'pid']
```
mRrvz, line 8: For constants it's better to use tuples: ``('log', 'workdir', 'console-sock', 'notify-sock', 'pid')``

dokshina:
We've decided that this change is so huge that is would be introduces in a new CLI in the future.

Issue #396, opened 2020-11-06
[16pt] Topology Management


dokshina:
See RFC #388

dokshina:
Closed via #408, #409, #412, #414, #416, #417

PR #408, opened 2020-11-24
Manage Topology: Command-line

dokshina:
All commands that allows to manage topology from command-line are implemented.

This is the first part of #396 

I didn't forget about

- [x] Tests 
- [ ] Changelog
- [x] Documentation #412

### DIFF
--- a/cli/common/utils.go
+++ b/cli/common/utils.go
@@ -13,6 +13,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/adam-hanna/arrayOperations"
+	"github.com/tarantool/cartridge-cli/cli/context"
 	"gopkg.in/yaml.v2"
 )
 
@@ -133,13 +135,13 @@ func IntsToStrings(numbers []int) []string {
 
 // ParseYmlFile reads YAML file and returns it's content as a map
 func ParseYmlFile(path string) (map[string]interface{}, error) {
-	fileContent, err := GetFileContent(path)
+	fileContent, err := GetFileContentBytes(path)
 	if err != nil {
 		return nil, fmt.Errorf("Failed to read file: %s", err)
 	}
 
 	res := make(map[string]interface{})
-	if err := yaml.Unmarshal([]byte(fileContent), res); err != nil {
+	if err := yaml.Unmarshal(fileContent, res); err != nil {
 		return nil, fmt.Errorf("Failed to parse %s: %s", path, err)
 	}
 
@@ -239,9 +241,9 @@ Loop:
 }
 
 func ConvertToStringsSlice(s interface{}) ([]string, error) {
-	sliceRaw, ok := s.([]interface{})
-	if !ok {
-		return nil, fmt.Errorf("Value should be an array, got %#v", s)
+	sliceRaw, err := ConvertToSlice(s)
+	if err != nil {
+		return nil, err
 	}
 
 	stringsSlice := make([]string, len(sliceRaw))
@@ -255,6 +257,35 @@ func ConvertToStringsSlice(s interface{}) ([]string, error) {
 	}
 
 	return stringsSlice, nil
+}
+
+func ConvertToMapWithStringKeys(raw interface{}) (map[string]interface{}, error) {
+	rawMap, ok := raw.(map[interface{}]interface{})
+	if !ok {
+		return nil, fmt.Errorf("Should be a map with string keys, got %#v", raw)
+	}
+
+	mapWithStringKeys := make(map[string]interface{})
+
+	for keyRaw, valueRaw := range rawMap {
+		keyString, ok := keyRaw.(string)
+		if !ok {
+			return nil, fmt.Errorf("Has non-string key: %#v", keyRaw)
+		}
+
+		mapWithStringKeys[keyString] = valueRaw
+	}
+
+	return mapWithStringKeys, nil
+}
+
+func ConvertToSlice(raw interface{}) ([]interface{}, error) {
+	iterfacesSlice, ok := raw.([]interface{})
+	if !ok {
+		return nil, fmt.Errorf("Should be a list, got %#v", raw)
+	}
+
+	return iterfacesSlice, nil
 }
 
 func StringsSliceElemIndex(s []string, elem string) int {
@@ -279,3 +310,43 @@ func InsertInStringSlice(s []string, i int, elem string) []string {
 
 	return res
 }
+
+func GetInstancesFromArgs(args []string, ctx *context.Ctx) ([]string, error) {
+	foundInstances := make(map[string]struct{})
+	var instances []string
+
+	for _, instanceName := range args {
+		if instanceName == ctx.Project.Name {
+			return nil, fmt.Errorf(appNameSpecifiedError)
+		}
+
+		parts := strings.SplitN(instanceName, ".", 2)
+
+		if len(parts) > 1 {
+			return nil, fmt.Errorf(instanceIDSpecified)
+		}
+
+		if instanceName != "" {
+			if _, found := foundInstances[instanceName]; found {
+				return nil, fmt.Errorf("Duplicate instance name specified: %s", instanceName)
+			}
+
+			instances = append(instances, instanceName)
+			foundInstances[instanceName] = struct{}{}
+		}
+	}
+
+	return instances, nil
+}
+
+func GetStringSlicesDifference(s1, s2 []string) []string {
+	uniqueStrings := arrayOperations.DifferenceString(s1, s2)
+	return arrayOperations.IntersectString(s1, uniqueStrings)
+}
+
+const (
+	appNameSpecifiedError = "Application name is specified. " +
+		"Please, specify instance name(s)"
+	instanceIDSpecified = `[APP_NAME].INSTANCE_NAME is specified. ` +
+		"Please, specify instance name(s)"
+)
--- a/cli/replicasets/bootstrap_vshard.go
+++ b/cli/replicasets/bootstrap_vshard.go
@@ -1,0 +1,52 @@
+package replicasets
+
+import (
+	"fmt"
+	"net"
+	"strings"
+
+	"github.com/apex/log"
+	"github.com/tarantool/cartridge-cli/cli/context"
+
+	"github.com/tarantool/cartridge-cli/cli/common"
+)
+
+func BootstrapVshard(ctx *context.Ctx, args []string) error {
+	conn, err := connectToSomeJoinedInstance(ctx)
+	if err != nil {
+		return err
+	}
+
+	if err := bootstrapVshard(conn); err != nil {
+		return fmt.Errorf("failed to bootstrap vshard: %s", err)
+	}
+
+	log.Infof("Vshard is bootstrapped successfully")
+
+	return nil
+}
+
+func bootstrapVshard(conn net.Conn) error {
+	_, err := common.EvalTarantoolConn(conn, bootstrapVshardBody)
+	if err != nil {
+		if strings.Contains(err.Error(), `Sharding config is empty`) {
+			// XXX: see https://github.com/tarantool/cartridge/issues/1148
+			log.Warnf(
+				`It's possible that there is no running instances of some configured vshard groups. ` +
+					`In this case existing storages are bootstrapped, but Cartridge returns an error`,
+			)
+		}
+
+		return err
+	}
+
+	return nil
+}
+
+var (
+	bootstrapVshardBody = `
+local cartridge = require('cartridge')
+local ok, err = cartridge.admin_bootstrap_vshard()
+return ok, err
+`
+)
--- a/cli/replicasets/common.go
+++ b/cli/replicasets/common.go
@@ -1,0 +1,178 @@
+package replicasets
+
+import (
+	"fmt"
+	"net"
+	"os"
+	"strings"
+
+	"github.com/apex/log"
+	"github.com/tarantool/cartridge-cli/cli/common"
+	"github.com/tarantool/cartridge-cli/cli/context"
+	"github.com/tarantool/cartridge-cli/cli/project"
+	"github.com/tarantool/cartridge-cli/cli/running"
+	"gopkg.in/yaml.v2"
+)
+
+type InstanceConf struct {
+	URI string `yaml:"advertise_uri"`
+}
+
+type InstancesConf map[string]*InstanceConf
+
+// connectToSomeRunningInstance connects to some running instance.
+// It's used for some actions that can be performed via any instance socket,
+// no matter if this instance is joined to cluster or not.
+// For example, to get known roles or vhsard groups list.
+func connectToSomeRunningInstance(ctx *context.Ctx) (net.Conn, error) {
+	instancesConf, err := getInstancesConf(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get instances configuration: %s", err)
+	}
+
+	runningInstancesNames := getRunningInstances(instancesConf, ctx)
+	if len(runningInstancesNames) == 0 {
+		return nil, fmt.Errorf("No running instances found")
+	}
+
+	instanceName := runningInstancesNames[0]
+	conn, err := connectToInstance(instanceName, ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	return conn, nil
+}
+
+// connectToSomeJoinedInstance connects to some instance joined to cluster.
+// It's used for actions with joined cluster, e.g. setting replicaset parameters.
+func connectToSomeJoinedInstance(ctx *context.Ctx) (net.Conn, error) {
+	instancesConf, err := getInstancesConf(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get instances configuration: %s", err)
+	}
+
+	joinedInstanceName, err := getJoinedInstanceName(instancesConf, ctx)
+	if err != nil || joinedInstanceName == "" {
+		return nil, fmt.Errorf("Failed find some instance joined to cluster")
+	}
+
+	conn, err := connectToInstance(joinedInstanceName, ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	return conn, nil
+}
+
+func getInstancesConf(ctx *context.Ctx) (*InstancesConf, error) {
+	var err error
+
+	log.Debugf("Instances configuration file is %s", ctx.Running.RunDir)
+
+	if _, err := os.Stat(ctx.Running.ConfPath); err != nil {
+		return nil, fmt.Errorf("Failed to use instances configuration file: %s", err)
+	}
+
+	fileContentBytes, err := common.GetFileContentBytes(ctx.Running.ConfPath)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to read instances configuration file: %s", err)
+	}
+
+	var allSectionsConf InstancesConf
+	if err := yaml.Unmarshal([]byte(fileContentBytes), &allSectionsConf); err != nil {
+		return nil, fmt.Errorf("Failed to parse instances configuration file %s: %s", ctx.Running.ConfPath, err)
+	}
+
+	instancesConf := make(InstancesConf)
+
+	appInstancePrefix := fmt.Sprintf("%s.", ctx.Project.Name)
+	for key, sectionConf := range allSectionsConf {
+		if strings.HasPrefix(key, appInstancePrefix) {
+			parts := strings.SplitN(key, ".", 2)
+			instanceName := parts[1]
+			instancesConf[instanceName] = sectionConf
+		}
+	}
+
+	return &instancesConf, nil
+}
+
+// getJoinedInstanceName returns a name of instancethat is already joined to cluster
+// It gets all membership instances and checks if there is some instance that has
+// UUID (it means that this instance is joined to cluster).
+// The main reason of using membership here is that instances that aren't joined
+// can see joined instances only if they are connected to the one membership.
+func getJoinedInstanceName(instancesConf *InstancesConf, ctx *context.Ctx) (string, error) {
+	membershipInstances, err := getMembershipInstances(instancesConf, ctx)
+	if err != nil {
+		return "", err
+	}
+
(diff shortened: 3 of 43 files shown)


cli/common/socket.go
```diff
@@ -91,7 +91,7 @@ func EvalTarantoolConn(conn net.Conn, funcBody string) (interface{}, error) {
 	end
 
 	if err ~= nil then
-		return { success = false, err = err }
+		return { success = false, err = tostring(err) }
```
olegrok, line 94: Could be moved into separate patch as completely.

cli/common/utils.go
```diff
@@ -238,9 +240,14 @@ Loop:
 	return lastNewLinePos, nil
 }
 
-func ConvertToStringsSlice(s []interface{}) ([]string, error) {
-	stringsSlice := make([]string, len(s))
-	for i, elem := range s {
+func ConvertToStringsSlice(s interface{}) ([]string, error) {
```
olegrok, line 243: I think a proper name is "ConvertArrayToStringSlice" but I don't insist

cli/common/utils.go
```diff
@@ -274,3 +281,43 @@ func InsertInStringSlice(s []string, i int, elem string) []string {
...
+
+		parts := strings.SplitN(instanceName, ".", 2)
+
+		if len(parts) > 1 {
+			return nil, fmt.Errorf(instanceIDSpecified)
+		}
+
+		if instanceName != "" {
+			if _, found := foundInstances[instanceName]; found {
+				return nil, fmt.Errorf("Duplicate instance name: %s", instanceName)
```
olegrok, line 302: "... found in config"?

Otherwise it could confuse me as I don't understand the root of problem.

cli/replicasets/bootstrap_vshard.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+
+		return err
+	}
+
+	return nil
+}
+
+var (
+	bootstrapVshardBody = `
```
olegrok, line 52: Not sure but maybe it's reasonable to call "can_bootstrap_vshard" before (see https://github.com/tarantool/cartridge/blob/e4ecef6fb2176889e4ebcdb99d8a19cbee70c876/cartridge/vshard-utils.lua#L416)

cli/replicasets/common.go
```diff
@@ -0,0 +1,185 @@
...
+
+	var allSectionsConf InstancesConf
+	if err := yaml.Unmarshal([]byte(fileContentBytes), &allSectionsConf); err != nil {
+		return nil, fmt.Errorf("Failed to parse instances configuration file %s: %s", ctx.Running.ConfPath, err)
+	}
+
+	instancesConf := make(InstancesConf)
+
+	appInstancePrefix := fmt.Sprintf("%s.", ctx.Project.Name)
+	for key, sectionConf := range allSectionsConf {
```
olegrok, line 45: Should we ignore/consider stateboard here and at https://github.com/tarantool/cartridge-cli/pull/408/files#diff-4fb23dc0ae21422ebe8be6bd721532279e77e346e549a5294f020b4c7443fa62R285 ?

cli/replicasets/common.go
```diff
@@ -0,0 +1,185 @@
...
+		}
+	}
+
+	return &instancesConf, nil
+}
+
+func getControlConn(instancesConf *InstancesConf, ctx *context.Ctx, joinInstances []string) (net.Conn, error) {
+	controlInstanceName, err := getJoinedInstanceName(instancesConf, ctx)
+	if err != nil {
+		return nil, fmt.Errorf("Failed find some instance joined to custer")
```
olegrok, line 59: `custer ` -> `cluster `

cli/replicasets/common.go
```diff
@@ -0,0 +1,185 @@
...
+			parts := strings.SplitN(key, ".", 2)
+			instanceName := parts[1]
+			instancesConf[instanceName] = sectionConf
+		}
+	}
+
+	return &instancesConf, nil
+}
+
+func getControlConn(instancesConf *InstancesConf, ctx *context.Ctx, joinInstances []string) (net.Conn, error) {
```
olegrok, line 56: Not sure `getControlConn` should have `joinInstances ` argument. Consider a better name for this function please

cli/replicasets/common.go
```diff
@@ -0,0 +1,185 @@
...
+
+	log.Debugf("Connected to %s", consoleSockPath)
+
+	return conn, nil
+}
+
+func getJoinedInstanceName(instancesConf *InstancesConf, ctx *context.Ctx) (string, error) {
+	joinedInstances, err := getJoinedInstances(instancesConf, ctx)
+	if err != nil {
+		return "", fmt.Errorf("Failed to get instances connected to membership: %s", err)
```
olegrok, line 86: `connected to membership`

I'm not sure it's correct phrase. May be it should be `in membership` (see https://github.com/tarantool/cartridge/blob/dbd7e3385e9afb13f8e754eb562e9d46e60d3217/cartridge/lua-api/get-topology.lua#L156)

cli/replicasets/common.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	var joinedInstanceName string
+	for instanceURI, instance := range *joinedInstances {
+		if instance.UUID != "" {
+			if instance.Alias == "" {
+				return "", fmt.Errorf("Failed to get alias for instance %s", instanceURI)
+			}
```
olegrok, line 94: `if instance.UUID == "" && instance.Alias == ""  { ... }`

or it's valid case when instance doesn't have UUID?

cli/replicasets/common.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	instanceName := runningInstancesNames[0]
+	instanceSocketPath := project.GetInstanceConsoleSock(ctx, runningInstancesNames[0])
+	conn, err := common.ConnectToTarantoolSocket(instanceSocketPath)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to connect to instance %s: %s", instanceName, err)
+	}
+
+	log.Debugf("Connect all replicasets instances to membership")
```
olegrok, line 117: The same as above. Here and below.

cli/replicasets/common.go
```diff
@@ -0,0 +1,185 @@
...
+
+			joinedInstanceName = instance.Alias
+			break
+		}
+	}
+
+	return joinedInstanceName, nil
+}
+
+func getJoinedInstances(instancesConf *InstancesConf, ctx *context.Ctx) (*MembershipInstances, error) {
```
olegrok, line 104: To be honest I don't understand how "join" is related to "membership".

You can found "joined" instances in cartridge topology config. Also seems you need rename "connect to membership" to "probe".

cli/replicasets/common.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+func convertToSlice(raw interface{}) ([]interface{}, error) {
+	iterfacesSlice, ok := raw.([]interface{})
+	if !ok {
+		return nil, fmt.Errorf("Should be a list, got %#v", raw)
+	}
+
+	return iterfacesSlice, nil
+}
```
olegrok, line 185: Consider moving this code to "utils"

cli/replicasets/edit_topology.go
```diff
@@ -0,0 +1,235 @@
...
+
+type EditInstanceOpts struct {
+	InstanceUUID string
+	Expelled     bool
+}
+
+type EditInstancesOpts []*EditInstanceOpts
+
+func editReplicaset(conn net.Conn, opts *EditReplicasetOpts) (*TopologyReplicaset, error) {
+	// wait for replicaset is healthy
```
olegrok, line 36: It's not clear how this phrase is related to "editReplicaset" function name

cli/replicasets/edit_topology.go
```diff
@@ -0,0 +1,235 @@
...
+{{ .FormatTopologyReplicasetFunc }}
+
+local res, err = cartridge.admin_edit_topology({
+	replicasets = {
+		{ {{ .ReplicasetInput }} },
+	}
+})
+
+if err ~= nil then
+	return nil, err
```
olegrok, line 204: This error is an object from "errors" module. Do you consider it?

cli/replicasets/edit_topology.go
```diff
@@ -0,0 +1,235 @@
...
+	joinServerStrings := make([]string, len(instancesURIs))
+	for i, instancesURI := range instancesURIs {
+		joinServerStrings[i] = fmt.Sprintf("{ uri = '%s' }", instancesURI)
+	}
+
+	optString := fmt.Sprintf("%s = { %s }", optName, strings.Join(joinServerStrings, ", "))
+	*optsStrings = append(*optsStrings, optString)
+}
+
+func serializeStringsSlice(stringsSlice []string) string {
```
olegrok, line 182: I'm sure you need to write general purpose serializer or find ready solution.

Just remember about one place where similar code exists.
https://github.com/tarantool/cartridge-cli/blob/a7387abfe26e5578f97c0bda77d2e120fd57256a/cli/admin/call.go#L136

cli/replicasets/join.go
```diff
@@ -0,0 +1,106 @@
...
+	topologyReplicasets *TopologyReplicasets, instancesConf *InstancesConf) (*EditReplicasetOpts, error) {
+
+	editReplicasetOpts := EditReplicasetOpts{}
+
+	topologyReplicaset := topologyReplicasets.GetByAlias(replicasetAlias)
+	if topologyReplicaset != nil {
+		editReplicasetOpts.ReplicasetUUID = topologyReplicaset.UUID
+	} else {
+		editReplicasetOpts.ReplicasetAlias = replicasetAlias
+	}
```
olegrok, line 83: Why can't you return error here. This case looks strange.

cli/replicasets/join.go
```diff
@@ -0,0 +1,106 @@
...
+
+	return &editReplicasetOpts, nil
+}
+
+func getInstancesURIs(instanceNames []string, instancesConf *InstancesConf) (*[]string, error) {
+	instancesURIs := make([]string, len(instanceNames))
+	for i, instanceName := range instanceNames {
+		instanceConf, found := (*instancesConf)[instanceName]
+		if !found {
+			return nil, fmt.Errorf("Configuration for instance %s isn't found in %s", instanceName, instancesFile)
```
olegrok, line 99: Please check `in` or `at` should be here.

cli/replicasets/list.go
```diff
@@ -0,0 +1,9 @@
+package replicasets
+
+import (
+	"github.com/tarantool/cartridge-cli/cli/context"
+)
+
+func List(ctx *context.Ctx, args []string) error {
+	return nil
```
olegrok, line 8: Where is a code? :)

cli/replicasets/membership.go
```diff
@@ -0,0 +1,134 @@
+package replicasets
```
olegrok, line 1: Membership is an implementation detail of cartridge. You don't have any guarantees it will work in future or it won't be changed to built-in "swim" module. I'm strictly against this module, I'm strictly against usage of membership. Please, use only cartridge API.

cli/replicasets/topology_replicasets.go
```diff
@@ -0,0 +1,333 @@
...
+
+func getFloatValueFromMap(m map[string]interface{}, key string, valuePtr *float64) error {
+	valueRaw, found := m[key]
+	if !found {
+		return nil
+	}
+
+	var value float64
+
+	switch i := valueRaw.(type) {
```
olegrok, line 225: I think it could be rewrite in more laconic way with reflect.

All int values you can cast with https://golang.org/pkg/reflect/#Value.SetInt and then to float. All float values you can convent with https://golang.org/pkg/reflect/#Value.SetFloat

Also please revise again is it really needed. E.g. json always decode values to float https://golang.org/pkg/encoding/json/#Token (`float64, for JSON numbers`)

cli/replicasets/bootstrap_vshard.go
dokshina: I was thinking about it, but `can_bootstrap_vshard` returns just `false` if there is no router or all groups are bootstrapped. But `bootstrap_vshard` returns an error. In fact, `can_bootstrap_vshard` is needed only for front-end to decide if `Bootstrap vshard` button should be drawn or not.

cli/replicasets/common.go
dokshina: Stateboard isn't a part of topology, so I think it should be ignored.
`common. GetInstancesFromArgs` function is also used for starting/stopping. Stateboard isn't passed as an argument, it should be specified by flag.

cli/replicasets/weight.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	instancesConf, err := getInstancesConf(ctx)
+	if err != nil {
+		return fmt.Errorf("Failed to get instances configuration: %s", err)
+	}
+
+	conn, err := getControlConn(instancesConf, ctx, nil)
+	if err != nil {
+		return fmt.Errorf("Failed to connect to Tarantool instance: %s", err)
```
olegrok, line 36: I'm tired to see such errors. Consider moving connection to tarantool directly to command context. You use it anywhere.

cli/replicasets/bootstrap_vshard.go
olegrok: Ok

cli/replicasets/common.go
olegrok: Thanks for explanation

cli/common/utils.go
dokshina: In fact, it takes interface argument, not array.

cli/common/utils.go
dokshina: `Duplicate instance name specified`? It's parsing arguments specified by user.

cli/replicasets/common.go
dokshina: This function finds membership instances that have UUIDs (these instances are joined to cluster). 

I think that I need to rework function names and write more comments about this. I will do it.

cli/replicasets/edit_topology.go
dokshina: Sometimes I will write a separate module for serialization.

cli/replicasets/common.go
```diff
@@ -0,0 +1,185 @@
...
+		if len(joinInstances) > 0 {
+			controlInstanceName = joinInstances[0]
+		}
+	}
+
+	if controlInstanceName == "" {
+		return nil, fmt.Errorf("Failed to find joined instance")
+	}
+
+	consoleSockPath := project.GetInstanceConsoleSock(ctx, controlInstanceName)
```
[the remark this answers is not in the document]
dokshina: I don't think so. All running instances joined to cluster have available sockets.

cli/replicasets/edit_topology.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	editReplicasetBody, err := templates.GetTemplatedStr(&editReplicasetBodyTemplate, map[string]string{
+		"ReplicasetInput":                  replicasetInput,
+		"FormatTopologyReplicasetFuncName": formatTopologyReplicasetFuncName,
+		"FormatTopologyReplicasetFunc":     formatTopologyReplicasetFunc,
+	})
+
+	if err != nil {
+		return nil, project.InternalError("Failed to get edit topology body by template: %s", err)
```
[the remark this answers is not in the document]
dokshina: Ok, I will fix it. But it's internal error. All that you need to understand is that it's my mistake.

cli/replicasets/membership.go
dokshina: Please, show me the way to do this things without using a membership and I will rewrite this code.

cli/repair/clusterwide_config.go
```diff
@@ -288,7 +288,7 @@ func (topologyConf *TopologyConfType) setReplicasetsConf() error {
 				replicasetConf.LeadersIsString = true
 				replicasetConf.Leaders = append(replicasetConf.Leaders, leadersConverted)
 			case []interface{}:
-				leaders, err := common.ConvertToStringsSlice(leadersConverted)
+				leaders, err := common.ConvertToStringsSlice(leadersRaw)
```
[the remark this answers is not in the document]
dokshina: See #411

cli/common/socket.go
dokshina: See #410

cli/replicasets/edit_topology.go
dokshina: Yes, `tostring(err)` is called for every error before returning it from `EvalTarantoolConn`.

cli/replicasets/join.go
dokshina: This is very important code (I will add a comment):

* if replicaset already exists, we have to specify it's URI to join new instances to it
* if replicaset doesn't exist, we specify it's alias to create a new replicaset with specified instances

cli/common/utils.go
olegrok: Yes, but you can convert only arrays

cli/common/utils.go
olegrok: My bad. Agree

cli/replicasets/edit_topology.go
olegrok: sometimes means "иногда". But seems you want to say "one day" (однажды).

cli/repair/clusterwide_config.go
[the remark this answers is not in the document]
olegrok: Great

cli/replicasets/common.go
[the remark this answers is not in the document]
olegrok: Ok

cli/replicasets/join.go
olegrok: Yes, please, do

cli/replicasets/membership.go
olegrok: Well, it was discussed verbally. Ok.

cli/replicasets/weight.go
dokshina: I've added `connectToSomeRunningInstance` and `connectToSomeJoinedInstance` functions with comments.

cli/replicasets/common.go
dokshina: This function is completely removed, see new `common` functions.

cli/common/utils.go
dokshina: And function returns error if specified interface isn't array. But in fact it converts interfaces.

cli/common/utils.go
```diff
@@ -257,6 +259,35 @@ func ConvertToStringsSlice(s interface{}) ([]string, error) {
...
+	if !ok {
+		return nil, fmt.Errorf("Should be a map with string keys, got %#v", raw)
+	}
+
+	mapWithStringKeys := make(map[string]interface{})
+
+	for keyRaw, valueRaw := range rawMap {
+		keyString, ok := keyRaw.(string)
+		if !ok {
+			return nil, fmt.Errorf("Has non string key: %#v", keyRaw)
```
olegrok, line 273: `non string` -> `non-string`

cli/replicasets/common.go
```diff
@@ -0,0 +1,178 @@
...
+			instanceName := parts[1]
+			instancesConf[instanceName] = sectionConf
+		}
+	}
+
+	return &instancesConf, nil
+}
+
+// getJoinedInstanceName returns a name of instancethat is already joined to cluster
+// It gets all membership instances and checks if there is come instance that has
```
olegrok, line 102: `come` -> `some`

cli/replicasets/expel.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	instancesToExpelMap := make(map[string]bool)
+	for _, instanceName := range instancesToExpelNames {
+		instancesToExpelMap[instanceName] = false
+	}
+
+	joinedInstanceName := ""
+
+	// We need to find some  instance that is joined and not specified to expel
```
olegrok, line 39: `some  instance` -> `some instance`

cli/replicasets/replicasets.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	if ctx.Running.AppDir, err = filepath.Abs(ctx.Running.AppDir); err != nil {
+		return fmt.Errorf("Failed to get application directory absolute path: %s", err)
+	}
+
+	if ctx.Project.Name == "" {
+		if ctx.Project.Name, err = project.DetectName(ctx.Running.AppDir); err != nil {
+			return fmt.Errorf(
+				"Failed to detect application name: %s. Please pass it explicitly via --name ",
```
olegrok, line 38: `via --name ",` -> `via --name",`

cli/replicasets/roles.go
```diff
@@ -0,0 +1,186 @@
...
+	newTopologyReplicaset, err := editReplicaset(conn, editReplicasetOpts)
+	if err != nil {
+		return fmt.Errorf("Failed to update roles list: %s", err)
+	}
+
+	if len(newTopologyReplicaset.Roles) == 0 {
+		log.Infof("Now replicaset %s has no roles enabled", ctx.Replicasets.ReplicasetName)
+	} else {
+		log.Infof(
+			"Replicaset %s now has these roles enabled: ",
```
olegrok, line 146: `roles enabled: "` -> `roles enabled:"`

doc/replicasets.rst
```diff
@@ -0,0 +1,226 @@
+.. _cartridge-cli.replicasets:
+
+===============================================================================
+Setting up replicasets
+===============================================================================
+
+``cartridge replicasets`` command is used to set up replicasets on local
```
olegrok, line 7: `on development` -> `for development`?

cli/replicasets/join.go
dokshina: I've checked - `in`.

doc/replicasets.rst
dokshina: Doc moved to #412 and will be fixed separately.

olegrok (approved):

PR #409, opened 2020-11-24
Manage Topology: File

dokshina:
Commands that allows to manage topology described in a file..

This is the second part of #396 

I didn't forget about

- [x] Tests
- [ ] Changelog
- [x] Documentation #412

### DIFF
--- a/cli/common/result.go
+++ b/cli/common/result.go
@@ -13,6 +13,8 @@ const (
 	ResStatusSkipped
 	ResStatusFailed
 	ResStatusExited
+	ResStatusUpdated
+	ResStatusCreated
 )
 
 type ResMessageType int
@@ -105,4 +107,6 @@ func init() {
 	resStrings[ResStatusSkipped] = ColorWarn.Sprintf("SKIPPED")
 	resStrings[ResStatusFailed] = ColorErr.Sprintf("FAILED")
 	resStrings[ResStatusExited] = ColorErr.Sprintf("EXITED")
+	resStrings[ResStatusCreated] = ColorYellow.Sprintf("CREATED")
+	resStrings[ResStatusUpdated] = ColorYellow.Sprintf("UPDATED")
 }
--- a/cli/replicasets/edit_topology.go
+++ b/cli/replicasets/edit_topology.go
@@ -25,15 +25,20 @@ type EditReplicasetOpts struct {
 	FailoverPriorityUUIDs []string
 }
 
+type EditReplicasetsListOpts []*EditReplicasetOpts
+
 type EditInstanceOpts struct {
 	InstanceUUID string
 	Expelled     bool
 }
 
-type EditInstancesOpts []*EditInstanceOpts
+type EditInstancesListOpts []*EditInstanceOpts
 
-func editReplicaset(conn net.Conn, opts *EditReplicasetOpts) (*TopologyReplicaset, error) {
-	replicasetInput := serializeEditReplicasetOpts(opts)
+func editReplicasetsList(conn net.Conn, opts *EditReplicasetsListOpts) (*TopologyReplicasets, error) {
+	replicasetsInput, err := serializeEditReplicasetsListOpts(opts)
+	if err != nil {
+		return nil, err
+	}
 
 	formatTopologyReplicasetFunc, err := templates.GetTemplatedStr(
 		&formatTopologyReplicasetFuncTemplate, map[string]string{
@@ -45,8 +50,8 @@ func editReplicaset(conn net.Conn, opts *EditReplicasetOpts) (*TopologyReplicase
 		return nil, project.InternalError("Failed to compute get topology replicaset function body: %s", err)
 	}
 
-	editReplicasetBody, err := templates.GetTemplatedStr(&editReplicasetBodyTemplate, map[string]string{
-		"ReplicasetInput":                  replicasetInput,
+	editReplicasetsBody, err := templates.GetTemplatedStr(&editReplicasetsBodyTemplate, map[string]string{
+		"ReplicasetsInput":                 replicasetsInput,
 		"FormatTopologyReplicasetFuncName": formatTopologyReplicasetFuncName,
 		"FormatTopologyReplicasetFunc":     formatTopologyReplicasetFunc,
 	})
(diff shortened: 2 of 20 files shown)


cli/replicasets/setup.go
```diff
@@ -0,0 +1,287 @@
...
+	var err error
+
+	if err := FillCtx(ctx); err != nil {
+		return err
+	}
+
+	if ctx.Replicasets.File == "" {
+		ctx.Replicasets.File = defaultReplicasetsFile
+	}
+	if ctx.Replicasets.File, err = filepath.Abs(ctx.Replicasets.File); err != nil {
```
olegrok, line 46: Does this function return an error when file doesn't exist?

If answer "yes", it's non-informative to see `Failed to get replicasets configuration file absolute path` instead of `file "123.yml" doesn't exist`

cli/replicasets/setup.go
```diff
@@ -0,0 +1,287 @@
...
+
+	i := 0
+	for replicasetAlias, replicasetConf := range replicasetsConf {
+		replicasetConf.Alias = replicasetAlias
+
+		replicasetsList[i] = replicasetConf
+		i++
+	}
+
+	// move vshard-router replicasets at the list beginning
```
olegrok, line 175: I guess such approach is a bit wrong.

You shouldn't sort array for that. You need to join ALL instances at the same time but not by one.
edit_topology accepts array of instances.

cli/replicasets/setup.go
```diff
@@ -0,0 +1,287 @@
...
+	// to prevent "no remotes with role vshard-router avaliable" error
+	// that happens when vshard-router replicaset is set up last
+	// Probably, this relate to https://github.com/tarantool/cartridge/issues/1139
```
olegrok, line 178: Even if it's true it is already fixed: https://github.com/tarantool/cartridge/issues/1139. However, I don't think it's related.

cli/replicasets/setup.go
dokshina: No, this function just computes a path (using simple string-concatenate rules). `doesn't exists` error is returned later on trying to open this file.

cli/replicasets/setup.go
dokshina: I join replicasets one-by-one, not instances. The one reason is that in old versions of Cartridge joining more than one replicaset causes an error. The other is that it's much more simple to manage replicasets one-by-one (believe me, I've done it many times). About moving router to the beginning - it's not the best solution, but could you suggest something else to prevent vshard bootstrapping error?

cli/replicasets/setup.go
dokshina: Yes, but this version of cartridge isn't released yet. Even if it was released, what about users who didn't update cartridge right after release?

cli/replicasets/setup.go
olegrok: Is there test for this case?

cli/replicasets/setup.go
olegrok: Currently I suggest to use some timeout for that.

cli/replicasets/setup.go
olegrok: > The one reason is that in old versions of Cartridge joining more than one replicaset causes an error. 

Please show me commit/PR where it was fixed

cli/replicasets/setup.go
dokshina: Added a test.

cli/common/result.go
```diff
@@ -105,4 +107,6 @@ func init() {
 	resStrings[ResStatusSkipped] = ColorWarn.Sprintf("SKIPPED")
 	resStrings[ResStatusFailed] = ColorErr.Sprintf("FAILED")
 	resStrings[ResStatusExited] = ColorErr.Sprintf("EXITED")
+	resStrings[ResStatusCreated] = ColorYellow.Sprintf("CREATED")
```
olegrok, line 110: Why not "ColorGreen"?

cli/replicasets/edit_topology.go
```diff
@@ -89,7 +109,28 @@ func editInstances(conn net.Conn, opts *EditInstancesOpts) (bool, error) {
...
+		"ReplicasetsInput": strings.Join(replicasetsResults, ", "),
+	})
+	if err != nil {
+		return "", project.InternalError("Failed to serialize edit replicasets opts: %s", err)
+	}
+
+	return res, nil
+}
+
+func serializeEditReplicasetOpts(opts *EditReplicasetOpts) (string, error) {
```
olegrok, line 133: Do you really need it?

Seems it's the same as `editReplicasetsOptsTemplate` but with single element.

Also `editReplicasetOptsTemplate` and `editReplicasetsOptsTemplate` are quite similar next person who will be change this code could be confused.

`editReplicasetOptsTemplate` + `editReplicasetListOptsTemplate` if you really need it.

cli/replicasets/setup.go
```diff
@@ -0,0 +1,304 @@
...
+
+func logSetupSummary(topologyReplicasets, newTopologyReplicasets *TopologyReplicasets) {
+	for replicasetUUID, newTopologyReplicaset := range *newTopologyReplicasets {
+		replicasetRes := common.Result{
+			ID: newTopologyReplicaset.Alias,
+		}
+
+		if oldTopologyReplicaset, found := (*topologyReplicasets)[replicasetUUID]; !found {
+			replicasetRes.Status = common.ResStatusCreated
+		} else if reflect.DeepEqual(newTopologyReplicaset, oldTopologyReplicaset) {
```
olegrok, line 180: If nothing occurred it's OK, am I right? Why not "not modified"?

cli/replicasets/setup.go
```diff
@@ -0,0 +1,304 @@
...
+
+	i := 0
+	for replicasetAlias, replicasetConf := range replicasetsConf {
+		replicasetConf.Alias = replicasetAlias
+
+		replicasetsList[i] = replicasetConf
+		i++
+	}
+
+	if len(replicasetsList) == 0 {
```
olegrok, line 217: Seems it's the same as `len(replicasetsConf) == 1` and could be done above.

cli/replicasets/setup.go
dokshina: Completely rewrote this part, now all replicasets are managed by single `edit_topology` call.

cli/common/result.go
dokshina: I was thinking about Ansible logs:

`changed` - yellow (we have two different `changed`: `created` and `updated`)
`ok` - not changed, green

cli/replicasets/setup.go
dokshina: See comment above.

cli/replicasets/setup.go
dokshina: You mean `len(replicasetsConf) == 0`? Yes, you are right.

olegrok: yes, `== 0`

olegrok (approved):
`With great power comes great responsibility`

Use my approve only for good.

PR #410, opened 2020-12-02
Refactor error handling on socket  eval

dokshina:
This patch fixes two problems:
* function can return non-string error, `tostring(err)` is performed before return now
* parsing result yaml error was confusing, now it shows `Unmarshal` error


olegrok (approved):
Thanks for your patch. LGTM.

PR #411, opened 2020-12-02
Refactor common.ConvertToStringsSlice function

dokshina:
Old function signature required `s.([]interface{})` converting before each call. 
I decided that it should be performed in function itself.


olegrok:
Please reflect your motivation of this changes in commit message

dokshina:
Added a description (it will be a part of merge commit message).

olegrok (approved):

PR #414, opened 2020-12-08
Topology management: List

dokshina:
- [x] Tests
- [ ] Changelog
- [ ] Documentation

### DIFF
--- a/cli/replicasets/list.go
+++ b/cli/replicasets/list.go
@@ -1,0 +1,162 @@
+package replicasets
+
+import (
+	"fmt"
+	"sort"
+	"strconv"
+	"strings"
+
+	"github.com/apex/log"
+	"github.com/tarantool/cartridge-cli/cli/common"
+	"github.com/tarantool/cartridge-cli/cli/context"
+)
+
+const (
+	instanceMarker       = "•"
+	leaderInstanceMarker = "★"
+)
+
+func List(ctx *context.Ctx, args []string) error {
+	var err error
+
+	if err := FillCtx(ctx); err != nil {
+		return err
+	}
+
+	conn, err := connectToSomeJoinedInstance(ctx)
+	if err != nil {
+		return err
+	}
+
+	topologyReplicasets, err := getTopologyReplicasets(conn)
+	if err != nil {
+		return fmt.Errorf("Failed to get current topology replica sets: %s", err)
+	}
+
+	replicasetsSummary := getTopologyReplicasetsSummary(topologyReplicasets)
+
+	log.Infof("Current replica sets:\n%s", replicasetsSummary)
+
+	return nil
+}
+
+func getTopologyReplicasetsSummary(topologyReplicasets *TopologyReplicasets) string {
+	// sort replicasets by aliases
+	replicasetsList := make([]*TopologyReplicaset, len(*topologyReplicasets))
+	i := 0
+	for _, topologyReplicaset := range *topologyReplicasets {
+		replicasetsList[i] = topologyReplicaset
+		i++
+	}
+
+	sort.Slice(replicasetsList, func(i, j int) bool {
+		return replicasetsList[i].Alias < replicasetsList[j].Alias
+	})
+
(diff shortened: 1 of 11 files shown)


cli/replicasets/list.go
```diff
@@ -0,0 +1,138 @@
...
+	return nil
+}
+
+func getTopologyReplicasetsSummary(topologyReplicasets *TopologyReplicasets) string {
+	replicasetsSummaryByAliases := make(map[string]string)
+	replicasetsAliases := make([]string, len(*topologyReplicasets))
+
+	// sort replicasets summaries by aliases
+
+	i := 0
```
olegrok, line 49: This block looks big:
- Get all aliases
- Sort them
- Create an array

I suggest:
- Create an array
- Sort by name

Should be more laconically

cli/replicasets/list.go
```diff
@@ -0,0 +1,138 @@
...
+		formattedWeight := strconv.FormatFloat(*topologyReplicaset.Weight, 'f', -1, 64)
+		additionalInfo = append(additionalInfo, formattedWeight)
+	}
+	if topologyReplicaset.AllRW != nil && *(topologyReplicaset.AllRW) {
+		additionalInfo = append(additionalInfo, "ALL RW")
+	}
+
+	if len(additionalInfo) > 0 {
+		replicasetTitle = fmt.Sprintf(
+			"%-30s    %s",
```
olegrok, line 90: May be we should relay on console width? (https://stackoverflow.com/questions/16569433/get-terminal-size-in-go)

cli/replicasets/list.go
```diff
@@ -0,0 +1,138 @@
...
+			replicasetTitle,
+			common.ColorHiBlue.Sprint(strings.Join(additionalInfo, " | ")),
+		)
+	}
+
+	// roles
```
olegrok, line 96: It would be great to extend comments with an example how it looks like:
```
 Role: my-role1 | my-role2
```

Of course this comment is relevant for ALL comments and places where you format servers/replicasets.

cli/replicasets/list.go
dokshina: Discussed and decided that current solution is OK.

cli/replicasets/list.go
dokshina: This part is rewritten.

cli/replicasets/list.go
dokshina: Added comments and unit tests.

test/integration/replicasets/test_list_replicasets.py
```diff
@@ -0,0 +1,105 @@
...
+    project = project_with_instances.project
+
+    # list replicasets
+    cmd = [
+        cartridge_cmd, 'replicasets', 'list',
+    ]
+
+    rc, output = run_command_and_get_output(cmd, cwd=project.path)
+    assert rc == 1
+    assert "Failed find some instance joined to cluster" in output
```
olegrok, line 52: `Failed find` -> `Failed to find`?

test/integration/replicasets/utils.py
```diff
@@ -39,7 +39,10 @@ def __init__(self, cli, project, instances_list, replicasets_list=[], vshard_gro
                 }
             })
 
-        write_conf(os.path.join(project.path, DEFAULT_CFG), instances_conf)
+        instances_conf_path = os.path.join(project.path, DEFAULT_CFG)
+        if not os.path.exists(instances_conf_path):
+            write_conf(os.path.join(project.path, DEFAULT_CFG), instances_conf)
```
olegrok, line 44:
```suggestion
            write_conf(instances_conf_path, instances_conf)
```

dokshina: Sorry, I've commited this change without applying your suggestion.

olegrok: I don't resent you in the least

olegrok (approved):

PR #416, opened 2020-12-09
Topology Management: Old Cartridge Support

dokshina:

### DIFF
--- a/cli/common/socket.go
+++ b/cli/common/socket.go
@@ -17,6 +17,19 @@ const (
 	endOfTarantoolOutput = "\n...\n"
 )
 
+var (
+	evalTarantoolConnTimeout time.Duration
+)
+
+func init() {
+	// In fact, sometimes we need to disable this timeout on production.
+	// For example, if replica is joined to a big storage.
+	// In this case Cartridge waits for loading all data.
+	// But since `cartridge replicasets` command is developed for local
+	// running, we ignore this case for a while.
+	evalTarantoolConnTimeout = 1 * time.Minute
+}
+
 func ConnectToTarantoolSocket(socketPath string) (net.Conn, error) {
 	conn, err := net.Dial("unix", socketPath)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


cli/common/socket.go
```diff
@@ -39,6 +39,7 @@ func ReadFromConn(conn net.Conn) ([]byte, error) {
 	data := make([]byte, 0)
 
 	for {
+		conn.SetReadDeadline(time.Now().Add(10 * time.Second))
```
olegrok, line 42: How does it work? Could we place this line before the cycle?

Also IMHO 10 seconds it's quite huge timeout for console application. 3-4 seconds are enough but it's up to you

cli/common/tarantool.go
```diff
@@ -91,6 +92,31 @@ func GetNextMajorVersion(versionStr string) (string, error) {
 	return strconv.Itoa(major + 1), nil
 }
 
+func GetMajorCartridgeVersion(conn net.Conn) (int, error) {
```
olegrok, line 95: How is "scm" handled here?

olegrok: Also AFAIK "VERSION" generation it's quite new feature.

olegrok: But no, It's not a new feature. But "scm-1" is possible (https://github.com/tarantool/cartridge/blob/master/cartridge/VERSION.lua)

cli/replicasets/common.go
```diff
@@ -176,3 +176,12 @@ func connectToInstance(instanceName string, ctx *context.Ctx) (net.Conn, error)
 
 	return conn, nil
 }
+
+func needToWaitForHealthy(conn net.Conn) (bool, error) {
```
olegrok, line 180: `needToWaitForHealthy ` calls `GetMajorCartridgeVersion `. Is it mistake?

cli/common/tarantool.go
dokshina: Oh, thanks a lot! I will research.

cli/replicasets/common.go
dokshina: No, I was chosen between this approach and calling `GetMajorCartridgeVersion` in each place `editReplicasets` is called.

cli/common/socket.go
dokshina: I'm not sure about 3-4 second because this time depends on duration of Cartridge operation performed by this call. And it can be long sometimes (new Cartridge joins replica sets synchronously).

`SetReadDeadline` says when (since now) read should fail. I think this line can be placed before cycle, you are right.

cli/common/tarantool.go
dokshina: For 1.x `nil` is returned. In this case `1` is returned.
For `scm-1` version `2` is returned.

olegrok (approved):

PR #419, opened 2020-12-17
Enter and Connect commands

### LINKED ISSUE #317: [8pt] `cartridge enter/connect` command
I sure that we need an alternative for `tarantoolctl connect ./tmp/run/myapp.router.control`.

I suggest adding `cartridge connect INSTANCE_NAME` command.

dokshina:
I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Closes #317

### DIFF
--- a/cli/commands/connect.go
+++ b/cli/commands/connect.go
@@ -1,0 +1,52 @@
+package commands
+
+import (
+	"github.com/apex/log"
+	"github.com/spf13/cobra"
+	"github.com/tarantool/cartridge-cli/cli/connect"
+)
+
+func init() {
+	var enterCmd = &cobra.Command{
+		Use:   "enter INSTANCE_NAME",
+		Short: "Enter to application instance console",
+		Run: func(cmd *cobra.Command, args []string) {
+			if err := connect.Enter(&ctx, args); err != nil {
+				log.Fatalf(err.Error())
+			}
+		},
+		ValidArgsFunction: ShellCompRunningInstances,
+		Args:              cobra.MaximumNArgs(1),
+	}
+
+	rootCmd.AddCommand(enterCmd)
+
+	// FLAGS
+	configureFlags(enterCmd)
+
+	// application name flag
+	addNameFlag(enterCmd)
+	// run-dir flag
+	enterCmd.Flags().StringVar(&ctx.Running.RunDir, "run-dir", "", runDirUsage)
+
+	var connectCmd = &cobra.Command{
+		Use:   "connect URI",
+		Short: "Connect to specified URI",
+		Run: func(cmd *cobra.Command, args []string) {
+			if err := connect.Connect(&ctx, args); err != nil {
+				log.Fatalf(err.Error())
+			}
+		},
+		Args: cobra.MaximumNArgs(1),
+	}
+
+	rootCmd.AddCommand(connectCmd)
+
+	// FLAGS
+	configureFlags(connectCmd)
+
+	// username flag
+	connectCmd.Flags().StringVarP(&ctx.Connect.Username, "username", "u", "", connectUsernameUsage)
+	// password flag
+	connectCmd.Flags().StringVarP(&ctx.Connect.Password, "password", "p", "", connectPasswordUsage)
+}
--- a/cli/connect/binary.go
+++ b/cli/connect/binary.go
@@ -1,0 +1,95 @@
+package connect
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/FZambia/tarantool"
+	"github.com/apex/log"
+	"github.com/c-bata/go-prompt"
+	"github.com/tarantool/cartridge-cli/cli/common"
+)
+
+func binaryConnect(console *Console) error {
+	var err error
+
+	connectStr := fmt.Sprintf("%s://%s", console.connOpts.Network, console.connOpts.Address)
+	console.binaryConn, err = tarantool.Connect(connectStr, tarantool.Opts{
+		User:           console.connOpts.Username,
+		Password:       console.connOpts.Password,
+		RequestTimeout: common.EvalTarantoolConnTimeout,
+	})
+	if err != nil {
+		return fmt.Errorf("Failed to connect: %s", err)
+	}
+
+	return nil
+}
+
+func binaryEval(console *Console, funcBody string, args ...interface{}) (interface{}, error) {
+	if args == nil {
+		args = []interface{}{}
+	}
+
+	resp, err := console.binaryConn.Exec(tarantool.Eval(funcBody, args))
+
+	if err != nil {
+		return nil, err
+	}
+
+	if len(resp.Data) == 0 {
+		return nil, nil
+	}
+
+	dataRaw := resp.Data[0]
+	return dataRaw, nil
+}
+
+func getBinaryCompleter(console *Console) prompt.Completer {
+	getSuggestionsBinary := func(console *Console, lastWord string) interface{} {
+		res, err := console.Eval(
+			getSuggestionsBinaryFuncBody,
+			lastWord, len(lastWord),
+		)
+
+		if err != nil {
+			return nil
+		}
+
+		return res
+	}
+
+	completer := func(in prompt.Document) []prompt.Suggest {
+		return getSuggestions(console, in, getSuggestionsBinary)
+	}
+
+	return completer
+}
+
+func binaryExecute(console *Console, in string) string {
+	dataRaw, err := console.Eval(
+		"return require('console').eval(...)",
+		in,
+	)
+
+	if err == io.EOF {
+		log.Fatalf("Connection was closed. Probably instance process isn't running anymore")
+	}
+	if err != nil {
+		log.Fatalf("Failed to eval: %s", err)
+	}
+
+	data, ok := dataRaw.(string)
+	if !ok {
+		log.Fatalf("Failed to eval: Data received in wrong format")
+	}
+
+	return data
+}
+
+const (
+	getSuggestionsBinaryFuncBody = `
+local last_word, last_word_len = ...
+return require('console').completion_handler(last_word, 0, last_word_len)
+`
+)
--- a/cli/connect/common.go
+++ b/cli/connect/common.go
@@ -1,0 +1,114 @@
+package connect
+
+import (
+	"sort"
+	"strings"
+
+	"github.com/adam-hanna/arrayOperations"
+	"github.com/c-bata/go-prompt"
+	"github.com/tarantool/cartridge-cli/cli/common"
+	"github.com/tarantool/cartridge-cli/cli/context"
+)
+
+const (
+	MaxHistoryLines = 10000
+
+	TCPNetwork  = "tcp"
+	UnixNetwork = "unix"
+)
+
+type ConnOpts struct {
+	Network  string
+	Address  string
+	Username string
+	Password string
+}
+
+type GetRawSuggestionsFunc func(console *Console, lastWord string) interface{}
+
+func getConnOpts(connString string, ctx *context.Ctx) (*ConnOpts, error) {
+	connOpts := ConnOpts{
+		Username: ctx.Connect.Username,
+		Password: ctx.Connect.Password,
+	}
+
+	connStringParts := strings.SplitN(connString, "@", 2)
+	address := connStringParts[len(connStringParts)-1]
+
+	if len(connStringParts) > 1 {
+		authString := connStringParts[0]
+		authStringParts := strings.SplitN(authString, ":", 2)
+
+		if connOpts.Username == "" {
+			connOpts.Username = authStringParts[0]
+		}
+		if len(authStringParts) > 1 && connOpts.Password == "" {
+			connOpts.Password = authStringParts[1]
+		}
+	}
+
+	addrLen := len(address)
+	switch {
+	case addrLen > 0 && (address[0] == '.' || address[0] == '/'):
+		connOpts.Network = UnixNetwork
+		connOpts.Address = address
+	case addrLen >= 7 && address[0:7] == "unix://":
+		connOpts.Network = UnixNetwork
+		connOpts.Address = address[7:]
+	case addrLen >= 5 && address[0:5] == "unix:":
+		connOpts.Network = UnixNetwork
+		connOpts.Address = address[5:]
+	case addrLen >= 6 && address[0:6] == "unix/:":
+		connOpts.Network = UnixNetwork
+		connOpts.Address = address[6:]
+	case addrLen >= 6 && address[0:6] == "tcp://":
+		connOpts.Network = TCPNetwork
+		connOpts.Address = address[6:]
+	case addrLen >= 4 && address[0:4] == "tcp:":
+		connOpts.Network = TCPNetwork
+		connOpts.Address = address[4:]
+	default:
+		connOpts.Network = TCPNetwork
+		connOpts.Address = address
+	}
+
+	return &connOpts, nil
+}
+
+func getSuggestions(console *Console, in prompt.Document,
+	getRawSuggestionsFunc GetRawSuggestionsFunc) []prompt.Suggest {
+
+	if len(in.Text) == 0 {
+		return nil
+	}
+
+	lastWordStart := in.FindStartOfPreviousWordUntilSeparator(tarantoolWordSeparators)
+	lastWord := in.Text[lastWordStart:]
+
+	if len(lastWord) == 0 {
+		return nil
+	}
+
+	suggestionsRaw := getRawSuggestionsFunc(console, lastWord)
+
+	suggestionsTexts, err := common.ConvertToStringsSlice(suggestionsRaw)
+	if err != nil {
+		return nil
+	}
+
+	suggestionsTexts = arrayOperations.DifferenceString(suggestionsTexts)
+	if len(suggestionsTexts) == 0 {
+		return nil
+	}
+
+	sort.Strings(suggestionsTexts)
+
+	suggestions := make([]prompt.Suggest, len(suggestionsTexts))
+	for i, suggestionText := range suggestionsTexts {
+		suggestions[i] = prompt.Suggest{
+			Text: suggestionText,
+		}
+	}
+
+	return suggestions
+}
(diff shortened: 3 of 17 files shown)


cli/connect/console.go
```diff
@@ -0,0 +1,352 @@
...
+	default:
+		return nil, fmt.Errorf("Unknown protocol: %s", console.protocol)
+	}
+
+	return evalFunc(console, funcBody, args...)
+
+}
+
+func loadHistory(console *Console) {
+	var err error
```
mRrvz, line 158: This variable declaration is not necessary

cli/connect/console.go
```diff
@@ -0,0 +1,352 @@
...
+		return
+	}
+
+	console.historyFilePath = filepath.Join(homeDir, HistoryFileName)
+
+	console.historyLines, err = common.GetLastNLines(console.historyFilePath, MaxHistoryLines)
+	if err != nil {
+		return
+	}
+}
```
mRrvz, line 171: Maybe this function should return error or log it?

cli/connect/console.go
```diff
@@ -0,0 +1,352 @@
...
+
+		console.input = ""
+		console.livePrefixEnabled = false
+	}
+
+	return executor, nil
+}
+
+func inputIsCompleted(input string, luaState *lua.LState) bool {
+	var err error
```
mRrvz, line 242: This variable declaration is not necessary

cli/connect/console.go
```diff
@@ -0,0 +1,352 @@
...
+	}
+
+	if ConsoleOutputMode(inWords[2]) == ConsoleYAMLOutput {
+		console.outputMode = ConsoleYAMLOutput
+		return
+	}
+
+	if ConsoleOutputMode(inWords[2]) == ConsoleLuaOutput {
+		console.outputMode = ConsoleLuaOutput
+		return
```
mRrvz, line 289: This  return is not necessary

cli/connect/console.go
```diff
@@ -0,0 +1,352 @@
...
+			prompt.ASCIICodeBind{ // move to one word left
+				ASCIICode: []byte{0x1b, 0x62},
+				Fn: func(buf *prompt.Buffer) {
+					d := buf.Document()
+					wordLen := len([]rune(d.GetWordBeforeCursorWithSpace()))
+					buf.CursorLeft(wordLen)
+				},
+			},
+			prompt.ASCIICodeBind{ // move to one word right
+				ASCIICode: []byte{0x1b, 0x66},
```
mRrvz, line 316: Im not sure, but its not clear what these bytes are. Maybe add some comments?

cli/connect/console.go
dokshina: This variable is needed here

```go
// 1
console.historyLines, err = common.GetLastNLines(console.historyFilePath, MaxHistoryLines)
```

In fact, it's created anyway here

```go
// 2
homeDir, err := common.GetHomeDir()
```

but I don't like when some line (1) depend on such variable that is created by some other line (2).

cli/connect/console.go
dokshina: The commands history is a very good feature, but it's not required to work with interactive console.
So, error shouldn't be returned.

But you are right, I will add debug log messages on console creation.

cli/connect/console.go
dokshina: This return makes this `if` block the same as `if` block above. Moreover, it allows to add any code below without changing this block and prevents future mistakes.

cli/connect/console.go
dokshina: Moved to variables.

test/conftest.py
```diff
@@ -22,11 +23,9 @@
...
 from utils import DEFAULT_RUN_DIR
 
-
-FILES_DIR = 'test/files'
-INIT_NO_CARTRIDGE_FILEPATH = os.path.join(FILES_DIR, 'init_no_cartridge.lua')
-INIT_IGNORE_SIGTERM_FILEPATH = os.path.join(FILES_DIR, 'init_ignore_sigterm.lua')
-INIT_ADMIN_FUNCS_FILEPATH = os.path.join(FILES_DIR, 'init_admin_funcs.lua')
+from project import INIT_NO_CARTRIDGE_FILEPATH
+from project import INIT_IGNORE_SIGTERM_FILEPATH
+from project import INIT_ADMIN_FUNCS_FILEPATH
```
mRrvz, line 28: I'm not sure, but PEP8 doesn't welcome such imports. 

```python
from project import (INIT_NO_CARTRIDGE_FILEPATH, 
     INIT_IGNORE_SIGTERM_FILEPATH, ...)
```

Maybe something like this?

test/integration/connect/conftest.py
```diff
@@ -0,0 +1,96 @@
...
+import os
+
+from project import Project
+from project import patch_cartridge_proc_titile
+from project import remove_dependency
+from project import replace_project_file
+from project import INIT_NO_CARTRIDGE_FILEPATH
+
+from utils import DEFAULT_CFG, DEFAULT_RPL_CFG
+from utils import ProjectWithTopology, Instance
```
mRrvz, line 12: why imports from ``project`` looks different from ``utils``? import from utils seems to me (and pep8) better

test/integration/connect/conftest.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+@pytest.fixture(scope="function")
+def project_with_instances(built_project, start_stop_cli, request):
+    cli = start_stop_cli
+    project = built_project
+
+    router = Instance('router', 8081, 'localhost:3301')
+    s1_master = Instance('s1-master', 8082, 'localhost:3302')
+    s1_replica = Instance('s1-replica', 8083, 'localhost:3303')
```
mRrvz, line 64: Just an idea (not sure): forward this constants to something like ``@dataclass``

test/integration/connect/utils.py
```diff
@@ -0,0 +1,93 @@
...
+        stdout=subprocess.PIPE,
+        stdin=subprocess.PIPE,
+    )
+
+    user_input = '\n'.join([c.command for c in commands]) + '\n'
+
+    out, _ = process.communicate(user_input.encode('utf-8'))
+    out = out.decode('utf-8')
+
+    print(out)
```
mRrvz, line 63: This print can be removed

cli/commands/connect.go
```diff
@@ -0,0 +1,57 @@
...
+		},
+		Args: cobra.MaximumNArgs(1),
+	}
+
+	rootCmd.AddCommand(connectCmd)
+
+	// FLAGS
+	configureFlags(connectCmd)
+
+	// username flag
```
olegrok, line 48: username/password is not a flag, it's an option. Flag usually is a boolean value

cli/connect/binary.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/c-bata/go-prompt"
+	"github.com/tarantool/go-tarantool"
+)
+
+func binaryConnect(console *Console) error {
+	var err error
+
+	console.binaryConn, err = tarantool.Connect(console.connOpts.Address, tarantool.Opts{
+		User: console.connOpts.Username,
+		Pass: console.connOpts.Password,
```
olegrok, line 17: AFAIR you use timeouts for unix sockets. I think you should use it for tcp sockets as well

cli/connect/connect.go
```diff
@@ -0,0 +1,115 @@
...
+	"os"
+
+	"github.com/tarantool/cartridge-cli/cli/common"
+	"github.com/tarantool/cartridge-cli/cli/context"
+	"github.com/tarantool/cartridge-cli/cli/project"
+	"github.com/tarantool/cartridge-cli/cli/running"
+)
+
+const (
+	tarantoolWordSeparators = "\t\r\n !\"#$%&'()*+,-/;<=>?@[\\]^`{|}~"
```
olegrok, line 14: Please add a comment where you got such set of symbols.

cli/connect/console.go
```diff
@@ -0,0 +1,399 @@
...
+
+		if !inputIsCompleted(console.input, console.luaState) {
+			console.livePrefixEnabled = true
+			return
+		}
+
+		handleSetOutput(console, console.input)
+
+		if console.historyFilePath != "" {
+			common.AppendStringToFile(console.historyFilePath, console.input+"\n")
```
olegrok, line 243: Looks like you every time reopen the same file. I'm not sure but looks like possible place where your prompt will lag.

cli/connect/console.go
```diff
@@ -0,0 +1,399 @@
...
+
+		console.input = ""
+		console.livePrefixEnabled = false
+	}
+
+	return executor, nil
+}
+
+func inputIsCompleted(input string, luaState *lua.LState) bool {
+	if _, err := luaState.LoadString(input); err == nil || !strings.Contains(err.Error(), "at EOF") {
```
olegrok, line 258: `!strings.Contains(err.Error(), "at EOF")`

Please, describe it in comment

test/files/init_no_cartridge.lua
```diff
@@ -18,3 +20,9 @@ if tnt_major < 2 or (tnt_major == 2 and tnt_minor < 2) then
         sock:sendto('unix/', notify_socket, 'READY=1')
     end
 end
+
+local sock_path = os.getenv('TARANTOOL_CONSOLE_SOCK')
+if sock_path ~= nil then
+    local console = require('console')
+    assert(pcall(console.listen, sock_path))
+end
```
olegrok, line 28: `box.cfg{listen = ...}` should be tested as well.

test/integration/connect/conftest.py
```diff
@@ -0,0 +1,96 @@
...
+
+    p = ProjectWithTopology(
+        cli,
+        project,
+        instances_list=[router, s1_master, s1_replica],
+    )
+
+    request.addfinalizer(lambda: p.stop())
+
+    p.start()
```
olegrok, line 95: Seems you need to put "stop" after "start".

If start for some reasons fails, you call finalizer and double fail occurs - you tries to stop instance that didn't start.

test/integration/replicasets/utils.py
```diff
@@ -1,65 +1,3 @@
...
-        check_instances_running(self.cli, self.project, [name for name in self.instances], daemonized=True)
-
-    def stop(self):
-        self.cli.stop(self.project, force=True)
-        check_instances_stopped(self.cli, self.project, [name for name in self.instances])
-        self.cli.clean(self.project)
-
-        os.remove(os.path.join(self.project.path, DEFAULT_CFG))
-
-
```
olegrok, line 62: This changes should be moved into separate patch.

test/conftest.py
dokshina: I like your approach, but I suggest you to prettify our tests in separated patch. Please, create an issue and describe these problems.

cli/commands/connect.go
dokshina: I agree, but cobra framework calls all options flags. And I call them flags to prevent confuses.

test/integration/connect/utils.py
dokshina: This print is needed for `pytest -s`, it makes output more verbose. I will add a comment.

cli/connect/console.go
dokshina: Tarantool `console` uses this file too. What about a case when it is used at the same time?

cli/connect/console.go
dokshina: There are two comments below. Added one more with a link to `console` code.

test/integration/replicasets/utils.py
dokshina: Yes, I agree. But it's not enough. I would like test helpers and fixtures to be re-organized, since we added a lot of commands and tests become more complicated.

dokshina: See https://github.com/tarantool/cartridge-cli/pull/421

cli/connect/console.go
olegrok: In should be clarified additionally.

E.g. https://unix.stackexchange.com/questions/346062/concurrent-writing-to-a-log-file-from-many-processes

doc/connect.rst
```diff
@@ -0,0 +1,35 @@
...
+
+    cartridge enter INSTANCE_NAME [flags]
+
+Flags:
+
+* ``--name`` - application name
+* ``--run-dir`` - directory where PID and socket files are stored
+  (defaults to ./tmp/run or "run-dir" in .cartridge.yml)
+
+Connects to instance via it's1 console socket placed in ``run-dir``.
```
olegrok, line 21: `1` is seems to be misprint

test/integration/connect/utils.py
```diff
@@ -0,0 +1,113 @@
...
+        cmd,
+        cwd=project.path,
+        stderr=subprocess.STDOUT,
+        stdout=subprocess.PIPE,
+        stdin=subprocess.PIPE,
+    )
+
+    output, _ = process.communicate()
+    output = output.decode('utf-8')
+    print(output)
```
olegrok, line 110: Do you really need print here?

olegrok: `assert process.returncode == 1, 'Got: ' + output`

cli/connect/binary.go
dokshina: Added `common.EvalTarantoolConnTimeout`.

test/integration/connect/conftest.py
dokshina: No, `start` starts more that one instance, in case when some of them failed (`start` failed), we need to stop other started instances.

cli/connect/console.go
dokshina: Fixed.

test/integration/connect/utils.py
dokshina: Yea, I would like this output to be printed anyway. I often use it for local tesing.

cli/common/socket.go
```diff
@@ -75,6 +75,18 @@ func ReadFromConn(conn net.Conn, endOfOutput string) ([]byte, error) {
...
+func endOfOutputFound(data []byte, endOfOutput string) bool {
+	endOfOutputLen := len(endOfOutput)
+	dataLen := len(data)
+
+	if dataLen < endOfOutputLen {
+		return false
+	}
+
+	endOfData := data[dataLen-endOfOutputLen:]
+	return string(endOfData) == endOfOutput
```
olegrok, line 87: Maybe hasPostfix is better here?

olegrok: Of course we don't want to cast whole data to string (in case it leads to copy of whole data buffer). But this "endswith" implementation isn't clear enough IMO.

cli/common/socket.go
```diff
@@ -50,10 +50,10 @@ func ConnectToTarantoolSocket(socketPath string) (net.Conn, error) {
 }
 
 func ReadFromConn(conn net.Conn, endOfOutput string) ([]byte, error) {
-	tmp := make([]byte, 1024)
+	tmp := make([]byte, 1)
```
olegrok, line 53: Looks a bit strange. Why do you change it?

cli/common/socket.go
```diff
@@ -20,7 +20,7 @@ const (
 )
```
olegrok, line 20: Would be great to move this changes into separate commit. With good description and brilliant commit message.

cli/connect/common.go
```diff
@@ -0,0 +1,114 @@
...
+	Username string
+	Password string
+}
+
+type GetRawSuggestionsFunc func(console *Console, lastWord string) interface{}
+
+func getConnOpts(connString string, ctx *context.Ctx) (*ConnOpts, error) {
+	connOpts := ConnOpts{
+		Username: ctx.Connect.Username,
+		Password: ctx.Connect.Password,
```
olegrok, line 32: + `timeout`?

olegrok, line 32: \+ timeout?

dokshina: Timeout is a constant that doesn't depend on conn string.

cli/connect/plain_text.go
```diff
@@ -0,0 +1,151 @@
...
+	case console.outputMode == ConsoleLuaOutput:
+		return strings.HasPrefix(dataPortion, tagPushPrefixLua)
+	}
+
+	return false
+}
+
+func getReadFromConnSetMode(console *Console, newOutputMode ConsoleOutputMode) ReadFromConnFunc {
+	readFromConnFunc := func(conn net.Conn) ([]byte, error) {
+		setModeResBytes := make([]byte, 1024)
```
olegrok, line 112: Here you don't read byte by byte as here https://github.com/tarantool/cartridge-cli/pull/425. Is it ok?

dokshina: Yes, here I read a result of `\set output <mode>` command. If everything before was read correclty, it should be `true` value or error message.

cli/common/socket.go
dokshina: See #425

cli/connect/plain_text.go
dokshina: I will add a comment here, just because your Birthday tomorrow. I sure that my comment is a great gift.

olegrok: > I sure that my comment is a great gift

The best gift is a gift that is made by own hands.

cli/connect/connect.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	if ctx.Project.Name == "" {
+		if ctx.Project.Name, err = project.DetectName(ctx.Running.AppDir); err != nil {
+			return fmt.Errorf(
+				"Failed to detect application name: %s. Please pass it explicitly via --name",
+				err,
+			)
+		}
+	}
```
olegrok, line 96: There are at least three places with such code fragment. Consider some refactoring of FillCtx. Seems you need some GenericFillCtx to perform such checks.

Maybe not now but in this case please file an issue.

olegrok (approved):

cli/connect/connect.go
dokshina: See #427

PR #424, opened 2020-12-22
Creation of static file system with files required for the 'create cartridge' command

### LINKED ISSUE #413: Don't embed static files to .go files manually
There are an examples (e.g. templates) when we need to deliver some static files to user: https://github.com/tarantool/cartridge-cli/tree/master/cli/create/templates. 
Such approach looks a bit strange and incorrect.
There are some special tools (see https://github.com/shurcooL/vfsgen) that allows to do that properly.

mRrvz:
The ```cartridge create``` command now uses a virtual file system with a template that statically builds in Go code during the build phase of the project.

Closes #413


cli/create/codegen/static/content/build/Dockerfile.build.cartridge
```diff
@@ -0,0 +1,13 @@
...
+#
+# The base image must be centos:8
+FROM centos:8
+
+# Here you can install some packages required
+#   for your application build
+#
+# RUN set -x \
+#    && curl -sL https://rpm.nodesource.com/setup_10.x | bash - \
+#    && yum -y install nodejs
```
dokshina, line 13: Please, add a newlines here and below.

cli/create/templates/app_files.go
```diff
@@ -1,6 +1,8 @@
 package templates
 
-import "github.com/tarantool/cartridge-cli/cli/templates"
+import (
+	"github.com/tarantool/cartridge-cli/cli/templates"
+)
 
 var appFilesTemplate = templates.FileTreeTemplate{
```
dokshina, line 7: I think that now we don't need to declare `templates.FileTreeTemplate` manually. 
You can scan `content` directory tree and collect all files content, mode and path.

Imagine, that you've added a new file to application template. I think it would be great if adding a new file to templates directory is enough.

So, I suggest creating some directory (e.g.`./cli/create/templates/cartridge`, `cartridge` is a template name, it mean Cartridge application) that is exactly the same as a result application directory (except some templates like`{{ .Name }}`).

cli/create/codegen/static/content/build/cartridge.post-build
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+# Simple post-build script
+# Will be ran after "tarantoolctl rocks make" on application packing
+# Could be useful to remove some build artifacts from result package
+
+# For example:
+# rm -rf third_party
+# rm -rf node_modules
+# rm -rf doc
```
olegrok, line 10: git warns about missing empty lines at the end of file, please don't ignore it

magefile.go
```diff
@@ -148,6 +161,11 @@ func Build() error {
 	var err error
 
 	fmt.Println("Building...")
+
+	if err := generateGoCode(); err != nil {
```
dokshina, line 165: Maybe Magefile `Deps` will help you to avoid copy paste.

cli/create/templates/templates.go
```diff
@@ -85,6 +89,42 @@ func Instantiate(ctx *context.Ctx) error {
...
+			})
+		} else {
+			fileContent, err := getStaticFileContent(fs, filePath)
+			if err != nil {
+				return fmt.Errorf("Failed to get file content: %s", err)
+			}
+
+			tmpl.AddFiles(templates.FileTemplate{
+				Path:    filePath,
+				Mode:    0755,
```
dokshina, line 113: Is it possible to collect file modes on generating static filesystem and then use them here?

I mean smth like `fileModeByPaths["<file-path>"] = <file-mode>`.

cli/create/templates/templates.go
```diff
@@ -19,16 +23,19 @@ const (
...
-	)
+	/*
+		knownTemplates[CartridgeTemplateName] = templates.Combine(
+			appFilesTemplate,
+			buildFilesTemplate,
+			configFilesTemplate,
+			devFilesTemplate,
+			testFilesTemplate,
+		)
+	*/
```
dokshina, line 38: It can be removed.

magefile.go
```diff
@@ -25,6 +25,12 @@ var cliExe = "cartridge"
 var goPackageName = "github.com/tarantool/cartridge-cli/cli"
 var packagePath = "./cli"
 
+var generateModePath = packagePath + "/create/codegen/generate_mode.go"
```
dokshina, line 28: Please, use `filepath.Join`.

cli/create/codegen/generate_mode.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/apex/log"
+	"github.com/dave/jennifer/jen"
+)
+
+func generateFileModeFile(path string, filename string) error {
```
dokshina, line 11: I think that it would be great to add a comment that shows result file structure.

cli/create/codegen/generate_mode.go
```diff
@@ -0,0 +1,63 @@
...
+
+	fileModeMap, err := getFileModes(path)
+
+	if err != nil {
+		return err
+	}
+
+	f.Var().Id("FileModes").Op("=").Map(jen.String()).Int().Values(jen.DictFunc(func(d jen.Dict) {
+		for key, element := range fileModeMap {
+			d[jen.Lit(key)] = jen.Lit(element)
```
dokshina, line 23: Result file will contain decimal values (e.g. `493` for `0755`). I think it would be helpful in the future to add comments with filemodes in octal format.

cli/commands/create.go
```diff
@@ -60,6 +61,10 @@ func runCreateCommand(cmd *cobra.Command, args []string) error {
 		ctx.Create.Template = templates.CartridgeTemplateName
 	}
 
+	if ctx.Create.FileSystem == nil && ctx.Create.From == "" {
```
dokshina, line 64: I would like this code to be performed in `create` module.

cli/context/context.go
```diff
@@ -23,8 +26,9 @@ type ProjectCtx struct {
 }
 
 type CreateCtx struct {
-	Template string
-	From     string
+	FileSystem http.FileSystem
```
dokshina, line 29: maybe `TemplateFS`?

cli/create/codegen/static/static.go
```diff
@@ -0,0 +1,9 @@
+// +build dev
+
+package static
+
+//go:generate go run -tags=dev ../generate_vfs.go
+
+import "net/http"
+
+var CartridgeData http.FileSystem = http.Dir("../../templates/cartridge")
```
dokshina, line 9: What about `CartridgeTemplateFS` instead of `CartridgeData`?

cli/commands/create.go
```diff
@@ -60,6 +61,10 @@ func runCreateCommand(cmd *cobra.Command, args []string) error {
...
+		ctx.Create.FileSystem = static.CartridgeData
```
dokshina, line 65: In fact, `cartridge` template is just one of possible application templates (`create` command has `--template` flag, the default value is `cartridge`).
Sometimes we will add more templates (I believe).

So, generally, we choose Cartridge template filesystem only if `ctx.Create.Template` is `cartridge`. Otherwise, it should be an error.

magefile.go
```diff
@@ -25,6 +25,12 @@ var cliExe = "cartridge"
 var goPackageName = "github.com/tarantool/cartridge-cli/cli"
 var packagePath = "./cli"
 
+var generateModePath = filepath.Join(packagePath, "create", "codegen", "generate_mode.go")
+var generatedFilesPath = "." + string(filepath.Separator) + filepath.Join(packagePath, "create", "codegen", "static")
```
dokshina, line 29: Please, use `filepath.Join(".", ...)` or just `fmt.Sprintf("./%s", ...)`.

Using `fileptah.Separator` is good, but look at the code above and don't care ;)

magefile.go
```diff
@@ -25,6 +25,12 @@ var cliExe = "cartridge"
...
+
+var generatedFSFile = "cartridgedata_vfsdata.go"
```
dokshina, line 31: Maybe it should be `cartridge_vfsdata `?

magefile.go
```diff
@@ -92,6 +98,11 @@ func init() {
 
 // Run go vet and flake8
 func Lint() error {
+	fmt.Println("Generating Go code...")
+	if err := generateGoCode(); err != nil {
```
dokshina, line 102: `mg.Deps(generateGoCode)`?

magefile.go
```diff
@@ -164,6 +180,31 @@ func Build() error {
 	return nil
 }
 
+// Generate Go code that statically implements filesystem
+// and map with modes for that filesystem.
+func generateGoCode() error {
```
dokshina, line 185: I think we can expose this function (to use for debug, for examlpe).

cli/create/codegen/generate_mode.go
```diff
@@ -0,0 +1,72 @@
...
+func getFileModes(root string) (map[string]int, error) {
+	fileModeMap := make(map[string]int)
+
+	err := filepath.Walk(root, func(filePath string, fileInfo os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if !fileInfo.IsDir() {
+			rel, _ := filepath.Rel(root, filePath)
```
dokshina, line 49: Generally, ignoring error leads to `"": filemode` value in a map, I think we can return it.

cli/create/templates/cartridge/test/helper/integration.lua
```diff
@@ -0,0 +1,29 @@
...
+    datadir = shared.datadir,
+    use_vshard = false,
+    replicasets = {
+        {
+            alias = 'api',
+            uuid = cartridge_helpers.uuid('a'),
+            roles = {'app.roles.custom'},
+            servers = {
+				{
+                    instance_uuid = cartridge_helpers.uuid('a', 1)
```
dokshina, line 19: Broken idents?

magefile.go
```diff
@@ -25,6 +25,12 @@ var cliExe = "cartridge"
 var goPackageName = "github.com/tarantool/cartridge-cli/cli"
 var packagePath = "./cli"
 
+var generateModePath = filepath.Join(packagePath, "create", "codegen", "generate_mode.go")
+var generatedFilesPath = fmt.Sprintf("./%s", filepath.Join(packagePath, "create", "codegen", "static"))
```
dokshina, line 29: `packagePath` already has `./` at the beginning, is it removed on `filepath.Join`?

mRrvz: yes

dokshina (dismissed):

dokshina (approved):

PR #430, opened 2020-12-23
Refactor reading from Tarantool conn timeouts

dokshina:
Timeouts is a good approach, but sometimes we have to wait for command
to be completed
For example, connect,enter,admin commands can be used to collect statistics from huge clusters.
Generally, there is no "right" timeout for `edit_topology` calls, so
it's called without timeout.
Some simple things (like fetching Cartridge version) or completion
operations are performed with a little timeout.

### DIFF
--- a/cli/common/socket.go
+++ b/cli/common/socket.go
@@ -19,19 +19,6 @@ const (
 	endOfLuaOutput  = ";"
 )
 
-var (
-	EvalTarantoolConnTimeout time.Duration
-)
-
-func init() {
-	// In fact, sometimes we need to disable this timeout on production.
-	// For example, if replica is joined to a big storage.
-	// In this case Cartridge waits for loading all data.
-	// But since `cartridge replicasets` command is developed for local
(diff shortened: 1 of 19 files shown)


cli/common/socket.go
```diff
@@ -77,11 +77,15 @@ func ConnectToTarantoolSocket(socketPath string) (net.Conn, error) {
...
+func ReadFromConn(conn net.Conn, endOfOutput string, readTimeout time.Duration) ([]byte, error) {
 	tmp := make([]byte, 1)
 	data := make([]byte, 0)
 
-	conn.SetReadDeadline(time.Now().Add(EvalTarantoolConnTimeout))
+	if readTimeout > 0 {
+		conn.SetReadDeadline(time.Now().Add(readTimeout))
+	} else {
+		conn.SetReadDeadline(time.Time{})
+	}
```
olegrok, line 88: Negative timeout should be an error IMO

olegrok (approved):

PR #438, opened 2020-12-28
Fix handling logs on interactive start

dokshina:
When instances are started in interactive mode, colorized log writer
is used to prettify log messages. After writing some big amount of data 
to log, this function become waiting forever.
This patch fixes `ColorizedWriter.Write` function to handle number
of bytes that was wrote correctly.

### DIFF
--- a/cli/running/writer.go
+++ b/cli/running/writer.go
@@ -38,6 +38,7 @@ func init() {
 
 type ColorizedWriter struct {
 	prefix string
+	out    io.Writer
 }
 
 func (w *ColorizedWriter) Write(p []byte) (int, error) {
@@ -54,7 +55,7 @@ func (w *ColorizedWriter) Write(p []byte) (int, error) {
 		}
 
 		// prefix
-		if _, err := os.Stdout.Write([]byte(w.prefix)); err != nil {
+		if _, err := w.out.Write([]byte(w.prefix)); err != nil {
 			return n, err
 		}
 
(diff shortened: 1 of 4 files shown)


cli/running/writer_test.go
```diff
@@ -0,0 +1,64 @@
...
+func TestWrite(t *testing.T) {
+	t.Parallel()
+
+	assert := assert.New(t)
+
+	out := bytes.NewBuffer(nil)
+	var logs []string
+	var logBytes []byte
+	var n int
+	var err error
```
mRrvz, line 34: This is looks like a Pascal-style: variables are declared at the beginning, not in the place where they will be used

cli/running/writer.go
```diff
@@ -38,23 +38,26 @@ func init() {
...
 			break
 		}
-		if err != nil {
-			return n, err
-		}
 
 		// prefix
-		if _, err := os.Stdout.Write([]byte(w.prefix)); err != nil {
+		if nPrefix, err := w.out.Write([]byte(w.prefix)); err != nil {
+			n += nPrefix
```
mRrvz, line 60: What is ```n```?

cli/running/writer_test.go
dokshina: See below, this variables are used on each test.

cli/running/writer.go
dokshina: `Write` function of `io.Writer` interface returns `n, err`, wrere `n` is number of bytes that was written.

olegrok (approved):

PR #439, opened 2020-12-29
Update template app dependencies

dokshina:


cli/create/templates/cartridge/{{ .Name }}-scm-1.rockspec
```diff
@@ -8,7 +8,7 @@ dependencies = {
     'tarantool',
     'lua >= 5.1',
     'checks == 3.0.1-1',
```
olegrok, line 10: * cartridge uses checks "3.1.0"
* metrics could be updated to "0.6.0"

olegrok (approved):
🔥

PR #444, opened 2021-01-13
Refactor connect to Tarantool instance

dokshina:


cli/connector/eval_plain_text.go
```diff
@@ -0,0 +1,389 @@
...
+	if args == nil {
+		args = []interface{}{}
+	}
+
+	argsEncoded, err := msgpack.Marshal(args)
+	if err != nil {
+		return fmt.Errorf("Failed to encode args: %s", err)
+	}
+
+	evalFunc, err := templates.GetTemplatedStr(&evalFuncTmpl, map[string]string{
```
olegrok, line 109: any "eval" could be considered as call('package.loaded.console.eval', 'string'). Do we need evals?

cli/connector/eval_plain_text.go
```diff
@@ -0,0 +1,389 @@
...
+}
+
+func writeToPlainTextConn(conn net.Conn, data string) error {
+	writer := bufio.NewWriter(conn)
+	if _, err := writer.WriteString(data); err != nil {
+		return fmt.Errorf("Failed to send to socket: %s", err)
+	}
+
+	if err := writer.Flush(); err != nil {
+		return fmt.Errorf("Failed to flush: %s", err)
```
olegrok, line 138: It would be great some details. "Failed to flush" says nothing for user.

cli/connector/eval_plain_text.go
```diff
@@ -0,0 +1,389 @@
...
+	{{ .FunctionBody }}
+end
+
+local ok, ret = pcall(function()
+	local args = require('msgpack').decode(string.fromhex('{{ .ArgsEncoded }}'))
+	local ret = { __func(unpack(args)) }
+	return ret
+end)
+
+if ret == nil then ret = box.NULL end
```
olegrok, line 361: Is this line really needed?

cli/connector/eval_plain_text.go
```diff
@@ -0,0 +1,389 @@
...
+
+if not ok then
+	return { success = false, err = ret}
+end
+
+return { success = true, data_enc = require('yaml').encode(ret) }
```
olegrok, line 367: Why can't you use something like:

```
return {message = require('yaml').encode(ret)} -- for success
return {err = err} -- for errors
```

cli/connector/eval_plain_text.go
olegrok:
```
tarantool> require('yaml').encode(nil)
---
- '--- null

  ...

  '
...

tarantool> require('yaml').encode(box.NULL)
---
- '--- null

  ...

  '
...
```

cli/admin/call.go
```diff
@@ -36,58 +33,91 @@ type FuncCallArg struct {
...
-		PushCallback: printMessage,
-	})
+	callResData, err := conn.Exec(callReq)
+
 	if err != nil {
 		return fmt.Errorf("Failed to call %q: %s", funcName, err)
 	}
 
-	printCallRes(callResRaw)
+	if len(callResData) < 1 || len(callResData) > 2 {
```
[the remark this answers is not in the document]
dokshina: It can be 1 or 2, I will add a comment.

cli/admin/call.go
```diff
@@ -36,58 +33,91 @@ type FuncCallArg struct {
...
+		return fmt.Errorf("Bad data len: %d", len(callResData))
+	}
+
+	if len(callResData) > 1 {
```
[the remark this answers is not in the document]
dokshina: Fixed to be more clear (see comment above).

cli/admin/call.go
```diff
@@ -189,25 +192,18 @@ func printCallRes(callResRaw interface{}) {
...
+	msg, ok := pushedData.(string)
+	if !ok {
+		log.Warnf("Intermediate message should be a string, got %v", pushedData)
 
-	var msg string
-	if err := yaml.UnmarshalStrict([]byte(msgEncoded), &msg); err != nil {
+		msgEncoded, err := yaml.Marshal(pushedData)
+		if err != nil {
+			log.Errorf("Failed to encode received intermediate message: %s", err)
+		}
```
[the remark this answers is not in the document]
dokshina: Pushed data is received unmarshalled, this is `interface{}` data (In case of plain text connection it is received in yaml format and then unmarshalled, for binary protocol it's received as a msgpack). But here we marshal this data to YAML to display it for user. Of course, `{a = 'b'}` table is handled properly.

cli/connector/eval_plain_text.go
dokshina: I just think that `res.Success` check looks good and it's pretty clear, but it also can be `res.Err != nil` check.

cli/connector/eval_plain_text.go
```diff
@@ -0,0 +1,409 @@
...
+	// tarantool> error('XXX')
+	// ---
+	// - error: XXX
+	// ...
+	//
+	// Lua output:
+	// tarantool> return 'XXX'
+	// "XXX";
+	// tarantool> error('XXX')
+	// "XXX";
```
olegrok, line 312:
```
Tarantool Enterprise 2.5.2-0-g05730d326-r366-macos
type 'help' for interactive help
tarantool> \set output lua
true;
tarantool> error('XXX')
{error = "XXX"};
```

cli/admin/common.go
```diff
@@ -183,6 +182,17 @@ func normalizeFlagName(name string) string {
...
+func getAdminFuncEvalTypedBody(adminFuncName string) (string, error) {
+	funcBody, err := templates.GetTemplatedStr(&evalFuncGetResBodyTmpl, map[string]string{
+		"FuncName": adminFuncName,
+	})
+	if err != nil {
+		return "", project.InternalError("Failed to instantiate func call body template: %s", err)
+	}
+
+	return funcBody, nil
+}
```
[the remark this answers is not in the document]
dokshina: I think that there is no reason to do it.

cli/connector/eval_plain_text.go
dokshina: We use plain text protocol, there wouldn't be any beautiful `call`.

cli/connector/binary.go
```diff
@@ -0,0 +1,89 @@
...
+
+	if execOpts.ReadTimeout != 0 {
+		ctx, cancel = context.WithTimeout(ctx, execOpts.ReadTimeout)
+		defer cancel()
+	}
+
+	return execTarantoolReqBinary(conn.binary, req, ctx, execOpts.ResData)
+}
+
+func execTarantoolReqBinary(binaryConn *tarantool.Connection, req *tarantool.Request, ctx context.Context, resData interface{}) ([]interface{}, error) {
```
olegrok, line 66: not sure but AFAIK usually in go context is passed as the first argument.

cli/connector/binary.go
```diff
@@ -0,0 +1,89 @@
...
+	var err error
+	var resp *tarantool.Response
+
+	if resData != nil {
+		// ExecTypedContext now fails, I will use connector fork with a fix.
+		// This change will be performed in a separated patch
+		//
+		// err = binaryConn.ExecTypedContext(ctx, req, resData)
+		err = binaryConn.ExecTyped(req, resData)
```
olegrok, line 75: Is it ok that resp will be nil after this line?

cli/connector/connector.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func (conn *Conn) Close() error {
+	switch conn.protocol {
+	case PlainTextProtocol:
+		return conn.plainText.Close()
+	case BinaryProtocol:
+		return conn.binary.Close()
+	default:
+		return fmt.Errorf("Unsupported protocol: %s", conn.protocol)
```
olegrok, line 93: I suppose this line should be unreachable. Seems you should place "panic" here.

cli/connector/connector.go
```diff
@@ -0,0 +1,122 @@
...
+		return BinaryProtocol, nil
+	default:
+		return "", fmt.Errorf("Unknown protocol: %s", greeting)
+	}
+}
+
+func readGreeting(conn net.Conn) (string, error) {
+	conn.SetReadDeadline(time.Now().Add(SimpleOperationTimeout))
+
+	greeting := make([]byte, 1024)
```
olegrok, line 116: Seems 1024 could be moved into constant (https://www.tarantool.io/en/doc/1.10/dev_guide/internals_index/#greeting-packet)

cli/connector/eval_plain_text.go
```diff
@@ -0,0 +1,413 @@
...
+	return pushedData, nil
+}
+
+func processEvalTarantoolRes(resBytes []byte, result interface{}) ([]interface{}, error) {
+	var err error
+	var evalResultEncHex string
+
+	var getResultEncHexFunc func([]byte) (string, error)
+	// Result data is returned as a table
+	// `{ data_enc = msgpack.encode(ret):hex() }`.
```
olegrok, line 292: Could you eliminate "hex" here?

Anyway it's an array of bytes.

cli/connector/eval_plain_text.go
```diff
@@ -0,0 +1,413 @@
...
+const (
+	evalFuncTmpl = `
+local function func(...)
+	{{ .FunctionBody }}
+end
+local args = require('msgpack').decode(string.fromhex('{{ .ArgsEncoded }}'))
+
+local ret = {
+	load(
+		'local func = ... return func(unpack(select(2, ...)))',
```
olegrok, line 407: Why do you need unpack here?

First of all, you could lose some values:
```
tarantool> unpack({1, nil, 2, nil, 2})
---
- 1
...
```

Secondly, you lose trailing nils.

cli/repair/clusterwide_config.go
```diff
@@ -264,6 +264,7 @@ func (topologyConf *TopologyConfType) setReplicasetsConf() error {
 				return fmt.Errorf("Replicaset %s config doesn't contain %q key", replicasetUUID, keyReplicasetRoles)
 			}
 
+			// XXX: empty map is a list!!
```
olegrok, line 267: Don't forget to file issues for all "XXX"

cli/repair/common.go
```diff
@@ -6,24 +6,21 @@ import (
...
 	"github.com/pmezard/go-difflib/difflib"
 	"github.com/tarantool/cartridge-cli/cli/common"
+	"github.com/tarantool/cartridge-cli/cli/connector"
 	"github.com/tarantool/cartridge-cli/cli/context"
 	"github.com/tarantool/cartridge-cli/cli/project"
 )
 
 var (
-	minCartridgeVersionForReload *goVersion.Version
+	minCartridgeMajorVersionForReload    = 2
```
olegrok, line 20: Reload? Is it connected with roles hot reload? (https://github.com/tarantool/cartridge/pull/1105)

cli/admin/call.go
```diff
@@ -36,58 +33,94 @@ type FuncCallArg struct {
...
+
+		switch funcCallArg.Type {
+		case argTypeString:
+			value = funcCallArg.StringValue
+		case argTypeNumber:
+			value = funcCallArg.NumberValue
+		case argTypeBool:
+			value = funcCallArg.BoolValue
+		default:
+			return nil, project.InternalError("Unknown argument type: %s ", funcCallArg.Type)
```
olegrok, line 97: trailing space
`: %s "` -> `: %s"`

cli/connector/binary.go
dokshina: See below - in this case nil data is returned.

cli/connector/eval_plain_text.go
dokshina: But...

```
tarantool> args = {1, 2, nil, 4}
---
...

tarantool> func = function(...) print(...) end
---
...

tarantool> load(
'local func = ... return func(unpack(select(2, ...)))',
'@eval'
)(func, args)
1	2	nil	4
---
...
```

cli/repair/clusterwide_config.go
dokshina: https://github.com/tarantool/cartridge-cli/issues/449

cli/repair/common.go
dokshina: Cluster-wide config reload. This is about `repair --reload`.

cli/connector/eval_plain_text.go
dokshina: For the end user it would be smth like `Failed to send eval function to socket: Failed to flush`.

cli/connector/eval_plain_text.go
dokshina: There is a problem with Lua output. The approach with hex encoding is the simplest now, I believe it become better sometimes.

```
tarantool> res = require('console').eval('return true')
;
tarantool> res
"true;";
tarantool> msgpack.encode(res)
"�true;";
```

cli/connector/eval_plain_text.go
dokshina: Since we use msgpack to pass arguments to instance, all nils are `box.NULL`, so there is no problems with loosing nils.

cli/connector/binary.go
```diff
@@ -0,0 +1,85 @@
+package connector
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/dokshina/tarantool"
```
olegrok, line 7: This is not the same as what is written in commit message.

"For binary protocol: https://github.com/FZambia/tarantool". But here you use your fork without any comments why do you do that. Finally this switch could be moved into separate commit (https://github.com/tarantool/cartridge-cli/compare/56bffec679eadda4c76424bb24ee9202f45915a3..ea5ef182ccd1d2958c57e069de3541f7089e0395)

dokshina: See https://github.com/tarantool/cartridge-cli/issues/452

I suggest merging this patch as is (without using github.com/dokshina/tarantool). After merging my patch into connector upstream I will fix calling `ExecTypedContext` in a a separated patch.

olegrok (approved):

Issue #460, opened 2021-01-27
[2pt] Persist --stateboard flag in .cartridge.yml


dokshina:

rosik:
I suggest also adding it to the template and making it `true` by default.

PR #463, opened 2021-01-28
Disable instance name prefix in logs when running interactively

### LINKED ISSUE #373: [1pt] Possibility to to disable instance name as log prefix when using cartridge start
When I start the cluster via `cartridge start`, it appends string like “myapp.router |” to log entries, it would be good to have a way to disable that.

P.S. I am aware that logs can be redirected to files by running `cartridge start` with `-d` option or by specifying `log` configuration variable for instances, but this issue is about disabling the prefix, while keeping output at stdout.

mRrvz:
Added `--no-log-prefix` option for `cartridge start` command to disable instance name prefix in logs when running interactively. Closes #373

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added 
+
+- `--no-log-prefix` option for `cartridge start` command to disable instance name prefix in logs
+
 ## [2.6.0] - 2021-01-27
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added 
 
-- `--no-log-prefix` option for `cartridge start` command to disable instance name prefix in logs
+- `--no-log-prefix` option for `cartridge start` command to disable instance name prefix in logs when running interactively.
 
 ## [2.6.0] - 2021-01-27
 
--- a/cli/commands/common.go
+++ b/cli/commands/common.go
@@ -52,6 +52,10 @@ func addStateboardRunningFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&ctx.Running.StateboardOnly, "stateboard-only", false, stateboardOnlyUsage)
 }
 
+func addDisablePrefixFlag(cmd *cobra.Command) {
+	cmd.Flags().BoolVar(&ctx.Running.DisableLogPrefix, "disable-prefix", false, disablePrefixUsage)
+}
+
 func addCommonRunningPathsFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&ctx.Running.RunDir, "run-dir", "", runDirUsage)
 	cmd.Flags().StringVar(&ctx.Running.ConfPath, "cfg", "", cfgUsage)
--- a/cli/commands/common.go
+++ b/cli/commands/common.go
@@ -53,7 +53,7 @@ func addStateboardRunningFlags(cmd *cobra.Command) {
 }
 
 func addDisablePrefixFlag(cmd *cobra.Command) {
-	cmd.Flags().BoolVar(&ctx.Running.DisableLogPrefix, "disable-prefix", false, disablePrefixUsage)
+	cmd.Flags().BoolVar(&ctx.Running.DisableLogPrefix, "no-log-prefix", false, disablePrefixUsage)
 }
 
 func addCommonRunningPathsFlags(cmd *cobra.Command) {
--- a/cli/commands/common.go
+++ b/cli/commands/common.go
@@ -52,10 +52,6 @@ func addStateboardRunningFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&ctx.Running.StateboardOnly, "stateboard-only", false, stateboardOnlyUsage)
 }
 
-func addDisablePrefixFlag(cmd *cobra.Command) {
-	cmd.Flags().BoolVar(&ctx.Running.DisableLogPrefix, "no-log-prefix", false, disablePrefixUsage)
-}
-
 func addCommonRunningPathsFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&ctx.Running.RunDir, "run-dir", "", runDirUsage)
 	cmd.Flags().StringVar(&ctx.Running.ConfPath, "cfg", "", cfgUsage)
--- a/cli/commands/start.go
+++ b/cli/commands/start.go
@@ -44,7 +44,7 @@ func init() {
 	addStateboardRunningFlags(startCmd)
 
 	// Disable instance(s) prefix flag
-	addDisablePrefixFlag(startCmd)
+	startCmd.Flags().BoolVar(&ctx.Running.DisableLogPrefix, "no-log-prefix", false, disableLogPrefixUsage)
 
 	// common running paths
 	addCommonRunningPathsFlags(startCmd)
@@ -56,6 +56,12 @@ func init() {
 
 func runStartCmd(cmd *cobra.Command, args []string) error {
 	var err error
+
+	if timeoutStr == "" {
+		ctx.Running.DefaultTimeout = true
+	} else {
+		ctx.Running.DefaultTimeout = false
+	}
 
 	if err := setDefaultValue(cmd.Flags(), "timeout", defaultStartTimeout.String()); err != nil {
 		return project.InternalError("Failed to set default timeout value: %s", err)
--- a/cli/commands/start.go
+++ b/cli/commands/start.go
@@ -43,13 +43,15 @@ func init() {
 	// stateboard flags
 	addStateboardRunningFlags(startCmd)
 
+	// Disable instance(s) prefix flag
+	addDisablePrefixFlag(startCmd)
+
 	// common running paths
 	addCommonRunningPathsFlags(startCmd)
 	// start-specific paths
 	startCmd.Flags().StringVar(&ctx.Running.DataDir, "data-dir", "", dataDirUsage)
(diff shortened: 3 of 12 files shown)


cli/commands/common.go
```diff
@@ -52,6 +52,10 @@ func addStateboardRunningFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&ctx.Running.StateboardOnly, "stateboard-only", false, stateboardOnlyUsage)
 }
 
+func addDisablePrefixFlag(cmd *cobra.Command) {
+	cmd.Flags().BoolVar(&ctx.Running.DisableLogPrefix, "disable-prefix", false, disablePrefixUsage)
```
olegrok, line 56: Maybe "--no-log-prefix" (as it was done for docker-compose - https://github.com/docker/compose/pull/7435)

cli/running/writer.go
```diff
@@ -101,7 +101,12 @@ func newColorizedWriter(prefix string) *ColorizedWriter {
 		out: os.Stdout,
 	}
 
-	prefixColor := nexPrefixColor()
-	writer.prefix = prefixColor.Sprintf("%s | ", prefix)
+	if prefix != "" {
```
olegrok, line 104: Maybe you need just "newDummy"/"newNoPrefix" writer to create it at https://github.com/tarantool/cartridge-cli/pull/463/files#diff-44e2d99ca4c277c2ae3e6bbd3ea8515486d311ab90fb3578ffd8e194bffa77fdR190

olegrok (commented):
Hi! Thanks for your patch. Consider several comments below.

Also please describe your changes in commit message - that you've done and why. Also it would be great to have some tests and changelog note.

Also please don't put any prefixes to your commit message title. The feature is not huge it will be merged as single commit: "allow to disable instance name prefix in logs" seems to be good enough.

dokshina:
Maybe we can persist this flag? See #460

olegrok:
> Maybe we can persist this flag? See #460

Sorry, but do you have sufficient grounds for this?

It seems reasonable for stateboard but we shouldn't persist each option in my opinion

dokshina:
> Sorry, but do you have sufficient grounds for this?

Let's discuss it in #460

cli/commands/common.go
```diff
@@ -52,6 +52,10 @@ func addStateboardRunningFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&ctx.Running.StateboardOnly, "stateboard-only", false, stateboardOnlyUsage)
 }
 
+func addDisablePrefixFlag(cmd *cobra.Command) {
+	cmd.Flags().BoolVar(&ctx.Running.DisableLogPrefix, "no-log-prefix", false, disablePrefixUsage)
```
dokshina, line 56: `disablePrefixUsage` -> `disableLogPrefixUsage `

cli/commands/start.go
```diff
@@ -43,13 +43,15 @@ func init() {
 	// stateboard flags
 	addStateboardRunningFlags(startCmd)
 
+	// Disable instance(s) prefix flag
+	addDisablePrefixFlag(startCmd)
```
dokshina, line 47: I think we can add this flag without wrapper function - I've used them for adding multiple flags that are the same for several commend.

cli/running/writer_test.go
```diff
@@ -13,6 +13,15 @@ func getLogsBytes(logs []string) []byte {
 	return []byte(strings.Join(logs, "\n"))
 }
 
+func getLogsNoPrefix(logs []string) string {
+	resLines := make([]string, len(logs))
+	for i := range logs {
+		resLines[i] = fmt.Sprintf("%s", logs[i])
+	}
+
+	return strings.Join(resLines, "\n")
```
dokshina, line 22: Why not just `return strings.Join(logs, "\n")`?

cli/running/writer_test.go
```diff
@@ -62,3 +71,37 @@ func TestWrite(t *testing.T) {
...
+	assert := assert.New(t)
+	out := bytes.NewBuffer(nil)
+
+	writer := newDummyWriter()
+	writer.out = out
+
+	out.Reset()
+	logs := []string{
+		"Line_01",
+		"Multiline02!",
```
dokshina, line 87: What do you mean using `Multiline02` name?

dokshina: BTW, now I think that test above does smth strange:

```go
/ multiline
	out.Reset()
        // What???
	logs = []string{
		"Some long",
		"multiline",
		"log",
	}

       // maybe?
      logs = []string{
		"Some long\nmultiline\log",
	}
```

cli/running/writer_test.go
```diff
@@ -62,3 +71,37 @@ func TestWrite(t *testing.T) {
...
+	out.Reset()
+	logs = []string{
+		"One line log without prefix.",
+	}
+
+	logBytes = getLogsBytes(logs)
+	n, err = writer.Write(logBytes)
+	assert.Nil(err)
+	assert.Equal(len(logBytes), n)
+	assert.Equal(getLogsNoPrefix(logs), out.String())
```
dokshina, line 106: `assert.Equal(strings.Join(logs, "\n"), out.String())`?

test/utils.py
```diff
@@ -726,6 +728,7 @@ def wait_instances(cli, project, instance_names=[], run_dir=None, stateboard=Fal
 def check_instances_running(cli, project, instance_names=[],
                             stateboard=False, stateboard_only=False,
                             daemonized=False,
+                            no_log_prefix=False,
```
dokshina, line 731: Seems that this option does nothing.

test/integration/running/test_running.py
```diff
@@ -702,3 +702,14 @@ def test_stop_signals(start_stop_cli, project_ignore_sigterm):
...
+
+def test_start_no_log_prefix(start_stop_cli, project_without_dependencies):
+    project = project_without_dependencies
+    cli = start_stop_cli
+
+    INSTANCE1 = 'instance-1'
+    INSTANCE2 = 'instance-2'
+
+    cli.start(project, [INSTANCE1, INSTANCE2], no_log_prefix=True)
+    check_instances_running(cli, project, [INSTANCE1, INSTANCE2], no_log_prefix=True)
```
dokshina, line 715: It looks like this test just checks that new option is correctly parsed by CLI. I'm not sure we need this test, but maybe we are. Please, add a comment if you decide that this test is necessary.

cli/context/context.go
```diff
@@ -66,8 +66,9 @@ type RunningCtx struct {
 	Daemonize    bool
 	StartTimeout time.Duration
 
-	LogFollow bool
-	LogLines  int
+	LogFollow        bool
+	LogLines         int
+	DisableLogPrefix bool
```
dokshina, line 71: What if user specified `start -d --disable-log-prefix`? Should we warn that `--disable-log-prefix` is ignored? What happens now for `start --timeout` (timeout is used for daemonized mode)?

cli/commands/start.go
```diff
@@ -43,18 +43,26 @@ func init() {
 	// stateboard flags
 	addStateboardRunningFlags(startCmd)
 
+	// Disable instance(s) prefix flag
```
dokshina, line 46: `// Disable instance name prefix in logs`

cli/commands/start.go
```diff
@@ -43,18 +43,26 @@ func init() {
...
 	startCmd.Flags().StringVar(&ctx.Running.DataDir, "data-dir", "", dataDirUsage)
 	startCmd.Flags().StringVar(&ctx.Running.LogDir, "log-dir", "", logDirUsage)
 	startCmd.Flags().StringVar(&ctx.Running.Entrypoint, "script", "", scriptUsage)
-
 }
 
 func runStartCmd(cmd *cobra.Command, args []string) error {
 	var err error
 
+	if timeoutStr == "" {
```
dokshina, line 60:
```go
if timeoutStr != "" and !ctx.Running.Daemonized {
    log.Warnf("--timeout flag is ignored due to starting instances in foreground")
}

And here you can add second waring (`ctx.Running.DisableLogPrefix && ctx.Running.Daemonize`)

It's not so much important to add new flag to context, it confuses a lot.

``

File: cli/commands/usage.go [hunk 13]
```diff
@@ -87,6 +87,8 @@ defaults to ./instances.yml ("cfg" in .cartridge.yml)`
 	logFollowUsage = `Output appended data as the log grows`
 
 	stopForceUsage = `Force instance(s) stop (sends SIGKILL)`
+
+	disableLogPrefixUsage = `Disable instance(s) prefix in logs`
```
dokshina, line 91: `instance(s) name prefix`?

Or maybe simple `Disable prefix in logs when run interactively`?

test/utils.py
```diff
@@ -116,9 +116,9 @@ def __init__(self, cartridge_cmd):
         self._instance_pids = set()
         self._subprocess = None
 
-    def start(self, project, instances=[], daemonized=False, stateboard=False, stateboard_only=False,
-              cfg=None, script=None, run_dir=None, data_dir=None, log_dir=None, timeout=None,
-              capture_output=False, exp_rc=0):
+    def start(self, project, instances=[], daemonized=False, stateboard=False,
+              stateboard_only=False, cfg=None, script=None, run_dir=None, data_dir=None,
+              log_dir=None, timeout=None, capture_output=False, exp_rc=0):
```
dokshina, line 121: Please, discard this change.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added 
+
+- `--no-log-prefix` option for `cartridge start` command to disable instance name prefix in logs
```
dokshina, line 12: .. when running interactively?

dokshina (approved):
Good!

mRrvz:
Added ``--no-log-prefix`` option for ``cartridge start`` command to disable instance name prefix in logs when running interactively. Closes #373

PR #442, opened 2020-12-29
ci: update to use go1.16

chenrui333:
relates to https://github.com/docker/go-docker/issues/19

PR #477, opened 2021-02-22
Add go.sum

roopakv:
Add `go.sum`. This is required to make sure signatures are right for dependencies. Further not adding this is not only a security concern but also makes it not possible to install cartridge-cli via brew. 

We are trying to move to go1.16 and this is one of the blocking bugs.

Please consider releasing a new version after merging this PR.


roopakv:
ok i did not notice #442. way better fix. closing this out

PR #482, opened 2021-02-26
Refactor Cartridge application template

### LINKED ISSUE #433: create: provide templated application with `README.md` file
An application created with `cartridge create` is lacking a README file

dokshina:
Add a README that describes created application
Rename app.path -> package_compat
Move test helper from test/helper/helper.lua to test/helper.lua

Closes #433


cli/create/templates/cartridge/README.md
```diff
@@ -0,0 +1,78 @@
...
+```bash
+cartridge build
+cartridge start -d
+cartridge replicasets setup --bootstrap-vshard
+```
+
+Now you can visit http://localhost:8081 and see your application's Admin Web UI.
+
+**Note**, that application stateboard is always started by default.
+See [`.cartridge.yml`](./.cartridge.yml) file to change this behavior.
```
mRrvz, line 18: Maybe add information about

```bash
--stateboard=false
```

flag?

cli/create/templates/cartridge/README.md
```diff
@@ -0,0 +1,78 @@
...
+It is used by `cartridge replicasets setup`.
+
+## Tests
+
+Simple tests are placed in [`test`](./test) directory.
+To run tests, say
+
+```bash
+cartridge stop  # to prevent "address already in use" error
+./deps.sh
```
mRrvz, line 62: There maybe a note, that ./deps.sh should be called once (not every test run)?

cli/create/templates/cartridge/README.md
```diff
@@ -0,0 +1,78 @@
...
+.rocks/bin/luacheck .
```
mRrvz, line 63: But it's not a tests, it's a lint...

cli/create/templates/cartridge/README.md
dokshina: It seems logical, by passing `--stateboard=false` on `start` requires passing it on `stop` (otherwise, you will get error).
I think that this little note is enough for the first start of application.

cli/create/templates/cartridge/README.md
dokshina: I've changed this section structure - installing dependencies and linter are described separately.

mRrvz (approved):

PR #486, opened 2021-03-02
Remove tarantool process name check

### LINKED ISSUE #476: [1pt] Eliminate test for tarantool process name
In `cli/running/process.go` we have a test
```
	if name != "tarantool" {
		process.Status = procStatusError
		process.Error = fmt.Errorf("Process %d does not seem to be tarantool", process.pid)
		return
	}
```

which prevents us to run modified\proxified tarantool. For example under the `valgrind`. Strictly speaking this test doesn't guarantee anything 'cause the executable may change own `comm` during runtime.

I think it would be better either remove this hunk completely or modify it to warning instead without breaking execution.

mRrvz:
Executable with any name (not only ``tarantool``) can run processes. Closes #476

### DIFF
--- a/test/integration/running/test_running.py
+++ b/test/integration/running/test_running.py
@@ -5,7 +5,7 @@ import re
 
 from utils import check_instances_running, check_instances_stopped
 from utils import STATUS_NOT_STARTED, STATUS_RUNNING, STATUS_STOPPED
-from utils import write_conf
+from utils import write_conf, find_file_in_path
 
 from project import patch_init_to_send_statuses
 from project import patch_init_to_send_ready_after_timeout
@@ -410,7 +410,7 @@ def test_start_stop_custom_tararntool(start_stop_cli, project_without_dependenci
     # fake tarantool script is always higher than the original tarantool.
     # As a result, our process is not called a 'tarantool'.
 
-    real_tarantool_abs_path = '/usr/local/bin/tarantool'
+    real_tarantool_abs_path = find_file_in_path(os.environ['PATH'], 'tarantool')
     tarantool_substitution_script = f"#!/bin/sh\n{real_tarantool_abs_path} $1"
     with open('tarantool', 'w') as f:
         f.write(tarantool_substitution_script)
--- a/test/integration/running/test_running.py
+++ b/test/integration/running/test_running.py
@@ -398,6 +398,40 @@ def test_start_stop_status_cfg(start_stop_cli, project_without_dependencies):
     assert status.get(ID2) == STATUS_STOPPED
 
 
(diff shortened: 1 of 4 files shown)


test/integration/running/test_running.py
```diff
@@ -398,6 +399,44 @@ def test_start_stop_status_cfg(start_stop_cli, project_without_dependencies):
...
+
+    old_path = os.environ['PATH']
+    os.environ['PATH'] = os.getcwd() + ':' + old_path
+
+    logs = cli.start(project, [INSTANCE1], stateboard=True, daemonized=True, capture_output=True, run_dir=RUN_DIR)
+
+    os.environ['PATH'] = old_path
+    os.remove('tarantool')
+
+    # Linux, unlike Darwin, which will have a process name 'sh',
```
dokshina, line 429: Maybe smth like `On Linux process have name 'tarantool' anyway, but for Darwin it's 'sh', so we can check that warning was displayed`?

test/integration/running/test_running.py
```diff
@@ -398,6 +399,44 @@ def test_start_stop_status_cfg(start_stop_cli, project_without_dependencies):
     assert status.get(ID2) == STATUS_STOPPED
 
 
+def test_start_stop_custom_tarantool(start_stop_cli, project_without_dependencies):
```
dokshina, line 402: custom_tarantool -> custom_executable

test/integration/running/test_running.py
```diff
@@ -398,6 +399,44 @@ def test_start_stop_status_cfg(start_stop_cli, project_without_dependencies):
...
+        f.write(tarantool_substitution_script)
+
+    os.chmod('tarantool', 0o755)
+
+    old_path = os.environ['PATH']
+    os.environ['PATH'] = os.getcwd() + ':' + old_path
+
+    logs = cli.start(project, [INSTANCE1], stateboard=True, daemonized=True, capture_output=True, run_dir=RUN_DIR)
+
+    os.environ['PATH'] = old_path
```
dokshina, line 426: What if `cli.start` fail? `os.environ['PATH']` will contain path to patched `tarantool` script.

You can pass `env` argument to `suprocess.run` on `start` command.

test/utils.py
```diff
@@ -1229,3 +1229,10 @@ def get_admin_connection_params(connection_type, project):
         ]
 
     assert False, "Unknown connection type: %s" % connection_type
+
+
+def find_file_in_path(path_variable, name):
```
dokshina, line 1234: This name says me "I can find a file in a specfiied path", but in fact it uses `PATH` variable.

dokshina: BTW, look at `shutil.which`. Is it what you need?

dokshina (approved):
Good!

PR #489, opened 2021-03-03
Move `package_compat` back in init.lua from separate file

### LINKED ISSUE #487: Move `package compat` module back from separate file
In cartridge template.

mRrvz:
* CI: removed xfail from e2e tests
* Cartridge template: package compat moved back to init.lua from separate file

Closes #487 and #497


dokshina (commented):
I think that we need to add some test that fails for realization with `package_compat` module. It's very-very important change that can break production and shouldn't be performed.

test/integration/pack/test_pack_docker.py
```diff
@@ -203,3 +207,45 @@ def test_image_tag_without_git(cartridge_cmd, project_without_dependencies, tmpd
...
+        image_name,
+        environment=environment,
+        ports={http_port: http_port},
+        name=f"{project.name}-{instance_name}",
+        detach=True,
+    )
+
+    request.addfinalizer(lambda: container.remove(force=True))
+    assert container.status == 'created'
+    wait_for_container_get_error(container, "'package_compat' not found", time.time())
```
dokshina, line 251: This error means that there is no package_compat module and it's true. But require shouldn't fail for some existing module when application is starting from other directory and setsearchroot (or hack) isn't called. So your test should fail on current master, but Ok on this branch.

dokshina (approved):

PR #490, opened 2021-03-03
Fix consumes all disk space in `pack docker` command

### LINKED ISSUE #474: [1pt] `pack docker` consumes all disk space
See https://github.com/docker/for-mac/issues/2038#issuecomment-328059910.

Seems it can be fixed by adding `-l` flag to `adduser` command in build Dockerfile.

mRrvz:
Fixed bug when ``cartridge pack docker`` command consumes all disk space. Closes #474

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 
 - Connector crashing on using `cartridge admin` with binary port
+- ``cartridge pack docker`` consumes all disk space
 
 ### Added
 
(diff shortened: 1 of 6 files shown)


dokshina (commented):
I'm not sure that it really works now, but anyway, we shouldn't place such code in this dockerfiles.

1. It can be accidentally removed by user - we don't want it.
2. User can use other dockerfile as a base - we loose it again

Take a look at https://github.com/tarantool/cartridge-cli/blob/master/cli/project/dockerfiles.go#L197.

Moreover, I think that it can be possible to check that user was added with `-l` flag. It affects `/var/log/lastlog` file, can we check it's content. Not sure it works, but we should try.

dokshina (approved):

PR #499, opened 2021-03-11
Update Go to 1.16

dokshina:
Add go.sum to git
Clearify go.mod

See #442


olegrok (approved):

dokshina:
Thanks @chenrui333,  @roopakv and @Bo98!

PR #502, opened 2021-03-11
Bump Homebrew formula using GutHub Actions

dokshina:
Updating Homebrew formula for Cartridge CLI was performed manually
on each release. Of course, sometimes we forgot to do this. But future 
is here - now we use GitHub Action dawidd6/action-homebrew-bump-formula
[1] to create formula update PR for new tags.

[1]: https://github.com/marketplace/actions/homebrew-bump-formula


olegrok (approved):

dawidd6:
You should run this Action only on tag pushes.

dokshina:
Yes, you are right. I'll fix it.

dokshina:
Fixed in #504

PR #515, opened 2021-03-23
Support multiple rockspec in project folder

### LINKED ISSUE #385: [4pt] support more than one rockspec in project folder
For some dev cases, it would be nice to have an ability to pack app bundle tgz when 2 or more different rockspecs present in project folder by pointing the desired rockspec. Now if even use --name and --version it reports an error:
_Found multiple rockspecs in app_folder_

AnaNek:
Closes #385


cli/build/local.go
```diff
@@ -32,8 +32,8 @@ func buildProjectLocally(ctx *context.Ctx) error {
 	}
 
 	// tarantoolctl rocks make
-	log.Infof("Running `tarantoolctl rocks make`")
-	rocksMakeCmd := exec.Command("tarantoolctl", "rocks", "make")
+	log.Infof("Running `tarantoolctl rocks make %s`", ctx.Build.Spec)
+	rocksMakeCmd := exec.Command("tarantoolctl", "rocks", "make", ctx.Build.Spec)
```
dokshina, line 36: I think that we shouldn't pass found rockspec to `tarantoolctl rocks make` if user didn't specify `--spec` flag.
We can do it simpler:

* `cartridge build` leads to `tarantoolctl rocks make`
* `cartridge build --spec <path>` leads to `tarantoolctl rocks make <path>`

test/integration/build/test_build.py
```diff
@@ -81,6 +81,63 @@ def test_app_without_rockspec(cartridge_cmd, project_without_dependencies):
...
+    assert rc == 0
+    assert all([log in output for log in build_logs])
+    assert prebuild_output not in output
+
+    # check that all expected rocks was installed
+    files = recursive_listdir(project.path)
+    assert '.rocks' in files
+    assert all([rock in files for rock in project.rocks_content])
+
+
```
dokshina, line 140: Think about this test cases:

* User has two rockspecs in project root: What if he didn't pass `--spec` flag? what if he passed one of existing specs? what if he passed bad path?
* User hasn't rockspecs in project root:  What if he didn't pass `--spec` flag? What if he passed path to a spec from other directory (I'm not sure that it's supported by luarocks, please, check it).
* User has one scm-1 rockspec (default case)
* User has one scm-2 rockspec: How does luarocks handle this?

cli/build/build.go
```diff
@@ -36,11 +36,15 @@ func Run(ctx *context.Ctx) error {
...
-	if rockspecPath, err := common.FindRockspec(ctx.Project.Path); err != nil {
-		return err
-	} else if rockspecPath == "" {
-		return fmt.Errorf("Application directory should contain rockspec")
+	if ctx.Build.Spec == "" {
+		// check that application directory contains rockspec
+		rockspecPath, err := common.FindRockspec(ctx.Project.Path)
+		if err != nil {
+			return err
+		} else if rockspecPath == "" {
```
mRrvz, line 44: Is this ``else`` branch needed?

```go
if err != nil {
...
}

if rockspecPath == "" {
...
}
```

cli/commands/usage.go
```diff
@@ -16,6 +16,10 @@ const (
 defaults to "package" in the rockspec`
 )
 
+// BUILD
+const (
+	specUsage = `File in .rockspec format for modules creation`
+)
```
mRrvz, line 22: Newline before comment

test/integration/build/test_build.py
```diff
@@ -81,6 +81,63 @@ def test_app_without_rockspec(cartridge_cmd, project_without_dependencies):
     assert 'Application directory should contain rockspec' in output
 
 
+def test_building_with_rockspec_specifying_bad_filename(cartridge_cmd, project_without_dependencies):
+    project = project_without_dependencies
+
+    rockspec_name = "{}-scm-1.rockspec".format(project.name)
+    rocks_make_output = "Error: File not found: {}".format(rockspec_name)
```
mRrvz, line 88: Better to use ``f-strings`` than ``.format``

test/integration/build/test_build.py
```diff
@@ -80,6 +81,272 @@ def test_app_without_rockspec(cartridge_cmd, project_without_dependencies):
     assert rc == 1, 'Building project should fail'
     assert 'Application directory should contain rockspec' in output
 
+    rocks_make_output = "Error: File not found: {}".format(project.rockspec_name)
```
dokshina, line 84: In fact, this is the other test - it's "bad path to rockspec passed" test.

You can simply call `... --spec bad-name` and check error.

Moreover, now this error looks not very nice:

```
Error: File not found: myapp-scm-2.rockspec
   ⨯ Failed to install rocks: Failed to run
/usr/local/bin/tarantoolctl rocks make myapp-scm-2.rockspec

exit status 1
```

I think that CLI can check file existing if `--spec` is specified and show beautiful error.

test/integration/build/test_build.py
```diff
@@ -80,6 +81,272 @@ def test_app_without_rockspec(cartridge_cmd, project_without_dependencies):
...
+
+    # with --spec
```
dokshina, line 86: This is the other test too.

test/integration/build/test_build.py
```diff
@@ -80,6 +81,272 @@ def test_app_without_rockspec(cartridge_cmd, project_without_dependencies):
...
+                version = '{}'
+                source  = {{ url = '/dev/null' }}
+                dependencies = {{ 'tarantool' }}
+                build = {{ type = 'none' }}
+            '''.format(project.name, version))
+
+    build_logs = [
+        'Build application in',
+        'Running `cartridge.pre-build`',
+        'Running `tarantoolctl rocks make {}`'.format(rockspec_path),
```
dokshina, line 118: I think that it's enough to check
```python
'Running `tarantoolctl rocks make` %s' % rockspec_path in logs
```

test/integration/build/test_build.py
```diff
@@ -80,6 +81,272 @@ def test_app_without_rockspec(cartridge_cmd, project_without_dependencies):
...
+    assert rc == 0
+    assert all([log in output for log in build_logs])
+    assert prebuild_output not in output
+
+    # check that all expected rocks was installed
+    files = recursive_listdir(project.path)
+    assert '.rocks' in files
+    assert all([rock in files for rock in project.rocks_content])
+
+    # with *-scm-2.rockspec in project root
```
dokshina, line 195: This test isn't a test of `cartridge build`. In fact, you are testing `rocks make`. But you should test CLI and it's responsibility is to pass correct spec to `rocks make`.

I suggest you removing this test. Building application with one rockspec without `--spec` specifying is a default case that is already tested.

test/integration/build/test_build.py
```diff
@@ -80,6 +81,272 @@ def test_app_without_rockspec(cartridge_cmd, project_without_dependencies):
...
+            '''.format(project.name, version))
+
+    with open(os.path.join(project.path, 'cartridge.pre-build'), 'w') as f:
+        prebuild_script_lines = [
+            "#!/bin/sh",
+            "echo \"{}\"".format(prebuild_output)
+        ]
+        f.write('\n'.join(prebuild_script_lines))
+
+    build_logs = [
```
dokshina, line 282: You shouldn't test all build logs here. Let's concentrate on `tarantoolctl rocks make` in this test. (All other logs are checked in default test, pre-build hook output is used for verbosity flags tests).

test/integration/build/test_build.py
```diff
@@ -80,6 +81,272 @@ def test_app_without_rockspec(cartridge_cmd, project_without_dependencies):
...
+        "--spec",
+        second_rockspec_name,
+    ]
+
+    rc, output = run_command_and_get_output(cmd, cwd=project.path)
+    assert rc == 0
+    assert all([log in output for log in build_logs])
+    assert prebuild_output not in output
+
+    # check that all expected rocks was installed
```
dokshina, line 329: We can check that specified rockspec was used a little tricky.

We can use [command build type](https://github.com/luarocks/luarocks/wiki/Rockspec-format#command) in rockspec and ask it to say who it is. Then, just check which rockspec was called by output.

```lua
build = {
    type = 'command',
    build_command = 'echo I am scm-1',
}```

Note, that you should call `cartridge build --verbose` to handle build logs.

File: cli/commands/pack.go [hunk 12]
```diff
@@ -22,6 +22,7 @@ func init() {
 	packCmd.Flags().StringVar(&ctx.Pack.Version, "version", "", versionUsage)
 	packCmd.Flags().StringVar(&ctx.Pack.Suffix, "suffix", "", suffixUsage)
 	packCmd.Flags().StringSliceVar(&ctx.Pack.ImageTags, "tag", []string{}, tagUsage)
+	packCmd.Flags().StringVar(&ctx.Build.Spec, "spec", "", specUsage)
```
dokshina, line 25: Take a look at `addSpecFlag` function.

test/integration/pack/test_pack.py
```diff
@@ -696,6 +696,103 @@ def test_project_without_stateboard(cartridge_cmd, project_without_dependencies,
     assert '{}-stateboard.service'.format(project.name) not in systemd_files
 
 
+@pytest.mark.parametrize('pack_format', ['rpm', 'deb', 'tgz', 'docker'])
+def test_local_pack_with_two_rockspec_in_project_dir(cartridge_cmd, project_without_dependencies, pack_format, tmpdir):
```
dokshina, line 700: Gennerally, this is a test `pack with spec specified`.

test/integration/build/test_build.py
```diff
@@ -81,6 +81,113 @@ def test_app_without_rockspec(cartridge_cmd, project_without_dependencies):
...
+    dir_name = 'some_dir'
+    dir_path = os.path.join(project.path, dir_name)
+    os.mkdir(dir_path)
+
+    version = 'scm-2'
+    rockspec_name = '%s-%s.rockspec' % (project.name, version)
+    rockspec_path = os.path.join(dir_path, rockspec_name)
+    who_am_i = 'I am %s' % rockspec_name
+
+    with open(rockspec_path, 'w') as f:
```
dokshina, line 115: What about placing this multiple reused code in `project` module? To simply say `set_whoami_build_specs(project, ['scm-1', 'scm-2'])`.

test/project.py
```diff
@@ -40,14 +40,14 @@
 # ###############
 # Helpers
 # ###############
-def get_base_project_rocks(project_name, rockspec_name):
+def get_base_project_rocks(project_name, rockspec_name, version='scm-1'):
```
dokshina, line 43: Seems that now this option is redundant.

test/integration/build/test_build.py
```diff
@@ -81,6 +83,97 @@ def test_app_without_rockspec(cartridge_cmd, project_without_dependencies):
...
+    dir_name = 'some_dir'
+    dir_path = os.path.join(project.path, dir_name)
+    os.mkdir(dir_path)
+
+    version = 'scm-2'
+    rockspec_name = '%s-%s.rockspec' % (project.name, version)
+    rockspec_path = os.path.join(dir_path, rockspec_name)
+    who_am_i = 'I am %s' % rockspec_name
+
+    set_whoami_build_specs(dir_path, project.name, version, rockspec_name)
```
dokshina, line 117: Seems that we can reduce copy-paste.

```python
version = 'scm-2'
rockspec_path = get_rockspec_path(dir_path, project.name,  version)  # let's add such helper to utils

who_am_i = set_whoami_build_specs(rockspec_path, project.name, version)

... 

assert who_am_i in output
```

test/integration/build/test_build.py
```diff
@@ -81,6 +83,97 @@ def test_app_without_rockspec(cartridge_cmd, project_without_dependencies):
...
+
+    build_log = 'Running `tarantoolctl rocks make %s`' % rockspec_path
```
dokshina, line 119: Let's move `build_log` declaration closer to the check. It improve code structure - we have several blocks that performs separated checks.

```python
cmd = [...]

# here we run command and check it's rc
rc, output = run_command_and_get_output(cmd, cwd=project.path)
assert rc == 0

# here we check command output
build_log = ...
assert build_log in output
assert who_am_i in output
```

Comments in snippet above are for you, don't add them to patch.

test/integration/pack/test_pack.py
```diff
@@ -696,6 +698,87 @@ def test_project_without_stateboard(cartridge_cmd, project_without_dependencies,
     assert '{}-stateboard.service'.format(project.name) not in systemd_files
 
 
+@pytest.mark.parametrize('pack_format', ['rpm', 'deb', 'tgz', 'docker'])
```
dokshina, line 701: Let's reduce the number of tests. In fact, building the project is common for all types. Such common parts are tested only for `tgz` (see `test_packing_without_git` for example). It allows to reduce time of running tests.

dokshina: You can also add second paramtrization parameter for local build and build in docker. See `test_app_with_non_existing_hook` test.

test/integration/build/test_build.py
```diff
@@ -81,6 +84,121 @@ def test_app_without_rockspec(cartridge_cmd, project_without_dependencies):
...
+def test_app_with_rockspec_from_other_dir(cartridge_cmd, project_without_dependencies):
+    project = project_without_dependencies
+
+    dir_name = 'some_dir'
+    dir_path = os.path.join(project.path, dir_name)
+    os.mkdir(dir_path)
+
+    version = 'scm-2'
+    rockspec_path = get_rockspec_path(dir_name, project.name, version)
+    who_am_i = set_whoami_build_specs(os.path.join(project.path, rockspec_path), project.name, version)
```
dokshina, line 115: `os.path.join` seems redundant (just pass `dir_name` to `get_rockspec_path`).

cli/build/build.go
```diff
@@ -36,11 +38,27 @@ func Run(ctx *context.Ctx) error {
...
+		}
+	} else {
+		// check that specified file is valid
+		if fileInfo, err := os.Stat(ctx.Build.Spec); os.IsNotExist(err) {
+			return fmt.Errorf("Rockspec %s doesn't exist", ctx.Build.Spec)
+		} else if err != nil {
+			return fmt.Errorf("Unable to use rockspec %s: %s", ctx.Build.Spec, err)
+		} else if fileInfo.IsDir() {
+			return fmt.Errorf("Unable to use rockspec %s: it is a directory", ctx.Build.Spec)
+		}
```
dokshina, line 56: I suggest adding new empty line after this `if` to separate next code block.

cli/build/build.go
```diff
@@ -36,11 +38,27 @@ func Run(ctx *context.Ctx) error {
...
+		absoluteRockspecPath, err := filepath.Abs(ctx.Build.Spec)
+		if err != nil {
+			return err
+		}
+		ctx.Build.Spec = strings.TrimPrefix(absoluteRockspecPath, ctx.Project.Path + "/")
```
dokshina, line 61: Are you sure that it's required?

test/project.py
```diff
@@ -598,3 +598,20 @@ def add_custom_roles(project, roles):
...
+                package = '{}'
+                version = '{}'
+                source  = {{ url = '/dev/null' }}
+                dependencies = {{ 'tarantool' }}
+                build = {{ type = 'command',
+                          build_command = 'echo {}'}}
+            '''.format(project_name, version, who_am_i))
+        f.close()
+
+    return who_am_i
```
mRrvz, line 617: Logic of this function is not clear. It does two things at once:
* Creating rockspec file
* Returning whoami string

From the name of the function, it is not obvious that the function returns whoami string.

mRrvz: maybe move getting who_am_i to another function, like ``get_rockspec_path`` below?

test/utils.py
```diff
@@ -594,6 +594,11 @@ def check_package_files(project, basedir):
     validate_version_file(project, distribution_dir)
 
 
+def get_rockspec_path(path, project_name, version):
+    rockspec_name = '%s-%s.rockspec' % (project_name, version)
+    return os.path.join(path, rockspec_name)
```
mRrvz, line 599:
```python
return os.path.join(path, '%s-%s.rockspec' % (project_name, version))
```

maybe like this? anyway, you dont use this variable below

test/integration/pack/test_pack.py
```diff
@@ -696,6 +699,85 @@ def test_project_without_stateboard(cartridge_cmd, project_without_dependencies,
...
+    # with --spec
+    cmd = [
+        cartridge_cmd,
+        "pack", pack_format,
+        project.path,
+        "--spec",
+        bad_rockspec_path,
+    ]
+
+    if pack_format in ['rpm', 'deb'] and platform.system() == 'Darwin':
```
mRrvz, line 773: In this case use tuple, not list:

```python
if pack_format in ('rpm', 'deb') and platform.system() == 'Darwin':
```

test/integration/pack/test_pack.py
```diff
@@ -696,6 +699,85 @@ def test_project_without_stateboard(cartridge_cmd, project_without_dependencies,
...
+    # tarantoolctl performs build with the oldest version of rockspec files
+    assert who_am_i in output
+
+
+@pytest.mark.parametrize('pack_format', ['tgz'])
+def test_packing_with_rockspec_from_other_dir(cartridge_cmd, project_without_dependencies, pack_format, tmpdir):
+    project = project_without_dependencies
+
+    dir_name = 'some_dir'
+    dir_path = os.path.join(project.path, dir_name)
```
mRrvz, line 734:
```python
dir_path = os.path.join(project.path, 'some_dir')
```

test/integration/pack/test_pack.py
```diff
@@ -696,6 +699,85 @@ def test_project_without_stateboard(cartridge_cmd, project_without_dependencies,
...
+    assert who_am_i in output
+
+
+@pytest.mark.parametrize('pack_format', ['rpm', 'deb', 'tgz', 'docker'])
+def test_pack_with_rockspec_bad_name(cartridge_cmd, project_without_dependencies, pack_format, tmpdir):
+    project = project_without_dependencies
+
+    bad_rockspec_name = "bad_rockspec-scm-1.rockspec"
+    bad_rockspec_path = os.path.join(project.path, bad_rockspec_name)
+    rocks_make_output = "Rockspec %s doesn't exist" % (bad_rockspec_path)
```
mRrvz, line 762: In some places, you use () to pass a single argument to a formatted string, and in others, you [don't](https://github.com/tarantool/cartridge-cli/blob/61e43d1aab391776f1832630871468edf1a80869/test/integration/build/test_build.py#L197). Maybe it's better to stick to one style

cli/build/build.go
```diff
@@ -36,11 +38,28 @@ func Run(ctx *context.Ctx) error {
...
+		// check that specified file is valid
+		if fileInfo, err := os.Stat(ctx.Build.Spec); os.IsNotExist(err) {
+			return fmt.Errorf("Rockspec %s doesn't exist", ctx.Build.Spec)
+		} else if err != nil {
+			return fmt.Errorf("Unable to use rockspec %s: %s", ctx.Build.Spec, err)
+		} else if fileInfo.IsDir() {
+			return fmt.Errorf("Unable to use rockspec %s: it is a directory", ctx.Build.Spec)
+		}
+
+		absoluteRockspecPath, err := filepath.Abs(ctx.Build.Spec)
```
mRrvz, line 58:
```go
if absoluteRockspecPath, err := filepath.Abs(ctx.Build.Spec); err != nil { 
    return err
}
```

cli/build/build.go
```diff
@@ -36,11 +38,28 @@ func Run(ctx *context.Ctx) error {
...
 
-	// check that application directory contains rockspec
-	if rockspecPath, err := common.FindRockspec(ctx.Project.Path); err != nil {
-		return err
-	} else if rockspecPath == "" {
-		return fmt.Errorf("Application directory should contain rockspec")
+	if ctx.Build.Spec == "" {
+		// check that application directory contains rockspec
+		if rockspecPath, err := common.FindRockspec(ctx.Project.Path); err != nil {
+			return err
```
mRrvz, line 44: Some cases we use fmt.Errorf, some cases not. Why?

test/integration/build/test_build.py
```diff
@@ -81,6 +84,89 @@ def test_app_without_rockspec(cartridge_cmd, project_without_dependencies):
...
+def test_app_with_rockspec_from_other_dir(cartridge_cmd, project_without_dependencies):
+    project = project_without_dependencies
+
+    dir_path = os.path.join(project.path, 'some_dir')
+    os.mkdir(dir_path)
+
+    version = 'scm-2'
+    rockspec_path = get_rockspec_path(dir_path, project.name, version)
+
+    rocks_make_output = "Rockspec %s should be in project root" % rockspec_path
```
dokshina, line 115: Move this string closer to `rocks_make_output in output` check:

```go
    rc, output = run_command_and_get_output(cmd, cwd=project.path)
    assert rc == 1, 'Building project should fail'

    rocks_make_output = "Rockspec %s should be in project root" % rockspec_path
    assert rocks_make_output in output
```

test/integration/build/test_build.py
```diff
@@ -81,6 +84,89 @@ def test_app_without_rockspec(cartridge_cmd, project_without_dependencies):
...
+        rockspec_path,
+        "--verbose",
+    ]
+
+    rc, output = run_command_and_get_output(cmd, cwd=project.path)
+    assert rc == 1, 'Building project should fail'
+    assert rocks_make_output in output
+
+
+def test_building_with_two_rockspec_in_project_root(cartridge_cmd, project_without_dependencies):
```
dokshina, line 131: `two_rockspec` -> `two_rockspecS`

test/integration/build/test_build.py
dokshina: And for all tests below.

test/project.py
dokshina: It's very good comment.

Usually such functions are named `doSmthAndReturnSmth`. I don't like moving getting whoami to function, because it's too specific for this test.

What about `set_and_return_whoami_on_build`?

dokshina (approved):
Good!

PR #525, opened 2021-03-29
Improved ``cartridge connect`` responsiveness

### LINKED ISSUE #493: [2pt] cartridge connect significantly slower than tarantoolctl connect
Bootstrap cluster with vshard (30k buckets).
Call `cartridge.service_get('vshard-router').get('default')` in console.

Yes, output is quite huge and it's a mistake actually. But `tarantoolctl` handles this case quite faster than `cartridge-cli`
that simply hangs up to 10 seconds.

mRrvz:
Improved responsiveness of the ``cartridge enter`` and ``cartridge enter`` commands. Requests that work with a large amount of data have become faster. Closes #493

### DIFF
--- a/cli/connector/eval_plain_text.go
+++ b/cli/connector/eval_plain_text.go
@@ -2,6 +2,7 @@ package connector
 
 import (
 	"bufio"
+	"bytes"
 	"encoding/base64"
 	"fmt"
 	"io"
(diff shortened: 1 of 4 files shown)


dokshina (dismissed):
Good!

cli/connector/eval_plain_text.go
```diff
@@ -212,32 +216,66 @@ func readDataPortionFromPlainTextConn(conn net.Conn, readTimeout time.Duration)
...
-				break
-			}
+		nextByte, err := buffer.ReadByte()
+		if err != nil {
+			return nil, fmt.Errorf("Failed to get byte from buffer: %s", err)
+		}
+		f, err := os.OpenFile("ayo228.txt", os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0600)
+		if err != nil {
+			panic(err)
+		}
```
olegrok, line 248: Is it really needed?

cli/connector/eval_plain_text.go
```diff
@@ -59,8 +60,10 @@ func evalPlainTextConn(conn net.Conn, funcBody string, args []interface{}, opts
 		return nil, err
 	}
 
+	buffer := bytes.Buffer{}
```
dokshina, line 63: Why do you declare buffer here? Seems that it's used only by `readFromPlainTextConn`.

dokshina (approved):

PR #530, opened 2021-04-06
``cartridge pack``: external dependencies

### LINKED ISSUE #488: [2pt] rpm deps: please add parameters to handle external dependencies
There are cases when carg application uses some database connector. Database connector uses system database client library.

For e.g.

- postgresql connector uses libpq.so (from postgresq-libs)
- odbc connector uses postgresql-odbc
- etc

Could you please add extending rpm deps section for such cases?

mRrvz:
Added ability to specify dependencies for the RPM 
and DEB (command``cartridge pack``), using the 
``--deps`` and ``--deps-file`` flags. Closes #488

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -761,6 +761,43 @@ The options (``[flags]``) are as follows:
   flag that indicates if the SDK from the local machine should be delivered in the
   result artifact.
 
+* ``--deps`` (used for ``rpm`` and ``deb`` packages) is the dependencies of the package.
+
+* ``--deps-file`` (used for ``rpm`` and ``deb`` packages) is the path to the file file which
+  contains dependencies of the package. Defaults to ``package-deps.txt`` in the application root.
+
+Example of file with package dependencies:
+
+```
+dependency_01 >= 2.5
+dependency_01 <
+dependency_02 >= 1, < 5
+dependency_03==2
+dependency_04<5,>=1.5
+```
+
+One line should contain the description of only one dependency, but at the same
+time you can specify both the major and the minor version on this line:
+
+```
+dependency_05 >= 4, < 5
+```
+
+The format of the dependencies for the ``--dep`` flag is similar to the
+``--deps-file`` flag, except that you cannot specify the major and minor
+version of the dependency separated by commas:
+
+```bash
+# You can't do that!
+cartridge pack rpm --deps dependency_06>=4,<5 appname
+
+# The command above can be rewritten like this:
+cartridge pack rpm --deps dependency_06>=4,dependency_06<5 appname
+
+# Or like this:
+cartridge pack rpm --deps dependency_06>=4 --deps dependency_06<5 appname
+```
+
 For Tarantool Enterprise, you must specify one (and only one)
 of the ``--sdk-local`` and ``--sdk-path`` options.
 
--- a/README.rst
+++ b/README.rst
@@ -783,7 +783,7 @@ time you can specify both the major and the minor version on this line:
 
     dependency_05 >= 4, < 5
 
-The format of the dependencies for the ``--dep`` flag is similar to the
+The format of the dependencies for the ``--deps`` flag is similar to the
 ``--deps-file`` flag, except that you cannot specify the major and minor
 version of the dependency separated by commas:
 
--- a/README.rst
+++ b/README.rst
@@ -768,35 +768,35 @@ The options (``[flags]``) are as follows:
 
 Example of file with package dependencies:
 
-```
-dependency_01 >= 2.5
-dependency_01 <
-dependency_02 >= 1, < 5
-dependency_03==2
-dependency_04<5,>=1.5
-```
+.. code-block:: text
+
+    dependency_01 >= 2.5
+    dependency_01 <
+    dependency_02 >= 1, < 5
+    dependency_03==2
+    dependency_04<5,>=1.5.3
 
 One line should contain the description of only one dependency, but at the same
 time you can specify both the major and the minor version on this line:
 
-```
-dependency_05 >= 4, < 5
-```
+.. code-block:: bash
+
+    dependency_05 >= 4, < 5
 
 The format of the dependencies for the ``--dep`` flag is similar to the
 ``--deps-file`` flag, except that you cannot specify the major and minor
 version of the dependency separated by commas:
 
-```bash
-# You can't do that!
-cartridge pack rpm --deps dependency_06>=4,<5 appname
+.. code-block:: bash
 
-# The command above can be rewritten like this:
-cartridge pack rpm --deps dependency_06>=4,dependency_06<5 appname
+    # You can't do that:
+    cartridge pack rpm --deps dependency_06>=4,<5 appname
 
-# Or like this:
-cartridge pack rpm --deps dependency_06>=4 --deps dependency_06<5 appname
-```
+    # The command above can be rewritten like this:
+    cartridge pack rpm --deps dependency_06>=4,dependency_06<5 appname
+
+    # Or like this:
+    cartridge pack rpm --deps dependency_06>=4 --deps dependency_06<5 appname
 
 For Tarantool Enterprise, you must specify one (and only one)
 of the ``--sdk-local`` and ``--sdk-path`` options.
--- a/cli/commands/pack.go
+++ b/cli/commands/pack.go
@@ -42,7 +42,8 @@ func init() {
 		&ctx.Pack.StatboardUnitTemplatePath, "stateboard-unit-template", "", stateboardUnitTemplateUsage,
 	)
 
-	packCmd.Flags().StringVar(&ctx.Pack.DependenciesFile, "deps", "", dependenciesFileUsage)
+	packCmd.Flags().StringSliceVar(&ctx.Pack.Deps, "deps", []string{}, depsUsage)
+	packCmd.Flags().StringVar(&ctx.Pack.DepsFile, "deps-file", "", depsFileUsage)
 }
 
 var packCmd = &cobra.Command{
--- a/cli/commands/pack.go
+++ b/cli/commands/pack.go
@@ -55,17 +55,19 @@ func init() {
 }
 
 func addTarantoolDepIfNeeded(ctx *context.Ctx) error {
-	if !ctx.Tarantool.TarantoolIsEnterprise && (ctx.Pack.Type == pack.RpmType || ctx.Pack.Type == pack.DebType) {
-		var tarantoolVersion string
-		if ctx.Pack.Type == pack.RpmType {
-			tarantoolVersion = strings.SplitN(ctx.Tarantool.TarantoolVersion, "-", 2)[0]
-		} else if ctx.Pack.Type == pack.DebType {
-			tarantoolVersion = ctx.Tarantool.TarantoolVersion
-		}
+	if ctx.Tarantool.TarantoolIsEnterprise || !(ctx.Pack.Type == pack.RpmType || ctx.Pack.Type == pack.DebType) {
+		return nil
+	}
 
-		if err := ctx.Pack.Deps.AddTarantool(tarantoolVersion); err != nil {
-			return fmt.Errorf("Failed to get tarantool dependency: %s", err)
-		}
+	var tarantoolVersion string
+	if ctx.Pack.Type == pack.RpmType {
+		tarantoolVersion = strings.SplitN(ctx.Tarantool.TarantoolVersion, "-", 2)[0]
+	} else if ctx.Pack.Type == pack.DebType {
+		tarantoolVersion = ctx.Tarantool.TarantoolVersion
+	}
+
+	if err := ctx.Pack.Deps.AddTarantool(tarantoolVersion); err != nil {
+		return fmt.Errorf("Failed to get tarantool dependency: %s", err)
 	}
 
 	return nil
--- a/cli/commands/pack.go
+++ b/cli/commands/pack.go
@@ -41,6 +41,8 @@ func init() {
 	packCmd.Flags().StringVar(
 		&ctx.Pack.StatboardUnitTemplatePath, "stateboard-unit-template", "", stateboardUnitTemplateUsage,
 	)
+
+	packCmd.Flags().StringVar(&ctx.Pack.DependenciesFile, "deps", "", dependenciesFileUsage)
 }
 
 var packCmd = &cobra.Command{
--- a/cli/commands/pack.go
+++ b/cli/commands/pack.go
@@ -1,18 +1,25 @@
 package commands
 
 import (
+	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/apex/log"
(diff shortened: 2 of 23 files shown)


cli/common/utils.go
```diff
@@ -15,12 +15,20 @@ import (
...
 	"github.com/mitchellh/mapstructure"
 	"github.com/tarantool/cartridge-cli/cli/context"
 	"github.com/vmihailenco/msgpack/v5"
 	"gopkg.in/yaml.v2"
 )
 
+type PackDependency struct {
+	Name     string `@Ident`
+	Relation string `@( "=" "=" | "=" | ">" "=" | "<" "=" | ">" | "<" )`
+	Version  string ` @Number`
```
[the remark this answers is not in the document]
dokshina: Lets use named tags `parser:"@Ident"`.

cli/common/utils.go
```diff
@@ -489,6 +497,55 @@ func wrap(i, w int, s string) string {
 	return r
 }
 
+func ParseDependenciesFile(filepath string) ([]PackDependency, error) {
+	content, err := GetFileContent(filepath)
```
dokshina, line 501: This `err` variable is re-used by code below. In this case it's better to implicitly declare `var err error` at the top of function.

cli/common/utils.go
```diff
@@ -489,6 +497,55 @@ func wrap(i, w int, s string) string {
...
+		participle.Elide("Comment", "Whitespace"),
+	)
+
+	dependencies := []PackDependency{}
+	for _, line := range strings.Split(content, "\n") {
+		if line == "" {
+			continue
+		}
+
+		// We do this to process lines like: dependency>=2.4,<5
```
dokshina, line 526: Why don't you describe it in the grammar?
`PackDependency` structure can be a little more difficult: 

```
type DepRelation struct {
    Version string
    Relation string
}

type PackDependency struct {
	Name     string `@Ident`
	Relations []DepRelation
}
```

cli/context/context.go
```diff
@@ -104,6 +104,8 @@ type PackCtx struct {
 	UnitTemplatePath          string
 	InstUnitTemplatePath      string
 	StatboardUnitTemplatePath string
+
+	DependenciesFile string
```
dokshina, line 108: Let's use `Deps` instead of `Dependencies`.

README.rst
```diff
@@ -761,6 +761,43 @@ The options (``[flags]``) are as follows:
...
+    dependency_04<5,>=1.5.3
+
+One line should contain the description of only one dependency, but at the same
+time you can specify both the major and the minor version on this line:
+
+.. code-block:: bash
+
+    dependency_05 >= 4, < 5
+
+The format of the dependencies for the ``--dep`` flag is similar to the
```
dokshina, line 786: `--deps`

cli/common/utils.go
```diff
@@ -15,18 +15,86 @@ import (
...
+			switch r.Relation {
+			case ">":
+				relation = rpmSenseGreater
+			case ">=":
+				relation = rpmSenseGreater | rpmSenseEqual
+			case "<":
+				relation = rpmSenseLess
+			case "<=":
+				relation = rpmSenseLess | rpmSenseEqual
+			case "=":
```
dokshina, line 83: case "=", "=="

cli/common/utils.go
```diff
@@ -15,18 +15,86 @@ import (
...
+		}
+
+		debDeps = append(debDeps, dependency)
+	}
+
+	return debDeps
+}
+
+func (deps PackDependencies) FormatRPM() PackDependencies {
+	// We can't get constants from rpm package - cycle imports not allowed.
```
dokshina, line 64: Yes, this looks bad.

Seems that `rpm` package can call some function that accepts `deps PackDependencies` and converts them directly to `rpmTagType`.

cli/pack/deb_control_dir.go
```diff
@@ -55,11 +68,36 @@ func initControlDir(destDirPath string, ctx *context.Ctx) error {
...
+				{
+					Relation: "<<",
+					Version:  maxTarantoolVersion,
+				},
+			},
+		})
+	}
+
+	// Parse and add dependencies
+	if len(ctx.Pack.Deps) != 0 {
```
dokshina, line 87: This action should be performed on context preparing. 
The same for `rpm`.

You can do it on `pack.Run` and save somethere to `ctx.Pack.ParsedDeps`.
Or, you can perform this in `commands.pack`. Just parse specified flags to local variables (`deps` and `depsFile`), then call `common.ParseDependencies` and save the result to `ctx.Pack.Deps`.

cli/pack/pack.go
```diff
@@ -205,6 +220,32 @@ func FillCtx(ctx *context.Ctx) error {
 		log.Warnf("Specified %s is ignored", sdkPathEnv)
 	}
 
+	if ctx.Pack.Type == RpmType || ctx.Pack.Type == DebType {
+		if ctx.Pack.DepsFile == "" && len(ctx.Pack.Deps) == 0 {
+			defaultPackeDepsFilePath := filepath.Join(ctx.Project.Path, defaultPackageDepsFile)
+			if _, err := os.Stat(defaultPackeDepsFilePath); err == nil {
```
dokshina, line 226: If `!os.IsNotExists(err)` then the file is present, but we can't use it.

cli/pack/pack.go
```diff
@@ -205,6 +220,32 @@ func FillCtx(ctx *context.Ctx) error {
...
+				log.Debugf("Default build Dockerfile is used: %s", defaultPackeDepsFilePath)
+
+				ctx.Pack.DepsFile = defaultPackeDepsFilePath
+			} else if !os.IsNotExist(err) {
+				return fmt.Errorf("Failed to use default package dependencies file: %s", err)
+			}
+		}
+	}
+
+	if ctx.Pack.DepsFile != "" && (ctx.Pack.Type == RpmType || ctx.Pack.Type == DebType) {
```
dokshina, line 236: What if pack type is other that rpm and deb? We just ignore this file?

cli/rpm/header.go
```diff
@@ -93,28 +125,51 @@ func genRpmHeader(relPaths []string, cpioPath, compresedCpioPath string, ctx *co
...
 
-	if !ctx.Tarantool.TarantoolIsEnterprise {
-		// add Tarantool dependency
-		tarantoolDepName := "tarantool"
-		flagGreaterOrEqual := int32(rpmSenseGreater | rpmSenseEqual)
+	var deps common.PackDependencies
 
-		minVersion := strings.SplitN(ctx.Tarantool.TarantoolVersion, "-", 2)[0]
-		maxVersion, err := common.GetNextMajorVersion(minVersion)
+	if !ctx.Tarantool.TarantoolIsEnterprise {
```
dokshina, line 130: This part is now common for both rpm and deb.
Simply realize `deps.Add()` function.

cli/pack/deb_control_dir.go
```diff
@@ -61,40 +80,21 @@ func initControlDir(destDirPath string, ctx *context.Ctx) error {
...
-				},
-				{
-					Relation: "<<",
-					Version:  maxTarantoolVersion,
-				},
-			},
-		})
+	deps = append(deps, ctx.Pack.Deps...)
+	for _, dependency := range formatDeb(deps) {
+		addDependency(&debControlCtx, dependency)
```
dokshina, line 93: Why did you separated `formatDeb`? This logic can be placed directly in `addDependency` function. Moreover, it can be `addDependencies` function that hides a loop.

cli/rpm/header.go
```diff
@@ -126,48 +178,14 @@ func genRpmHeader(relPaths []string, cpioPath, compresedCpioPath string, ctx *co
...
-		}
-
-		deps = append(deps, parsedDeps...)
-	}
-
-	if len(deps) != 0 {
-		addDependencies(&rpmHeader, deps.FormatRPM())
-	}
+	deps = append(deps, ctx.Pack.Deps...)
+	addDependencies(&rpmHeader, formatRPM(deps))
```
dokshina, line 188: The same as for deb - why do you separate `formatRPM` from `addDependencies`?

cli/common/utils_test.go
```diff
@@ -115,39 +114,121 @@ func TestGetInstancesFromArgs(t *testing.T) {
...
+
+	assert := assert.New(t)
+
+	rawDeps := []string{
+		"dependency_01 > 1.2, <= 4",
+		"dependency_02 < 7,>=1.5",
+		"dependency_03==2.8",
+		"	dependency_04   <= 5.2   ",
+		"dependency_05>=2.4,<=5.1",
+		"dependency_06 =15",
```
dokshina, line 164: Maybe add some empty lines?

cli/pack/deb_control_dir.go
```diff
@@ -34,6 +34,47 @@ var (
...
+			(*debControlCtx)["Depends"] = fmt.Sprintf("%s%s, ", (*debControlCtx)["Depends"], depsString)
+		}
+
+		if len(dep.Relations) == 0 {
+			(*debControlCtx)["Depends"] = fmt.Sprintf("%s%s, ", (*debControlCtx)["Depends"], dep.Name)
+		}
+	}
+
+	// cut last ', ' symbols created by code before
+	if len(deps) != 0 {
```
dokshina, line 52: What about using `strings.Join(depStr, ", ")`?

cli/pack/deb_control_dir.go
```diff
@@ -48,20 +89,17 @@ func initControlDir(destDirPath string, ctx *context.Ctx) error {
...
 	}
 
+	var deps common.PackDependencies
 	if !ctx.Tarantool.TarantoolIsEnterprise {
-		minTarantoolVersion := ctx.Tarantool.TarantoolVersion
-		maxTarantoolVersion, err := common.GetNextMajorVersion(minTarantoolVersion)
-		if err != nil {
-			return project.InternalError("Failed to get next Tarantool major version: %s", err)
+		var err error
+		if deps, err = ctx.Pack.Deps.AddTarantool(ctx.Tarantool.TarantoolVersion); err != nil {
```
dokshina, line 95: This code is common for rpm and deb, it can be performed in `pack` module once.

BTW, why do you save this to `deps` variable?
It can be `func (deps *PackDependencies) AddTarantool` that modifies list inplace.

cli/rpm/header.go
```diff
@@ -30,8 +30,92 @@ type filesInfoType struct {
...
+func addDependencies(rpmHeader *rpmTagSetType, deps common.PackDependencies) {
+	if len(deps) == 0 {
+		return
+	}
+
+	var names []string
+	var versions []string
+	var relations []int32
+
+	for _, dep := range formatRPM(deps) {
```
dokshina, line 57: Think about `getRPMRelation(dep.Relation)`.

test/integration/pack/test_pack.py
```diff
@@ -990,3 +990,279 @@ def test_verbosity(cartridge_cmd, project_without_dependencies, pack_format):
...
+    if platform.system() == 'Darwin':
+        cmd.append('--use-docker')
+
+    rc, output = run_command_and_get_output(cmd, cwd=tmpdir)
+    assert rc == 1
+    assert "Invalid path to file with dependencies" in output
+
+
+@pytest.mark.parametrize('pack_format', ['deb', 'rpm'])
+def test_broken_dependencies(cartridge_cmd, project_without_dependencies, pack_format, tmpdir):
```
dokshina, line 1197: Seems redundantly - you've tested that in unit tests.

cli/commands/pack.go
```diff
@@ -41,6 +48,79 @@ func init() {
...
+
+		content, err := common.GetFileContent(depsFile)
+		if err != nil {
+			return fmt.Errorf("Failed to get file content: %s", err)
+		}
+
+		deps = strings.Split(content, "\n")
+	}
+
+	if !ctx.Tarantool.TarantoolIsEnterprise && (ctx.Pack.Type == pack.RpmType || ctx.Pack.Type == pack.DebType) {
```
dokshina, line 85: It's not about "parse".

What about 

```
fillDependencies(ctx *context.Ctx) {
  deps := parsePackageDeps(deps, depsFile)
  addTarantoolDepIfNeeded(ctx)
}
```

?

I'd like variables to be passed explicitly.

test/integration/pack/test_pack.py
mRrvz: No, unit tests do not check for incorrect dependencies (only correct)

dokshina: It would be better to move such tests into units.
Unit tests check that files are parsed correctly, they can test "error" cases too.
Integration tests check the whole CLI behavior, so you can place here one "error" case and validate that user will see nice error.

test/integration/pack/test_pack.py
```diff
@@ -990,3 +990,279 @@ def test_verbosity(cartridge_cmd, project_without_dependencies, pack_format):
...
+        assert_dependencies_deb(find_archive(tmpdir, project.name, 'deb'), deps, tmpdir)
+    else:
+        deps = (
+            ("dependency01", 0x08 | 0x04, "1.2"),  # >=
+            ("dependency01", 0x02, "3"),  # <
+            ("dependency02", 0x08, "2.5"),  # =
+            ("dependency03", 0, "")
+        )
+
+        assert_dependencies_rpm(find_archive(tmpdir, project.name, 'rpm'), deps)
```
dokshina, line 1031: Why don't you check Tarantool dependencies here?

test/integration/pack/test_pack.py
```diff
@@ -990,3 +1002,266 @@ def test_verbosity(cartridge_cmd, project_without_dependencies, pack_format):
...
+        assert_dependencies_deb(find_archive(tmpdir, project.name, 'deb'), deps, tmpdir)
+    else:
+        deps = (
+            ("dependency01", 0x08 | 0x04, "1.2"),  # >=
+            ("dependency01", 0x02, "3"),  # <
+            ("dependency02", 0x08, "2.5"),  # =
+            ("dependency03", 0, "")
+        )
+
+        if not tarantool_enterprise_is_used():
```
dokshina, line 1165: Seems like a function (not copy-paste). In fact, it can be moved to `assert_dependencies_*` functions.

cli/common/utils_test.go
```diff
@@ -115,39 +114,173 @@ func TestGetInstancesFromArgs(t *testing.T) {
...
+		"to parse: dependency , ,", err.Error())
+	assert.Equal(PackDependencies(nil), deps)
+
+	rawDeps = []string{
+		"dependency >= 3.2, < ",
+	}
+
+	deps, err = ParseDependencies(rawDeps)
+
+	assert.Equal("Error during parse dependencies file: 1:21: unexpected token \"<EOF>\" "+
```
dokshina, line 283: I think that `"Error during parse dependencies file" in err.Error()` check is enough.

dokshina (approved):

PR #545, opened 2021-05-12
Allowed to pack RPM and DEB packages without default dependencies file

mRrvz:
Allowed to pack `RPM` and `DEB` packages with ``cartridge pack``
comamnd without default dependencies file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 
 - Allowed to pack RPM and DEB packages with ``cartridge pack``
-  comamnd without default dependencies file.
+  command without default dependencies file.
 
 ## [2.9.0] - 2021-04-26
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- Allowed to pack RPM and DEB packages with ``cartridge pack``
+  comamnd without default dependencies file.
+
 ## [2.9.0] - 2021-04-26
 
 ### Changed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- Allowed to pack RPM and DEB packages with ``cartridge pack``
+  comamnd without default dependencies file.
```
dokshina, line 13: `comamnd` -> `command`

cli/commands/pack.go
```diff
@@ -81,12 +81,13 @@ func parsePackageDeps(deps []string, depsFile string) (common.PackDependencies,
 	}
 
 	if depsFile == "" && len(deps) == 0 {
-		defaultPackeDepsFilePath := filepath.Join(ctx.Project.Path, defaultPackageDepsFile)
-		if _, err := os.Stat(defaultPackeDepsFilePath); err != nil {
-			return nil, fmt.Errorf("Failed to use default package dependencies file: %s", err)
+		defaultPackDepsFilePath := filepath.Join(ctx.Project.Path, defaultPackageDepsFile)
+		if _, err := os.Stat(defaultPackDepsFilePath); err != nil {
```
dokshina, line 85: Here we need to separate two cases: 
* `os.IsNotExists(err)` - file doesn't exist, used didn't specify some deps, no warns, no errors;
* `err != nil` - user specified a file, but we can't use it - error;
* `err == nil` - file exists, and we can use it.

cli/commands/pack.go
```diff
@@ -81,12 +81,16 @@ func parsePackageDeps(deps []string, depsFile string) (common.PackDependencies,
...
+			if os.IsNotExist(err) {
+				defaultPackDepsFilePath = ""
+			} else {
+				return nil, fmt.Errorf("Failed to use default package dependencies file: %s", err)
+			}
```
dokshina, line 90: There is no reason for an additional `if` level.
Moreover, you shouldn't rewrite `defaultPackDepsFilePath`:

```go
if _, err := os.Stat(defaultPackDepsFilePath); err == nil {
    depsFile = defaultPackDepsFilePath
} else if !os.IsNotExist(err) {
    return nil, fmt.Errorf("Failed to use default package dependencies file: %s", err)
}
```

dokshina (approved):

PR #553, opened 2021-05-21
Change ``cartridge version`` behaviour

### LINKED ISSUE #526: [1pt] Users confuse cartidge rock and cartridge cli versions
Some of our users confuse `cartridge --version` output with cartridge rock veriosn that is used in their project.
I bet we could improve wording to remove this confusion

mRrvz:
``cartridge version`` and ``cartridge --version`` commands
  now show Cartridge CLI and Cartridge versions. Moreover, they
  can show version of the project rocks.

<details> 
  <summary>Example (images) </summary>
 
img2

</details>

Closes #526

### DIFF
--- a/cli/commands/cartridge.go
+++ b/cli/commands/cartridge.go
@@ -1,6 +1,8 @@
 package commands
 
 import (
+	"fmt"
+
 	"github.com/apex/log"
 	"github.com/apex/log/handlers/cli"
 	"github.com/spf13/cobra"
@@ -9,15 +11,18 @@ import (
 )
 
 var (
-	ctx context.Ctx
-
-	rootCmd = &cobra.Command{
+	ctx         context.Ctx
+	needVersion bool
+	rootCmd     = &cobra.Command{
 		Use:   "cartridge",
 		Short: "Tarantool Cartridge command-line interface",
 
-		Version: version.BuildCliVersionString(),
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
 			setLogLevel()
+		},
+
+		Run: func(cmd *cobra.Command, args []string) {
+			printVersion()
 		},
 	}
 )
@@ -28,6 +33,7 @@ func init() {
 	rootCmd.PersistentFlags().BoolVar(&ctx.Cli.Verbose, "verbose", false, "Verbose output")
 	rootCmd.PersistentFlags().BoolVar(&ctx.Cli.Quiet, "quiet", false, "Hide build commands output")
 	rootCmd.PersistentFlags().BoolVar(&ctx.Cli.Debug, "debug", false, "Debug mode")
+	rootCmd.PersistentFlags().BoolVar(&needVersion, "version", false, "Show version information")
 
 	initLogger()
 }
@@ -53,5 +59,11 @@ func setLogLevel() {
 
 	if ctx.Cli.Quiet {
 		log.SetLevel(log.ErrorLevel)
+	}
+}
+
+func printVersion() {
+	if needVersion {
+		fmt.Println(version.BuildVersionString(".", false))
 	}
 }
--- a/cli/commands/cartridge.go
+++ b/cli/commands/cartridge.go
@@ -1,8 +1,6 @@
 package commands
 
 import (
-	"os"
-
 	"github.com/apex/log"
 	"github.com/apex/log/handlers/cli"
 	"github.com/spf13/cobra"
@@ -68,6 +66,6 @@ func setLogLevel() {
 
 func printVersion(cmd *cobra.Command) {
 	if err := version.PrintVersionString(projectPath, cmd.Flags().Changed("project-path"), showRocksVersion); err != nil {
-		os.Exit(1)
+		log.Fatalf(err.Error())
 	}
 }
--- a/cli/commands/cartridge.go
+++ b/cli/commands/cartridge.go
@@ -67,5 +67,12 @@ func setLogLevel() {
 }
 
 func printVersion() {
-	fmt.Println(version.BuildVersionString(".", false))
+	versionString, err := version.BuildVersionString(projectPath, showRocksVersion)
+	fmt.Println(versionString)
+
+	if err != nil && showRocksVersion {
+		log.Errorf("%s", err)
+	} else if err != nil {
+		log.Warnf("%s", err)
+	}
 }
--- a/cli/commands/cartridge.go
+++ b/cli/commands/cartridge.go
@@ -65,7 +65,8 @@ func setLogLevel() {
 }
 
 func printVersion(cmd *cobra.Command) {
-	if err := version.PrintVersionString(projectPath, cmd.Flags().Changed("project-path"), showRocksVersion); err != nil {
+	projectPathIsSet := cmd.Flags().Changed("project-path")
+	if err := version.PrintVersionString(projectPath, projectPathIsSet, showRocksVersions); err != nil {
 		log.Fatalf(err.Error())
 	}
 }
--- a/cli/commands/cartridge.go
+++ b/cli/commands/cartridge.go
@@ -1,7 +1,7 @@
 package commands
 
 import (
-	"fmt"
+	"os"
 
 	"github.com/apex/log"
 	"github.com/apex/log/handlers/cli"
@@ -25,7 +25,7 @@ var (
 			if len(args) == 0 && !needVersion {
 				cmd.Help()
 			} else {
-				printVersion()
+				printVersion(cmd)
 			}
 		},
 	}
@@ -66,13 +66,8 @@ func setLogLevel() {
 	}
 }
 
-func printVersion() {
-	versionString, err := version.BuildVersionString(projectPath, showRocksVersion)
-	fmt.Println(versionString)
-
-	if err != nil && showRocksVersion {
-		log.Errorf("%s", err)
-	} else if err != nil {
-		log.Warnf("%s", err)
+func printVersion(cmd *cobra.Command) {
+	if err := version.PrintVersionString(projectPath, cmd.Flags().Changed("project-path"), showRocksVersion); err != nil {
+		os.Exit(1)
 	}
 }
--- a/cli/commands/cartridge.go
+++ b/cli/commands/cartridge.go
@@ -36,6 +36,7 @@ func init() {
 	rootCmd.PersistentFlags().BoolVar(&ctx.Cli.Quiet, "quiet", false, "Hide build commands output")
 	rootCmd.PersistentFlags().BoolVar(&ctx.Cli.Debug, "debug", false, "Debug mode")
 	rootCmd.Flags().BoolVarP(&needVersion, "version", "v", false, "Show version information")
+	addVersionFlags(rootCmd.Flags())
 
 	initLogger()
 }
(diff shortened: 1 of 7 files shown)


cli/version/version.go
```diff
@@ -19,9 +19,32 @@ var (
...
+	var rocksVersionsMap map[string]string
+	var err error
+
+	if rocksVersionsMap, err = common.LuaGetRocksVersions(projectPath); err != nil {
+		versionParts = append(versionParts, fmt.Sprintf("%s. See --project-path flag", err))
+		return strings.Join(versionParts, "\n ")
+	}
+
+	if len(rocksVersionsMap) == 0 {
+		versionParts = append(versionParts, fmt.Sprintf("Specify project path to see rocks version. See --project-path flag"))
```
dokshina, line 36: What if the project path doesn't contain `.rocks` directory?

cli/version/version.go
```diff
@@ -19,9 +19,32 @@ var (
...
+		return strings.Join(versionParts, "\n ")
+	}
+
+	for k, v := range rocksVersionsMap {
+		versionParts = append(versionParts, fmt.Sprintf("%s version: %s", k, v))
```
dokshina, line 41: Are we really need `version` word on each line?

dokshina (commented):
I am sure that this PR needs tests.

mRrvz:
img1

img2

rosik:
I suggest making `--rock` the default behavior.

cli/version/version.go
```diff
@@ -41,16 +101,39 @@ func BuildVersionString() string {
 		}
 	}
 
-	versionStr := fmt.Sprintf("v%s", version)
+	versionStr := fmt.Sprintf("Version:\t%s", version)
 	versionParts = append(versionParts, versionStr)
 
-	osArchStr := fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH)
+	osArchStr := fmt.Sprintf("OS/Arch:\t%s/%s", runtime.GOOS, runtime.GOARCH)
```
dokshina, line 107: It can be helpful to use `text/template` here. Just create a map with the keys you need and use it with the template.

cli/version/version.go
```diff
@@ -17,15 +21,75 @@ var (
...
+		return nil, fmt.Errorf("%s. Your project path is invalid", errorStr)
+	}
+
+	if rocksVersionsMap, err = common.LuaGetRocksVersions(projectPath); err != nil {
+		return nil, fmt.Errorf("%s. %s. See --project-path flag", err, errorStr)
+	}
+
+	if len(rocksVersionsMap) == 0 {
+		return nil, fmt.Errorf(`%s. Looks like your project directory
+does not contain a .rocks directory... Did your built your project?`, errorStr)
```
dokshina, line 54: `Did your` -> `Did you`.

I sure that we should separate cases "path is not a project" and "path is a project without .rocks".

I think it's OK to think that "path is a project" if it contains a rockspec.

cli/version/version.go
```diff
@@ -17,15 +21,75 @@ var (
 )
 
 const (
-	unknownVersion = "<unknown>"
-	cliName        = "Tarantool Cartridge CLI"
+	unknownVersion        = "<unknown>"
+	cliName               = "Tarantool Cartridge CLI"
+	cartridgeVersionTitle = "Tarantool Cartridge"
+	errorStr              = "Failed to get the version of the Cartridge"
```
dokshina, line 27: This variable name is very bad =) It doesn't say anything about this error and where it should be used.

cli/version/version.go
```diff
@@ -17,15 +21,75 @@ var (
...
 )
 
-func BuildVersionString() string {
-	var version string
+func format(template string, templateArgs map[string]string) string {
```
dokshina, line 30: `formatVersion`

cli/version/version.go
```diff
@@ -17,15 +21,75 @@ var (
...
+
+	return versionMsg
+}
+
+func getRocksVersions(projectPath string) (map[string]string, error) {
+	var rocksVersionsMap map[string]string
+	var err error
+
+	if _, err := os.Stat(filepath.Join(projectPath)); os.IsNotExist(err) {
+		return nil, fmt.Errorf("%s. Your project path is invalid", errorStr)
```
dokshina, line 45: What about other errors?

cli/version/version.go
```diff
@@ -17,15 +21,75 @@ var (
...
+	}
+
+	if rocksVersionsMap, err = common.LuaGetRocksVersions(projectPath); err != nil {
+		return nil, fmt.Errorf("%s. %s. See --project-path flag", err, errorStr)
```
dokshina, line 49: Why not

```go
return nil, err
```
?

This error can be handed in `BuildVersionString` function. If `--rocks` flag was specified, it's an error, otherwise - a warning.

cli/version/version.go
```diff
@@ -41,16 +105,44 @@ func BuildVersionString() string {
...
 
-	return strings.Join(versionParts, " ")
+	return strings.Join(versionParts, "\n\n")
 }
+
+var (
+	cliVersionTmpl = `{{ .Title }}
+ Version:	{{ .Version }}
+ OS/Arch: 	{{ .OS }}/{{ .Arch }}
+ Git commit: {{ .Commit }}`
```
dokshina, line 144: Simply add `\n` here to avoid using `\n\n` in `strings.Join`.

test/integration/cli/test_version.py
```diff
@@ -1,8 +1,73 @@
...
+
+def test_version_command_invalid_project(cartridge_cmd):
+    cmd = [
+        cartridge_cmd, "version",
+        "--project-path=invalid_path"
+    ]
+
+    rc, output = run_command_and_get_output(cmd)
+    assert rc == 0
+    assert 'Failed to get the version of the Cartridge. Your project path is invalid' in output
```
dokshina, line 73: What about other error cases?

cli/commands/cartridge.go
```diff
@@ -55,3 +65,14 @@ func setLogLevel() {
 		log.SetLevel(log.ErrorLevel)
 	}
 }
+
+func printVersion() {
+	versionString, err := version.BuildVersionString(projectPath, showRocksVersion)
+	fmt.Println(versionString)
+
+	if err != nil && showRocksVersion {
```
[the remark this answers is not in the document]
dokshina: The possible solution is to delegate it to `BuildVersionString` function. 

Let it be `PrintVersionString` function.

```go
if err := version.PrintVersionString(projectPath, projectPathIsSet, showRocksVersion); err != nll {
    return err
}
```

// PrintVersionString

```go
printCliVersion();
if err := printCartridgeVersion(); err != nil {
    if projectPathIsSet {
       return err
    }

    log.Warnf("Failed to show Cartridge version: %s", err)

    // possibly here we can return nil here, because getting rocks version will fail
}

if err := printRocksVersions(); err != nil {
    if showRocksVersion {
       return err
    }

   log.Warnf("Failed to show rocks versions: %s", err)

}
```

cli/version/version.go
```diff
@@ -17,15 +20,100 @@ var (
...
+
+	return versionMsg
+}
+
+func getRocksVersions(projectPath string) (map[string]string, error) {
+	var rocksVersionsMap map[string]string
+	var err error
+
+	if fileInfo, err := os.Stat(projectPath); os.IsNotExist(err) {
+		return nil, fmt.Errorf("Your project path is invalid")
```
dokshina, line 45: "Specified project path doesn't exist".

cli/version/version.go
```diff
@@ -17,15 +20,100 @@ var (
...
+	} else if err != nil {
+		return nil, fmt.Errorf("Failed to use specified project path: %s", err)
```
dokshina, line 47: "Impossible to use..."

cli/version/version.go
```diff
@@ -17,15 +20,100 @@ var (
...
+	} else if !fileInfo.IsDir() {
+		return nil, fmt.Errorf("%s is not a directory", projectPath)
```
dokshina, line 49: "Specified project path %s is not a directory"

cli/version/version.go
```diff
@@ -41,16 +129,48 @@ func BuildVersionString() string {
...
+	return formatVersion(cliVersionTmpl, map[string]string{
+		"Title":   cliVersionTitle,
+		"Version": version,
+		"OS":      runtime.GOOS,
+		"Arch":    runtime.GOARCH,
+		"Commit":  gitCommit,
+	})
+}
+
+func printCliVersion() {
```
dokshina, line 141: It's not necessary to separate this print into function.

cli/commands/version.go
```diff
@@ -1,22 +1,30 @@
...
 	var versionCmd = &cobra.Command{
 		Use:   "version",
-		Short: "Print the version number of Cartridge CLI",
-		Long:  `All software has versions. This is Cartridge CLI's`,
+		Short: "Show version information",
 		Args:  cobra.MaximumNArgs(0),
 		Run: func(cmd *cobra.Command, args []string) {
-			fmt.Println(version.BuildVersionString())
+			if err := version.PrintVersionString(projectPath, cmd.Flags().Changed("project-path"), showRocksVersion); err != nil {
+				os.Exit(1)
```
dokshina, line 22: `log.Fatalf(err.Error())` is used for other commands. So, the error shouldn't be logged in `PrintVersionString` function, it's simply returned and printed here.

cli/version/version.go
```diff
@@ -41,16 +129,48 @@ func BuildVersionString() string {
...
+	fmt.Println(BuildCliVersionString())
+}
 
-	osArchStr := fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH)
-	versionParts = append(versionParts, osArchStr)
+func PrintVersionString(projectPath string, projectPathIsSet bool, showRocksVersion bool) error {
```
dokshina, line 145: Let's move exposed functions to the file top.

cli/version/version.go
```diff
@@ -41,16 +129,48 @@ func BuildVersionString() string {
...
+	printCliVersion()
+	if err := printCartridgeVersion(projectPath); err != nil {
```
dokshina, line 147: I suggest modifying this error message in the case when `showRocksVersion` is set. User should see something like "Failed to get Cartridge and other rocks versions: %s".

cli/version/version.go
```diff
@@ -41,16 +44,133 @@ func BuildVersionString() string {
...
+
+var (
+	cliVersionTmpl = `{{ .Title }}
+ Version:	{{ .Version }}
+ OS/Arch: 	{{ .OS }}/{{ .Arch }}
+ Git commit:	{{ .Commit }}
+`
+
+	cartridgeVersionTmpl = `{{ .Title }}
+ Version:	{{ .Version }}`
```
dokshina, line 175: Let's add here `\n` and remove it from `versionParts = append(versionParts, "\nRocks")`.

mRrvz: We will get a empty line (after all prints) in case if --rocks flag not specified, but, if ``--rocks`` flag specified, we don't get empty line. Is it okay?

test/integration/cli/test_version.py
```diff
@@ -1,8 +1,109 @@
...
+    cmd = [
+        cartridge_cmd,
+        "build",
+        project.path
+    ]
+
+    process = subprocess.run(cmd, cwd=tmpdir)
+    assert process.returncode == 0
+
+    for version_cmd in ["version", "-v", "--version"]:
```
dokshina, line 26: You can simply parametrize this test for all cases.

test/integration/cli/test_version.py
```diff
@@ -1,8 +1,109 @@
...
+        cmd = [
+            cartridge_cmd, version_cmd,
+            f"--project-path={project.path}"
+        ]
+
+        rc, output = run_command_and_get_output(cmd)
+        assert rc == 0
+
+        assert 'Tarantool Cartridge CLI\n Version:\t2' in output
```
dokshina, line 35: We should also check that there is no other output (like rocks).

dokshina (approved):
Great!

PR #556, opened 2021-06-08
Pre and post install scripts for rpm and deb packages

### LINKED ISSUE #211: [2pt] Add ability to set pre and post scripts for rpm packages
This feature may be useful in case of creating a specific user on the target system.

AnaNek:
Add ability to set pre and post scripts for rpm and deb packages.
Closes #211


cli/commands/pack.go
```diff
@@ -15,9 +15,11 @@ import (
 )
 
 var (
-	packTypeArgs = []string{"tgz", "rpm", "deb", "docker"}
-	deps         = []string{}
-	depsFile     = ""
+	packTypeArgs    = []string{"tgz", "rpm", "deb", "docker"}
+	deps            = []string{}
+	depsFile        = ""
+	preInstallFile  = ""
```
dokshina, line 21: `preInstallFile` - > `preInstallScript`
`postInstallFile` - > `postInstallScript`

cli/commands/pack.go
```diff
@@ -52,6 +54,8 @@ func init() {
 
 	packCmd.Flags().StringSliceVar(&deps, "deps", []string{}, depsUsage)
 	packCmd.Flags().StringVar(&depsFile, "deps-file", "", depsFileUsage)
+	packCmd.Flags().StringVar(&preInstallFile, "pre-install", "", preInstUsage)
```
dokshina, line 57: Maybe we can use `--preinst` and `--postinst` names?

cli/commands/pack.go
```diff
@@ -134,6 +138,48 @@ func fillDependencies(ctx *context.Ctx) error {
 	return nil
 }
 
+func fillInstallScripts(ctx *context.Ctx) error {
```
dokshina, line 141: `fillPreAndPostInstallScripts`

cli/commands/usage.go
```diff
@@ -61,6 +61,11 @@ defaults to "TARANTOOL_SDK_PATH" env`
 	depsFileUsage = `Path to the file that contains dependencies
 for the RPM and DEB packages.
 `
+	preInstUsage = `Path to the file that contains pre-install
```
dokshina, line 64: ..that contains pre install script...

cli/common/utils.go
```diff
@@ -564,6 +564,36 @@ func ParseDependencies(rawDeps []string) (PackDependencies, error) {
 	return deps, nil
 }
 
+func isCommandAvailable(command string) bool {
+	commandName := strings.Split(command, " ")[0]
+
+	_, err := exec.LookPath(commandName)
```
dokshina, line 570: You are trying to check if a specified command exists on your local machine, but the script will be run on the remote host where the package is installed.

cli/rpm/header.go
```diff
@@ -79,6 +81,55 @@ func addDependenciesRPM(rpmHeader *rpmTagSetType, deps common.PackDependencies)
...
+	if err != nil {
+		return "", err
+	}
+
+	parsedPreInstScript, err := common.ParseUserInstallScript(preInst)
+	if err != nil {
+		return "", err
+	}
+
+	commands := struct {
```
dokshina, line 97: You can use a map with string keys here.

cli/project/files.go
```diff
@@ -367,12 +367,14 @@ const (
...
 /bin/sh -c 'chown tarantool:tarantool /var/run/tarantool 2>&1 || :'
+{{ .UserPreInst }}
 `
 
 	PostInstScriptContent = `
 /bin/sh -c 'chown -R root:root /usr/share/tarantool/{{ .Name }}'
 /bin/sh -c 'chown root:root /etc/systemd/system/{{ .Name }}.service'
 /bin/sh -c 'chown root:root /etc/systemd/system/{{ .Name }}@.service'
 /bin/sh -c 'chown root:root /usr/lib/tmpfiles.d/{{ .Name }}.conf'
+{{ .UserPostInst }}
```
dokshina, line 378: What about using simple strings concatenation?

cli/common/utils.go
```diff
@@ -564,6 +564,36 @@ func ParseDependencies(rawDeps []string) (PackDependencies, error) {
...
+
+	if err != nil {
+		return false
+	}
+	return true
+}
+
+func ParseUserInstallScript(instScript string) (string, error) {
```
dokshina, line 578: I understood your idea, but I also think that if a user wants to specify his own pre- or post-install script, he is clever enough to write it appropriately. I think we should simply add a note in the doc, that all executable's paths should be absolute, or `/bin/sh -c ""` should be used.

I mean, he should use one of the following

```
/bin/sh -c 'mkdir dir-path'
# or
/bin/mkdir dir-path
```

The idea of using `/bin/sh -c` is to delegate discovering executables to `sh`. It's used since on different systems it can be placed in different paths.

cli/commands/pack.go
```diff
@@ -134,6 +138,48 @@ func fillDependencies(ctx *context.Ctx) error {
 	return nil
 }
 
+func fillPreAndPostInstallScripts(ctx *context.Ctx) error {
+	if ctx.Pack.Type == pack.RpmType || ctx.Pack.Type == pack.DebType {
+		var err error
+
+		if preInstallScript != "" {
+			if _, err := os.Stat(preInstallScript); os.IsNotExist(err) {
```
dokshina, line 146: You ignore all other `Stat` errors. The common way is:

```go
if _, err := os.Stat(preInstallScript); os.IsNotExist(err) {
    return fmt.Errorf("Specified pre-install script %s doesn't exists", preInstallScript)
} else if err != nil {
    return fmt.Errorf("Impossible to use specified pre-install script %s: %s", preInstallScript, err)
}
```

cli/commands/pack.go
```diff
@@ -134,6 +138,48 @@ func fillDependencies(ctx *context.Ctx) error {
...
+				return fmt.Errorf("Invalid path to file with pre-install script: %s", err)
+			}
+
+			ctx.Pack.PreInstallScript, err = common.GetFileContent(preInstallScript)
+			if err != nil {
+				return fmt.Errorf("Failed to get file content: %s", err)
+			}
+		}
+
+		if postInstallScript != "" {
```
dokshina, line 156: You can get rid of this copy-paste.

What about 

```go
if err := fillScript(preInstallScript, &ctx.Pack.PreInstallScript); err != nil {
    return fmt.Errorf("Failed to use specified pre-build script: %s", err)
}

if err := fillScript(postInstallScript, &ctx.Pack.PostInstallScript); err != nil {
    return fmt.Errorf("Failed to use specified post-build script: %s", err)
}
```

?

cli/commands/pack.go
```diff
@@ -134,6 +138,48 @@ func fillDependencies(ctx *context.Ctx) error {
...
+
+		return nil
+	}
+
+	if preInstallScript != "" {
+		log.Warnf("You specified the --preinst flag, but you are not packaging RPM or DEB. "+
+			"Flag will be ignored")
+	}
+
+	if postInstallScript != "" {
```
dokshina, line 175: Seems that this copy-paste can be removed via a simple loop.

cli/commands/pack.go
```diff
@@ -173,6 +219,10 @@ func runPackCommand(cmd *cobra.Command, args []string) error {
 		return err
```
dokshina, line 219: Seems that this function can be a part of `pack.FillCtx`.

cli/rpm/const.go
```diff
@@ -48,6 +48,7 @@ const (
 	tagPayloadCompressor = 1125
 	tagPayloadFlags      = 1126
 	tagPrein             = 1023
+	tagPostin            = 1024
```
dokshina, line 51: Let's also add `tagPostinProg` equal to `tagPreinProg`.

cli/rpm/header.go
```diff
@@ -79,6 +81,33 @@ func addDependenciesRPM(rpmHeader *rpmTagSetType, deps common.PackDependencies)
...
+func addPreAndPostInstallScriptsRPM(rpmHeader *rpmTagSetType, preInst string, postInst string) error {
+	var commandsPreInst bytes.Buffer
+
+	preInstTemplate, err := template.New("preInst").Parse(project.PreInstScriptContent)
+	if err != nil {
+		return err
+	}
+
+	commands := map[string]interface{}{
+		"UserPreInst": preInst,
```
dokshina, line 93: I suggested using string concatenations here. 

`commandsPreInst = strings.Join([]string{project.PreInstScriptContent, preInst}, "\n")`.

dokshina (commented):
Imagine that you are using custom pre- or post-script for your project. I think that you don't like to specify `--preinst` or `--postinst` options on each `cartridge pack`. I sure that we can check some default file names, that are used (if exists in the project root). 

Look at the `ctx.Pack.DockerFrom` option. The only difference is that pre- and post-build scripts shouldn't be added to the project template.

test/integration/pack/test_pack.py
```diff
@@ -1247,3 +1248,149 @@ def test_broken_dependencies(cartridge_cmd, project_without_dependencies, pack_f
...
+    assert "Specified script not_exist_file.txt doesn't exists" in output
+
+
+@pytest.mark.parametrize('pack_format', ['docker', 'tgz'])
+def test_pre_and_post_install_scripts_not_rpm_deb(cartridge_cmd, project_without_dependencies, pack_format, tmpdir):
+    project = project_without_dependencies
+
+    pre_install_script = os.path.join(tmpdir, "pre.sh")
+    with open(pre_install_script, "w") as f:
+        f.write("/bin/sh -c 'touch $HOME/hello.txt'\n" +
```
mRrvz, line 1366: Let's use multiline strings:
```python
f.write("""
/bin/sh -c 'touch $HOME/hello.txt'
/bin/sh -c 'echo hello'
""")
```

test/utils.py
```diff
@@ -590,6 +590,58 @@ def assert_tarantool_dependency_rpm(filename):
...
+        user_install_scripts_keys = ['prein', 'postin']
+        for key in user_install_scripts_keys:
+            assert key in rpm.headers
+
+        preinst_script = rpm.headers['prein'].decode('ascii')
+        postinst_script = rpm.headers['postin'].decode('ascii')
+
+        with open(user_pre_install_script, "r") as pre_install_content:
+            lines = pre_install_content.read().splitlines()
+            for line in lines:
```
mRrvz, line 604:
```python
for line in pre_install_content:
```

test/utils.py
```diff
@@ -590,6 +590,58 @@ def assert_tarantool_dependency_rpm(filename):
...
+                    if not line.startswith('#!/bin/'):
+                        assert line in preinst_script
+
+        with open(os.path.join(control_dir, 'postinst')) as postinst_script_file:
+            postinst_script = postinst_script_file.read()
+            with open(user_post_install_script, "r") as post_install_content:
+                lines = post_install_content.read().splitlines()
+                for line in lines:
+                    if not line.startswith('#!/bin/'):
+                        assert line in postinst_script
```
mRrvz, line 642: maybe try to write this test via ```@pytest.mark.parametrize```? parametrize preinst and postinst params

cli/pack/pack.go
```diff
@@ -244,6 +251,54 @@ func setSDKPath(ctx *context.Ctx) error {
...
+	if filename == "" {
+		if _, err := os.Stat(defaultScriptFilePath); err == nil {
+			log.Debugf("Default script is used: %s", defaultScriptFilePath)
+			filename = defaultScriptFilePath
+		} else if !os.IsNotExist(err) {
+			return fmt.Errorf("Failed to use default script file: %s", err)
+		}
+	}
+
+	if filename != "" {
```
mRrvz, line 266: It seems this ``if`` necessary.

AnaNek: We need this `if` to check if the `filename` has changed to the default file name.

test/utils.py
AnaNek: There will be new supporting function to avoid copy-paste.

cli/pack/pack.go
mRrvz: Looks like at line 265 ``filename`` always not "".  I see code below and i think ``defaultScriptFilePath`` can't be a empty. Or I'm wrong?

AnaNek: `filename` is empty if user did not specify one, if there is file with default name (`preinst.sh` for pre-install, `postinst.sh` for post-install) and user did not specify some file name, then at line 260 `filename` sets to `defaultScriptFilePath`

mRrvz:
Please, add a CHANGELOG

cli/pack/pack.go
```diff
@@ -244,6 +251,54 @@ func setSDKPath(ctx *context.Ctx) error {
 	return nil
 }
 
+func fillScript(filename string, defaultScriptFilePath string, outputScript *string) error {
```
mRrvz, line 254: I don't gey why we have to return outputScript not explicitly, but with through a pointer

mRrvz (approved):

test/e2e/test_deb.py
```diff
@@ -18,6 +19,14 @@
 def deb_archive_with_cartridge(cartridge_cmd, tmpdir, project_with_cartridge, request):
     project = project_with_cartridge
 
+    pre_install_filepath = os.path.join(tmpdir, "pre.sh")
+    with open(pre_install_filepath, "w") as f:
+        f.write("/bin/sh -c 'touch $HOME/hello.txt'")
```
dokshina, line 24: What about 

```
f.write('''
/bin/sh -c 'touch $HOME/hello-bin-sh.txt'
/bin/touch $HOME/hello-absolute.txt
''')
```

?

test/e2e/test_deb.py
```diff
@@ -108,4 +119,10 @@ def test_deb(container_with_installed_deb, tmpdir):
     run_command_on_container(container, "stress")
     run_command_on_container(container, "neofetch")
 
+    output = check_contains_file(container, '$HOME/hello.txt')
```
dokshina, line 122: Let's fix `check_contains_file` to return `run_command_on_container(container, command) == 'true`

https://github.com/tarantool/cartridge-cli/blob/master/test/utils.py#L1069

dokshina (approved):
Great!

PR #562, opened 2021-06-22
Allow to specify env params (like `net_msg_max`) with systemd unit file

### LINKED ISSUE #551: [4pt] Set up the appropriate value for net_msg_max parameter for big clusters
During deployment at the 1k cluster, we've stuck into that problem.

The default limit for the `net_msg_max` parameter is 768. This is not enough for the cluster of 1024 nodes.
We need to dynamically set up an appropriate value during deployment.

For example, for a cluster of 1024 nodes, we need to have 1024 fibers for simultaneous network sessions for `stateboard` instance.

AnaNek:
Ability to specify in the `systemd-unit-params.yml` file arguments passing
by env with systemd unit file.
Closes #551

### DIFF
--- a/cli/commands/pack.go
+++ b/cli/commands/pack.go
@@ -5,7 +5,6 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"strconv"
 
 	"github.com/apex/log"
 	"github.com/spf13/cobra"
@@ -13,7 +12,6 @@ import (
 	"github.com/tarantool/cartridge-cli/cli/common"
 	"github.com/tarantool/cartridge-cli/cli/context"
 	"github.com/tarantool/cartridge-cli/cli/pack"
-	"github.com/tarantool/cartridge-cli/cli/project"
 )
 
 var (
@@ -57,7 +55,6 @@ func init() {
 	packCmd.Flags().StringVar(&ctx.Pack.PreInstallScriptFile, "preinst", "", preInstUsage)
 	packCmd.Flags().StringVar(&ctx.Pack.PostInstallScriptFile, "postinst", "", postInstUsage)
 	packCmd.Flags().StringVar(&ctx.Pack.SystemdUnitParamsPath, "unit-params-file", "", UnitParamsFileUsage)
-	packCmd.Flags().IntVar(&ctx.Pack.NetMsgMax, "net-msg-max", 0, netMsgMaxUsage)
 }
 
 func addTarantoolDepIfNeeded(ctx *context.Ctx) error {
@@ -166,10 +163,6 @@ func runPackCommand(cmd *cobra.Command, args []string) error {
 	ctx.Pack.Type = strings.ToLower(cmd.Flags().Arg(0))
 	ctx.Project.Path = cmd.Flags().Arg(1)
 	ctx.Cli.CartridgeTmpDir = os.Getenv(cartridgeTmpDirEnv)
-
-	if err := setDefaultValue(cmd.Flags(), "net-msg-max", strconv.Itoa(defaultNetMsgMax)); err != nil {
-		return project.InternalError("Failed to set default net_msg_max value: %s", err)
-	}
 
 	if err := pack.Validate(&ctx); err != nil {
 		return err
--- a/cli/commands/pack.go
+++ b/cli/commands/pack.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"strconv"
 
 	"github.com/apex/log"
 	"github.com/spf13/cobra"
@@ -12,6 +13,7 @@ import (
 	"github.com/tarantool/cartridge-cli/cli/common"
 	"github.com/tarantool/cartridge-cli/cli/context"
 	"github.com/tarantool/cartridge-cli/cli/pack"
+	"github.com/tarantool/cartridge-cli/cli/project"
 )
 
 var (
@@ -55,6 +57,7 @@ func init() {
 	packCmd.Flags().StringVar(&ctx.Pack.PreInstallScriptFile, "preinst", "", preInstUsage)
 	packCmd.Flags().StringVar(&ctx.Pack.PostInstallScriptFile, "postinst", "", postInstUsage)
 	packCmd.Flags().StringVar(&ctx.Pack.SystemdUnitParamsPath, "unit-params-file", "", UnitParamsFileUsage)
+	packCmd.Flags().IntVar(&ctx.Pack.NetMsgMax, "net-msg-max", 0, netMsgMaxUsage)
 }
 
 func addTarantoolDepIfNeeded(ctx *context.Ctx) error {
@@ -163,6 +166,10 @@ func runPackCommand(cmd *cobra.Command, args []string) error {
 	ctx.Pack.Type = strings.ToLower(cmd.Flags().Arg(0))
 	ctx.Project.Path = cmd.Flags().Arg(1)
 	ctx.Cli.CartridgeTmpDir = os.Getenv(cartridgeTmpDirEnv)
+
+	if err := setDefaultValue(cmd.Flags(), "net-msg-max", strconv.Itoa(defaultNetMsgMax)); err != nil {
+		return project.InternalError("Failed to set default net_msg_max value: %s", err)
+	}
 
 	if err := pack.Validate(&ctx); err != nil {
 		return err
--- a/cli/commands/usage.go
+++ b/cli/commands/usage.go
@@ -162,4 +162,7 @@ defaults to %s`, defaultStartTimeout.String())
 
 	logLinesUsage = fmt.Sprintf(`Count of last lines to output
 defaults to %d`, defaultLogLines)
+
+	netMsgMaxUsage = fmt.Sprintf(`Maximum count of messages the fibers handle
+defaults to %d`, defaultNetMsgMax)
 )
--- a/cli/commands/usage.go
+++ b/cli/commands/usage.go
@@ -157,15 +157,9 @@ to get the rocks versions`
 )
 
 var (
-	// START
 	timeoutUsage = fmt.Sprintf(`Time to wait for instance(s) start
 defaults to %s`, defaultStartTimeout.String())
 
-	// LOG
 	logLinesUsage = fmt.Sprintf(`Count of last lines to output
 defaults to %d`, defaultLogLines)
-
-	// PACK
-	netMsgMaxUsage = fmt.Sprintf(`Maximum count of messages the fibers handle
-defaults to %d`, defaultNetMsgMax)
 )
--- a/cli/commands/usage.go
+++ b/cli/commands/usage.go
@@ -157,12 +157,15 @@ to get the rocks versions`
 )
 
 var (
+	// START
 	timeoutUsage = fmt.Sprintf(`Time to wait for instance(s) start
 defaults to %s`, defaultStartTimeout.String())
 
+	// LOG
 	logLinesUsage = fmt.Sprintf(`Count of last lines to output
 defaults to %d`, defaultLogLines)
 
+	// PACK
 	netMsgMaxUsage = fmt.Sprintf(`Maximum count of messages the fibers handle
 defaults to %d`, defaultNetMsgMax)
 )
--- a/cli/pack/systemd_dir.go
+++ b/cli/pack/systemd_dir.go
@@ -340,7 +340,7 @@ func setTarantoolEnvValues(ctx *context.Ctx, systemdCtx *map[string]interface{},
 	return nil
 }
 
-func getUnitEnvStringValue(envArgs UnitEnvArgs, key string) (string, error){
+func getUnitEnvStringValue(envArgs UnitEnvArgs, key string) (string, error) {
 	if value, ok := envArgs[key]; ok {
 		result, ok := value.(string)
 		if !ok {
@@ -352,7 +352,7 @@ func getUnitEnvStringValue(envArgs UnitEnvArgs, key string) (string, error){
 	return "", nil
 }
 
-func getSpecifiedStringArg(defaultValue string, envArgs UnitEnvArgs, argName string) (string, error){
+func getSpecifiedStringArg(defaultValue string, envArgs UnitEnvArgs, argName string) (string, error) {
 	if arg, err := getUnitEnvStringValue(envArgs, argName); err != nil || arg != "" {
 		return arg, err
 	}
--- a/cli/pack/systemd_dir.go
+++ b/cli/pack/systemd_dir.go
@@ -243,6 +243,8 @@ func getSystemdCtx(ctx *context.Ctx, systemdUnitParams SystemdUnitParams) *map[s
 	systemdCtx["FdLimit"] = systemdUnitParams.FdLimit
 	systemdCtx["StateboardFdLimit"] = systemdUnitParams.StateboardFdLimit
 
+	systemdCtx["NetMsgMax"] = ctx.Pack.NetMsgMax
+
 	if ctx.Tarantool.TarantoolIsEnterprise {
 		systemdCtx["Tarantool"] = filepath.Join(ctx.Running.AppDir, "tarantool")
 	} else {
@@ -271,6 +273,7 @@ Environment=TARANTOOL_WORKDIR={{ .DefaultWorkDir }}
 Environment=TARANTOOL_CFG={{ .ConfPath }}
 Environment=TARANTOOL_PID_FILE={{ .DefaultPidFile }}
 Environment=TARANTOOL_CONSOLE_SOCK={{ .DefaultConsoleSock }}
+Environment=TARANTOOL_NET_MSG_MAX={{ .NetMsgMax }}
 
 LimitCORE=infinity
 # Disable OOM killer
@@ -306,6 +309,7 @@ Environment=TARANTOOL_CFG={{ .ConfPath }}
 Environment=TARANTOOL_PID_FILE={{ .InstancePidFile }}
 Environment=TARANTOOL_CONSOLE_SOCK={{ .InstanceConsoleSock }}
 Environment=TARANTOOL_INSTANCE_NAME=%i
+Environment=TARANTOOL_NET_MSG_MAX={{ .NetMsgMax }}
 
 LimitCORE=infinity
 # Disable OOM killer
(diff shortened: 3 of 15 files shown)


cli/pack/systemd_dir.go
```diff
@@ -161,6 +163,7 @@ Environment=TARANTOOL_WORKDIR={{ .DefaultWorkDir }}
 Environment=TARANTOOL_CFG={{ .ConfPath }}
 Environment=TARANTOOL_PID_FILE={{ .DefaultPidFile }}
 Environment=TARANTOOL_CONSOLE_SOCK={{ .DefaultConsoleSock }}
+Environment=TARANTOOL_NET_MSG_MAX={{ .NetMsgMax }}
```
olegrok, line 166:
```
tarantool> box.cfg{net_msg_max = -2}
...
2021-06-22 11:52:22.297 [39038] main/103/interactive F> can't initialize storage: Incorrect value for option 'net_msg_max': minimal value is 2
```

Are there any check to prevent such behaviour? In case if you write invalid value to unit file it won't be possible even start such service.

cli/commands/usage.go
```diff
@@ -161,4 +161,7 @@ defaults to %s`, defaultStartTimeout.String())
 
 	logLinesUsage = fmt.Sprintf(`Count of last lines to output
 defaults to %d`, defaultLogLines)
+
+	netMsgMaxUsage = fmt.Sprintf(`Maximum count of messages the fibers handle
```
mRrvz, line 165: It seems to me that this should be placed in the section with the ``pack`` usage commands

mRrvz: I was wrong, pack variables are constants and we can't place this string there. Anyway, I think we have to add comment for which ``cartridge`` command this string is used

cli/commands/pack.go
```diff
@@ -163,6 +166,10 @@ func runPackCommand(cmd *cobra.Command, args []string) error {
 	ctx.Project.Path = cmd.Flags().Arg(1)
 	ctx.Cli.CartridgeTmpDir = os.Getenv(cartridgeTmpDirEnv)
 
+	if err := setDefaultValue(cmd.Flags(), "net-msg-max", strconv.Itoa(defaultNetMsgMax)); err != nil {
+		return project.InternalError("Failed to set default net_msg_max value: %s", err)
```
mRrvz, line 170: Is this InternalError really needed here?

AnaNek: I did the same as `log` https://github.com/tarantool/cartridge-cli/blob/master/cli/commands/log.go#L53

dokshina: This is needed for logs to differ when 0 is specified by the user or nothing is specified. Seems that here you can do it simpler. Just use `0` default value for `net-msg-max` to indicate that this parameter shouldn't be set at all.

dokshina:
I'm looking at this patch and #566 and I don't like that we add a new flag for each option from systemd unit files.

I suggest changing the approach itself. `net_msg_max` is just one of the possible `box.cfg` or cartridge options that can be passed to the instance and parsed by `cartridge.argparse`. In fact, all that we do here is converting `--<param> <value>` to `Environment=TARANTOOL_<PARAM>: <value>` in the unit file. I think that such options can be placed in some file.

For example, if the user adds `systemd-unit-params.yml` file in the project root:

```
instance-args:
  net_msg_max: 1024
  pid_file: '/some/special/dir/{{ .Name }}.%i.pid'
stateboard-args:
  pid_file: '/some/special/dir/{{ .StateboardName }}.pid'
```

Note, that `pid_file` option is passed by default, but here we overwrite it. 
Moreover, the templated string is passed (see supported variables list [here](https://github.com/tarantool/cartridge-cli#package-details).

This leads to generating such unit files:

```
# myapp@.service
...
Environment=TARANTOOL_PID_FILE=/some/special/dir/myapp.%i.pid
# instead of TARANTOOL_PID_FILE=some/var/run/tarantool/myapp.%i.pid
Environment=TARANTOOL_NET_MSG_MAX=1024
...
```

```
# myapp@.service
...
Environment=TARANTOOL_PID_FILE=/some/special/dir/myapp-stateboard.pid
...
```

olegrok:
> @olegrok, what do you think?

I like an approach with file. But I don't understand how user should specify template parameters (e.g. `/some/special/dir/{{ .Name }}.%i.pid`). Is it some known convention or you just plan to use default Go templates?

cli/pack/systemd_dir.go
```diff
@@ -217,7 +226,90 @@ func getSystemdTemplate(ctx *context.Ctx) (templates.Template, error) {
...
+		"TARANTOOL_APP_NAME":     ctx.Project.StateboardName,
+		"TARANTOOL_WORKDIR":      project.GetStateboardWorkDir(ctx),
+		"TARANTOOL_CFG":          ctx.Running.ConfPath,
+		"TARANTOOL_PID_FILE":     project.GetStateboardPidFile(ctx),
+		"TARANTOOL_CONSOLE_SOCK": project.GetStateboardConsoleSock(ctx),
+		"TARANTOOL_NET_MSG_MAX":  defaultNetMsgMax,
+	}
+}
+
+func setUnitEnv(unitEnv *map[string]interface{}, envParams map[string]interface{}) error {
```
dokshina, line 258: What about making the function name more meaningful? Like `updateUnitEnvBySpecifiedArgs`?

dokshina: BTW, maybe declare a `type UnitEnvArgs map[string]interface{}`?

cli/pack/systemd_dir.go
```diff
@@ -217,7 +226,90 @@ func getSystemdTemplate(ctx *context.Ctx) (templates.Template, error) {
...
+			return fmt.Errorf("net_msg_max parameter type should be integer")
+		}
+
+		if err := checkMinValue("net_msg_max", net_msg_max, minNetMsgMax); err != nil {
+			return err
+		}
+	}
+
+	for key, value := range envParams {
+		(*unitEnv)[fmt.Sprintf("TARANTOOL_%s", strings.ToUpper(key))] = value
```
dokshina, line 271: Let's compute a map key on a separated string. It will make the code more readable. 
Moreover, I think that we can transform `long-param` into `TARANTOOL_LONG_PARAM`, like `cartridge.argparse` do. I mean replacing `"-"` by `"_"`.

cli/pack/systemd_dir.go
```diff
@@ -217,7 +226,90 @@ func getSystemdTemplate(ctx *context.Ctx) (templates.Template, error) {
...
+}
+
+func setTarantoolEnvValues(ctx *context.Ctx, systemdCtx *map[string]interface{}, systemdUnitParams SystemdUnitParams) error {
+	var ok bool
+
+	setDefaultTarantoolEnvValues(ctx, systemdCtx)
+
+	unitEnv, ok := (*systemdCtx)["UnitEnv"].(map[string]interface{})
+	if !ok {
+		return fmt.Errorf("Setting env values :cannot convert (type interface {}) to type map[string]interface{}")
```
dokshina, line 284: " :cannot" looks like a typo.

cli/pack/systemd_dir.go
```diff
@@ -217,7 +226,90 @@ func getSystemdTemplate(ctx *context.Ctx) (templates.Template, error) {
...
+	return nil
+}
+
+func setTarantoolEnvValues(ctx *context.Ctx, systemdCtx *map[string]interface{}, systemdUnitParams SystemdUnitParams) error {
+	var ok bool
+
+	setDefaultTarantoolEnvValues(ctx, systemdCtx)
+
+	unitEnv, ok := (*systemdCtx)["UnitEnv"].(map[string]interface{})
+	if !ok {
```
dokshina, line 283: It's an internal error.

cli/pack/systemd_dir.go
```diff
@@ -217,7 +226,90 @@ func getSystemdTemplate(ctx *context.Ctx) (templates.Template, error) {
...
+	instUnitEnv, ok := (*systemdCtx)["InstUnitEnv"].(map[string]interface{})
+	if !ok {
+		return fmt.Errorf("Setting env values :cannot convert (type interface {}) to type map[string]interface{}")
+	}
+
+	if err := setUnitEnv(&instUnitEnv, systemdUnitParams.InstanceEnv); err != nil {
+		return err
+	}
+
+	stateboardUnitEnv, ok := (*systemdCtx)["StateboardUnitEnv"].(map[string]interface{})
```
dokshina, line 300: This copy-paste can be moved to some function =)

dokshina (commented):
Good!

Let's add unit tests for systemd unit files creation. Like [this tests](https://github.com/tarantool/cartridge-cli/blob/master/cli/project/dockerfiles_test.go).

test/project.py
```diff
@@ -480,6 +480,67 @@ def patch_init_to_log_lines(project, lines):
...
+    {{
+        path = '/metrics',
+        format = 'prometheus'
+    }},
+    {{
+        path = '/health',
+        format = 'health'
+    }}
+}})
+'''
```
mRrvz, line 533: Let's move this to separate file in the ``test/files`` directory.

dokshina (commented):
Please, add documentation for this feature.
Moreover, would it work if `arg: 'bla-bla-bla-{{ .Name }}'` specified? Let's add a test. I would like you to add unit tests for all such cases.

cli/pack/systemd_dir.go
```diff
@@ -217,25 +228,122 @@ func getSystemdTemplate(ctx *context.Ctx) (templates.Template, error) {
...
-	systemdCtx["DefaultConsoleSock"] = project.GetInstanceConsoleSock(ctx, "default")
-	systemdCtx["InstanceConsoleSock"] = project.GetInstanceConsoleSock(ctx, instanceNameSpecifier)
-	systemdCtx["StateboardConsoleSock"] = project.GetStateboardConsoleSock(ctx)
+	systemdCtx["InstanceWorkDir"], err = getSpecifiedStringArg(project.GetInstanceWorkDir(ctx, instanceNameSpecifier), (*systemdUnitParams).InstanceEnv, "workdir")
+	if err != nil {
+		return nil, err
+	}
 
-	systemdCtx["ConfPath"] = ctx.Running.ConfPath
+	systemdCtx["StateboardWorkDir"], err = getSpecifiedStringArg(project.GetStateboardWorkDir(ctx), (*systemdUnitParams).StateboardEnv, "workdir")
```
dokshina, line 343: What about reducing copy-paste?

This snippet has more lines, but it's easier to read and understand:

```go
systemdCtxParams := []systemdCtxParam{
    {
        ArgName: "app-name",
        CtxKey: "Name",
        DefaultValue: ctx.Project.Name,
        EnvArgs: (*systemdUnitParams).InstanceEnv,
    },
   ...
}

for _, systemdCtxParam := range systemdCtxParams {
    systemdCtx[systemdCtxParam.CtxKey], err = getSpecifiedStringArg(...)
	if err != nil {
		return nil, err
	}
}
```

cli/pack/systemd_dir_test.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+	defer os.RemoveAll(tmpDir)
+
+	err = initSystemdDir(tmpDir, &ctx)
+	assert.Nil(err)
+
+	// read systemd directory content
+	systemdDirPath := filepath.Join(tmpDir, "/etc/systemd/system/")
+
+	for filename, expContent := range expContentByFilename {
```
dokshina, line 353: Seems like a function.

cli/pack/systemd_dir_test.go
```diff
@@ -0,0 +1,381 @@
...
+	t.Parallel()
+	assert := assert.New(t)
+
+	var err error
+	var ctx context.Ctx
+
+	newAppName := "new-name"
+	newStateboardAppName := "new-stateboard-name"
+
+	systemdUnitParamsContent := fmt.Sprintf(`fd-limit: 1024
```
dokshina, line 188: Just declare a map[string]interface and dump it into file as a YAML.

test/e2e/test_deb.py
```diff
@@ -39,12 +41,24 @@ def deb_archive_with_cartridge(cartridge_cmd, tmpdir, project_with_cartridge, re
...
+    net_msg_max = 1024
+
+    systemd_unit_params = os.path.join(tmpdir, "systemd-unit-params.yml")
+    with open(systemd_unit_params, "w") as f:
+        f.write(f"""
+                instance-env:
+                    net-msg-max: {net_msg_max}
+                """)
+
+    patch_init_to_net_msg_max(project, net_msg_max)
```
dokshina, line 53: I suggest testing not only `net_msg_max`, but also custom params like `some-user-param`. And params for stateboard.

test/integration/pack/test_pack.py
```diff
@@ -1738,3 +1742,95 @@ def test_rocks_noncache_flag(cartridge_cmd, light_project, tmpdir, pack_format):
...
+    project = project_without_dependencies
+
+    invalid_net_msg_max = -1
+
+    systemd_unit_params = os.path.join(tmpdir, "systemd-unit-params.yml")
+    with open(systemd_unit_params, "w") as f:
+        f.write(f"""
+                instance-env:
+                    net-msg-max: {invalid_net_msg_max}
+                """)
```
mRrvz, line 1821: Let's use ``yaml.dump``

cli/pack/systemd_dir.go
```diff
@@ -217,25 +237,147 @@ func getSystemdTemplate(ctx *context.Ctx) (templates.Template, error) {
...
+}
 
-	systemdCtx["ConfPath"] = ctx.Running.ConfPath
+func updateUnitEnvBySpecifiedArgs(unitEnv interface{}, envParams UnitEnvArgs) error {
+	mapUnitEnv, ok := unitEnv.(map[string]interface{})
+	if !ok {
+		return project.InternalError("Setting env values: can't convert (type interface {}) to type map[string]interface{}")
+	}
+
+	if value, ok := envParams["net-msg-max"]; ok {
```
dokshina, line 286: What if `TARANTOOL_NET_MSG_MAX` is specified?

test/e2e/test_deb.py
```diff
@@ -39,12 +41,33 @@ def deb_archive_with_cartridge(cartridge_cmd, tmpdir, project_with_cartridge, re
...
+                """)
+
+    expected_param_values = f"""
+                            {{
+                            net_msg_max="{net_msg_max}",
+                            my_param="{my_param}"
+                            }}
+                            """
+
+    patch_init_to_cartridge_argparse_params(project, expected_param_values)
```
dokshina, line 62: `patch_init_to_cartridge_argparse_params` -> `patch_init_to_check_passed_params`.

It's OK to hardcode params directly in the init file for this test. It's simpler and easier to understand.

test/files/init_check_passed_params.lua
```diff
@@ -0,0 +1,57 @@
...
+    package.cpath = app_dir .. '/?.dylib;' .. package.cpath
+    package.cpath = app_dir .. '/.rocks/lib/tarantool/?.so;' .. package.cpath
+    package.cpath = app_dir .. '/.rocks/lib/tarantool/?.dylib;' .. package.cpath
+end
+
+local argparse = require('cartridge.argparse')
+
+local actual_param_value = argparse.get_box_opts().net_msg_max
+assert(actual_param_value == 1024,
+        string.format('Mismatch of net_msg_max: %s != %s', actual_param_value, 1024))
```
dokshina, line 25:
```lua
assert(
    actual_param_value == 1024,
    string.format('Mismatch of net_msg_max: %s != %s', actual_param_value, 1024)
)
```

test/files/init_check_passed_params.lua
```diff
@@ -0,0 +1,57 @@
...
+
+actual_param_value = argparse.parse().user_param
+assert(actual_param_value == 'user_data',
+        string.format('Mismatch of user_param: %s != %s', actual_param_value, 'user_data'))
```
dokshina, line 29: The same.

test/integration/pack/test_pack.py
```diff
@@ -1475,7 +1475,8 @@ def test_fd_limit_default_file(cartridge_cmd, project_without_dependencies, pack
     process = subprocess.run(cmd, cwd=tmpdir)
     assert process.returncode == 0
 
-    check_fd_limits_in_unit_files(fd_limit, stateboard_fd_limit, project.name, pack_format, tmpdir)
+    check_param_in_unit_files(fd_limit, stateboard_fd_limit,
+                              "LimitNOFILE", project.name, pack_format, tmpdir)
```
dokshina, line 1479: Up to you, bit I like other style

```python
check_param_in_unit_files(
    fd_limit, stateboard_fd_limit,
    "LimitNOFILE", project.name, pack_format, tmpdir
)
```

test/integration/pack/test_pack.py
```diff
@@ -2010,3 +2014,159 @@ def test_path_is_not_directory(cartridge_cmd, light_project, tmpdir, pack_format
...
+    rc, output = run_command_and_get_output(cmd, cwd=tmpdir)
+    assert rc == 0
+
+    check_param_in_unit_files(instance_net_msg_max, stateboard_net_msg_max,
+                              "Environment=TARANTOOL_NET_MSG_MAX",
+                              project.name, pack_format, tmpdir)
+
+    # parameter specified in TARANTOOL_NET_MSG_MAX format
+    with open(systemd_unit_params, "w") as f:
+        yaml.dump({"instance-env":
```
dokshina, line 2054: Please, don't mix dictionary levels:

```python
yaml.dump({
  "instance-env": {"TARANTOOL_NET_MSG_MAX": instance_net_msg_max},
   "stateboard-env": {"TARANTOOL_NET_MSG_MAX": stateboard_net_msg_max}
}, f)

# or

yaml.dump({
   "instance-env": {
        "TARANTOOL_NET_MSG_MAX": instance_net_msg_max
    },
    "stateboard-env": {
        "TARANTOOL_NET_MSG_MAX": stateboard_net_msg_max
    }
}, f)
```

test/integration/pack/test_pack.py
```diff
@@ -2010,3 +2014,159 @@ def test_path_is_not_directory(cartridge_cmd, light_project, tmpdir, pack_format
...
+def test_net_msg_max_invalid_type(cartridge_cmd, project_without_dependencies, pack_format, tmpdir):
+    project = project_without_dependencies
+
+    invalid_net_msg_max = "string_value"
+
+    systemd_unit_params = os.path.join(tmpdir, "systemd-unit-params.yml")
+
+    # parameter specified in net-msg-max format
+    with open(systemd_unit_params, "w") as f:
+        yaml.dump({"instance-env":
```
dokshina, line 2087: The same as above, here and below.

dokshina (commented):
Good patch! Generally, I have only some code-style comments. Please, fix them before merging.

dokshina (approved):

PR #568, opened 2021-06-25
Cache .rocks produced on cartridge pack

### LINKED ISSUE #559: [1pt] Cache .rocks produced on `cartridge pack`

mRrvz:
Project rocks caching when using ``cartridge pack`` 
command. Closes #559

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   ``cartridge pack``), using the ``--preinst`` and ``--postinst`` flags.
 - ``cartridge pack`` generates ``VERSION.lua`` file with the current
   version of project.
+- Project rocks caching when using ``cartridge pack`` command.
 - Ability to specify fd limit in the systemd unit template
   (command ``cartridge pack``) in the ``systemd-unit-params.yml`` file.
 - ``cartridge pack`` now uses the VERSION file from the
--- a/cli/pack/app_dir.go
+++ b/cli/pack/app_dir.go
@@ -184,7 +184,11 @@ func updateRocksCache(ctx *context.Ctx, cachePaths *cacheProjectPaths) error {
 	}
 
 	log.Debugf("Rocks cache has been successfully saved in: %s", cachePaths.rocksPath)
-	return removeRocksByEditTime(ctx)
+	if err := removeRocksByEditTime(ctx); err != nil {
+		log.Warnf("Failed to clear the directory with rocks: %s", err)
+	}
+
+	return nil
 }
 
 func copyProjectFiles(dst string, ctx *context.Ctx) error {
--- a/cli/pack/app_dir.go
+++ b/cli/pack/app_dir.go
@@ -67,7 +67,7 @@ func initAppDir(appDirPath string, ctx *context.Ctx) error {
 	if !ctx.Pack.NoCache {
 		if _, err := os.Stat(cachePaths.rocksPath); err == nil {
 			// If rocks found in cache - we just copy them.
-			if err := copyFromCache(cachePaths.rocksPath, filepath.Join(appDirPath, ".rocks")); err != nil {
+			if err := copyPathFromCache(cachePaths.rocksPath, filepath.Join(appDirPath, ".rocks")); err != nil {
 				return err
 			}
 		} else if !os.IsNotExist(err) {
@@ -82,7 +82,7 @@ func initAppDir(appDirPath string, ctx *context.Ctx) error {
 	}
 
 	// Update rocks cache in cartridge temp directory
-	if err := updateCache(ctx, cachePaths, ".rocks"); err != nil {
+	if err := updateCachePath(ctx, cachePaths, ".rocks"); err != nil {
 		return err
 	}
 
@@ -112,12 +112,12 @@ func initAppDir(appDirPath string, ctx *context.Ctx) error {
 	return nil
 }
 
-func copyFromCache(cachedRocksDir string, appDirPath string) error {
-	log.Infof("Adding cached module %s", filepath.Base(appDirPath))
-	err := copy.Copy(cachedRocksDir, appDirPath)
+func copyPathFromCache(cachedPath string, destPath string) error {
+	log.Infof("Using cached path %s", filepath.Base(destPath))
+	err := copy.Copy(cachedPath, destPath)
 
 	if err != nil {
-		return fmt.Errorf("Failed to copy .rocks from cache to project directory: %s", err)
+		return fmt.Errorf("Failed to copy path %s from cache to project directory: %s", destPath, err)
 	}
 
 	return nil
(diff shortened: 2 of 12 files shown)


cli/pack/app_dir.go
```diff
@@ -81,6 +108,86 @@ func initAppDir(appDirPath string, ctx *context.Ctx) error {
...
+	}
+
+	rockspecPath, err := common.FindRockspec(ctx.Project.Path)
+	if err != nil {
+		return nil, fmt.Errorf("Unable to find rockspec: %s", err)
+	} else if rockspecPath == "" {
+		return nil, fmt.Errorf("Application directory should contain rockspec")
+	}
+
+	projectPathHash := common.StringSHA256Hex(ctx.Project.Path)
```
dokshina, line 134: Let's use the first 10-15 digits, I think it's enough.

dokshina (commented):
Great!

But what about caching not only `.rocks` but also `node_modules`? It can be implemented in a separated patch. The general idea is to cache different paths that can be specified by the user. The question is a cache key. Maybe it can be like this

```
.cartridge/tmp/cache/
  <project-hash>/
    .rocks/<rockspec-hash>
    .node_modules/<smth-other-hash>
```

It can be configured like this

```yaml
'.rocks':
  key-path: 'myapp-scm-1.rockspec'
node_modules:
  no-key: true. # cache always
```

cli/pack/app_dir.go
```diff
@@ -81,6 +110,83 @@ func initAppDir(appDirPath string, ctx *context.Ctx) error {
...
+			return fmt.Errorf("Failed to clear rocks cache directory: %s", err)
+		}
+	}
+
+	if err := copy.Copy(filepath.Join(ctx.Build.Dir, ".rocks"), filepath.Join(cachePaths.rocksPath)); err != nil {
+		return fmt.Errorf("Failed to copy: %s", err)
+	}
+
+	log.Debugf("Rocks cache has been successfully saved in: %s", cachePaths.rocksPath)
+	return removeRocksByEditTime(ctx)
```
dokshina, line 187: It shouldn't be an error, just Warn.

test/utils.py
```diff
@@ -1317,3 +1318,13 @@ def get_admin_connection_params(connection_type, project):
         ]
 
     assert False, "Unknown connection type: %s" % connection_type
+
+
+def get_rocks_cache_path():
+    return os.path.join(os.getenv("HOME"), ".cartridge", "tmp", "cache")
```
dokshina, line 1324: You clear all local cache this way. Let's clear cache by project path.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   ``cartridge pack``), using the ``--preinst`` and ``--postinst`` flags.
 - ``cartridge pack`` generates ``VERSION.lua`` file with the current
   version of project.
+- Project rocks caching when using ``cartridge pack`` command.
```
dokshina, line 16: Caching project rocks between different builds on `cartridge pack`?

cli/pack/tmp_dir.go
```diff
@@ -16,6 +16,7 @@ const (
 	tmpPackDirNameFmt   = "pack-%s"
 	defaultBuildDirName = "cartridge.tmp"
 	packageFilesDirName = "package-files"
+	cacheDirName        = "cache"
```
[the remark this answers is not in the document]
mRrvz: It looks like a separate PR.

cli/pack/app_dir.go
```diff
@@ -110,9 +112,9 @@ func initAppDir(appDirPath string, ctx *context.Ctx) error {
 	return nil
 }
 
-func copyRocksFromCache(cachedRocksDir string, appDirPath string) error {
-	log.Info("Adding cached rocks")
-	err := copy.Copy(cachedRocksDir, filepath.Join(appDirPath, ".rocks"))
+func copyFromCache(cachedRocksDir string, appDirPath string) error {
+	log.Infof("Adding cached module %s", filepath.Base(appDirPath))
```
dokshina, line 116: `appDirPath` - `appDir` prefix confuses, app-dir is the directory in the resulting package. 
I suggest renaming:
* `copyFromCache` -> `copyPathFromCache`
* `cachedRocksDir` -> `cachedPath`
* `appDirPath` -> `destPath`

`"Adding cached module %s` -> `"Using cached path %s`

cli/pack/app_dir.go
```diff
@@ -168,7 +170,7 @@ func removeRocksByEditTime(ctx *context.Ctx) error {
 	return nil
 }
 
-func updateRocksCache(ctx *context.Ctx, cachePaths *cacheProjectPaths) error {
+func updateCache(ctx *context.Ctx, cachePaths *cacheProjectPaths, moduleName string) error {
```
dokshina, line 173: What is "module" in `moduleName`? Seems that it's just a `path`.

cli/pack/app_dir.go
```diff
@@ -82,7 +82,7 @@ func initAppDir(appDirPath string, ctx *context.Ctx) error {
 	}
 
 	// Update rocks cache in cartridge temp directory
```
dokshina, line 84: Remove `rocks` word.

cli/pack/app_dir.go
```diff
@@ -49,12 +59,33 @@ func initAppDir(appDirPath string, ctx *context.Ctx) error {
...
+
 	ctx.Build.Dir = appDirPath
+	// Build project
 	if err := build.Run(ctx); err != nil {
 		return err
 	}
 
+	// Update rocks cache in cartridge temp directory
+	if err := updateCachePath(ctx, cachePaths, ".rocks"); err != nil {
+		return err
```
dokshina, line 86: Are you sure that we should return an error here? Seems that packing shouldn't fail if caching failed?

cli/pack/app_dir.go
```diff
@@ -49,12 +59,33 @@ func initAppDir(appDirPath string, ctx *context.Ctx) error {
...
+	}
+
 	ctx.Build.Dir = appDirPath
+	// Build project
 	if err := build.Run(ctx); err != nil {
 		return err
 	}
 
+	// Update rocks cache in cartridge temp directory
+	if err := updateCachePath(ctx, cachePaths, ".rocks"); err != nil {
```
dokshina, line 85: Maybe better `updateCache`? `updateCachePath` seems like you updates path to some cache.

dokshina, line 85: Maybe better `updateCache`? `updateCachePath` seems like you update path to some cache.

cli/pack/app_dir.go
```diff
@@ -81,6 +112,87 @@ func initAppDir(appDirPath string, ctx *context.Ctx) error {
...
+}
+
+func rotateCacheDirs(ctx *context.Ctx) error {
+	dir, err := ioutil.ReadDir(ctx.Cli.CacheDir)
+	if err != nil {
+		return err
+	}
+
+	if len(dir) > maxCachedProjects {
+		// Sort project rocks by change time
```
dokshina, line 158: `rocks` again.

cli/pack/tmp_dir.go
```diff
@@ -33,6 +34,8 @@ func init() {
 
 // tmp directory structure:
 // ~/.cartridge/tmp/            <- ctx.Cli.CartridgeTmpDir (can be changed by CARTRIDGE_TEMPDIR)
+//   cache/                     <- ctx.Cli.CacheDir (used for saving cache modules when packing application)
+//     <project-hash>/			<- Directory containing cached project modules this hash
```
dokshina, line 38: ...cached project paths?

cli/pack/app_dir.go
```diff
@@ -81,6 +112,83 @@ func initAppDir(appDirPath string, ctx *context.Ctx) error {
...
+	return nil
+}
+
+func updateCache(ctx *context.Ctx, cachePaths *cacheProjectPaths, path string) error {
+	if ctx.Pack.NoCache {
+		return nil
+	}
+
+	if _, err := os.Stat(cachePaths.projectPath); err == nil {
+		if err := common.ClearDir(cachePaths.projectPath); err != nil {
```
dokshina, line 179: Why do you clear all project caches when only `cachePath` should be updated?

cli/pack/app_dir.go
```diff
@@ -49,12 +59,33 @@ func initAppDir(appDirPath string, ctx *context.Ctx) error {
...
 	}
 
-	// build
+	cachePaths, err := getProjectCachePaths(ctx)
+	if err != nil {
+		log.Warnf("%s", err)
+	}
+
+	if !ctx.Pack.NoCache {
+		if _, err := os.Stat(cachePaths.cachePath); err == nil {
```
dokshina, line 68: What about some function like `copyFromCache(cachePath, ctx)`?

cli/pack/app_dir.go
dokshina: BTW, I'm not sure that you need to store `.projectPath`. It's used only when paths to cached rocks are computed. Then, you work with each cache separately. Maybe `cachePaths` can be a dictionary?

```go
cachePaths = map[string]string{
    ".rocks": "<path-computed-using-projectPath-and-key>"
}
```

cli/pack/app_dir.go
```diff
@@ -81,6 +97,104 @@ func initAppDir(appDirPath string, ctx *context.Ctx) error {
...
+			projectCachePath := filepath.Join(ctx.Cli.CacheDir, dir[i].Name())
+			log.Debugf("Removing project cache directory: %s", projectCachePath)
+			os.RemoveAll(projectCachePath)
+		}
+	}
+
+	return nil
+}
+
+func updateCache(cachePaths map[string]string, ctx *context.Ctx) error {
```
dokshina, line 173: Maybe use `type CachePaths map[string]string` declaration?

cli/pack/app_dir.go
```diff
@@ -81,6 +97,104 @@ func initAppDir(appDirPath string, ctx *context.Ctx) error {
...
+	if ctx.Pack.NoCache {
+		return nil
+	}
+
+	for path, cacheDir := range cachePaths {
+		// Delete other caches for this path,
+		// because we only store 1 cache for the path
+		currentPath := filepath.Dir(cacheDir)
+		if _, err := os.Stat(currentPath); err == nil {
```
dokshina, line 182: What if `err != nil && !os.IsNotExists(err)`?

dokshina (approved):
Great!

PR #584, opened 2021-07-06
Fix cleanup translation files

artembo:
Translated files downloaded from Crowdin with root owner
which the current user can not process. Change user for
translations before process it by cleanup script.


rosik (approved):

rosik:
Related to https://github.com/tarantool/cartridge/pull/1459

PR #586, opened 2021-07-07
Implement caching of any path when packaging the application

### LINKED ISSUE #578: [2pt] Implement caching of any modules in ``cartridge pack`` command
See https://github.com/tarantool/cartridge-cli/pull/568#pullrequestreview-695972127

mRrvz:
Added ability to caching any paths specified in ``pack-cache-config.yml`` file
when packaging application via ``cartridge pack`` command. Closes #578


README.rst
```diff
@@ -875,6 +878,51 @@ These options are supported now:
...
+.. code-block:: yaml
+
+    '.rocks':
+      key-path: myapp-scm-1.rockspec
+
+    'node_modules':
+      always-cache: true
+
+    'third_party/custom_module':
+      key: simple-hash-key
```
dokshina, line 896: I was thinking about this option. Is there some way to use this option for a real project? Only if we want manually update the cache key?

cli/pack/app_dir.go
```diff
@@ -23,11 +24,21 @@ const (
 	versionFileName    = "VERSION"
 	versionLuaFileName = "VERSION.lua"
 
-	maxCachedProjects = 5
+	cacheParamsFileName     = "pack-cache.yml"
```
dokshina, line 27: `pack-cache-config.yml`?

README.rst
mRrvz: Yes, it seems only in this case. But I thnik it might be useful

test/integration/pack/test_pack.py
```diff
@@ -1732,9 +1733,268 @@ def test_rocks_noncache_flag(cartridge_cmd, light_project, tmpdir, pack_format):
...
+    with open(os.path.join(first_cached_dir_path, "fst.txt"), "w") as f:
+        f.write("Dummy text 123")
+
+    with open(os.path.join(second_cached_dir_path, "snd.txt"), "w") as f:
+        f.write("321 txet ymmuD")
+
+    new_cache_yml_path = os.path.join(tmpdir, "new-pack-cache.yml")
+    with open(new_cache_yml_path, "w") as f:
+        yaml.dump({
+            ".rocks": {"always-cache": True},
```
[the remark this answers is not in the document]
mRrvz: Why? At ``instances.yml`` and ``replicasets.yml`` we use the same scheme: 
* for ``instances.ym``l key is instance name;
* for ``replicasets.yml`` key also is instance.

And we don't use ``instance-name`` or something like this field.

README.rst
```diff
@@ -916,12 +916,9 @@ cached paths. If a 6th project appears, we delete the oldest existing project
...
 cache as many paths as you like for one project.
 
-You can always disable caching by using the ``--no-cache`` flag or by
-removing paths from the ``pack-cache.yml`` file. If you completely delete
-``pack-cache.yml`` file and start packing without the ``--no-cache flag``,
-you will get an error.
-
-To completely reset the cache, delete ``~/.cartridge/tmp/cache`` directory.
+You can always disable caching by using the ``--no-cache`` flag or by removing
+paths from the ``pack-cache-config.yml`` file.  To completely reset the cache,
```
dokshina, line 920: Two spaces detected =)

cli/pack/app_dir.go
```diff
@@ -190,6 +197,49 @@ func parseCacheParamsFile(cacheParamsPath string) (*CachePathsParams, error) {
...
+	projectPathHash := common.StringSHA1Hex(ctx.Project.Path)[:cntFirstSymbolsFromHash]
+
+	switch {
+	case params.AlwaysCache == true:
+		keyHash = "always"
+	case params.KeyPath != "":
+		pathFromProjectRoot := filepath.Join(ctx.Project.Path, params.KeyPath)
+		if _, err := os.Stat(pathFromProjectRoot); err == nil {
+			if keyHash, err = common.FileSHA1Hex(pathFromProjectRoot); err != nil {
+				return "", fmt.Errorf("Failetd to get hash from file content for path %s: %s", path, err)
```
dokshina, line 211: Failetd -> Failed

cli/pack/app_dir.go
```diff
@@ -24,9 +24,11 @@ const (
 	versionFileName    = "VERSION"
 	versionLuaFileName = "VERSION.lua"
 
-	cacheParamsFileName     = "pack-cache.yml"
+	cacheParamsFileName     = "pack-cache-config.yml"
 	maxCachedProjects       = 5
 	cntFirstSymbolsFromHash = 10
+
+	cacheParamsErrorMsg = "You should set one of `always-true`, `key` and `key-path` for path %s"
```
dokshina, line 31: Yes, you can use this error message for all cases, but change it:

"You should set one of" -> "You should set one and only one of"

dokshina: Maybe CLI can be a little more polite and say "Please, specify...". Nobody likes when he should do smth.

test/integration/pack/test_pack.py
```diff
@@ -1759,9 +1759,9 @@ def test_no_cache_yml_file(cartridge_cmd, light_project, tmpdir, pack_format):
         cmd.append('--use-docker')
 
     rc, output = run_command_and_get_output(cmd, cwd=tmpdir)
-    assert rc == 1
+    assert rc == 0
     assert "Using cached path" not in output
-    assert f"File {os.path.join(project.path, 'pack-cache.yml')} with pack cache parameters doesn't exists"
+    assert "Failed to process pack-cache-config.yml file which contain cache paths" in output
```
dokshina, line 1764: I don't think that we should show this when the file doesn't exist.
If it exists, but stat failed - it's an error, somebody created this file but it's not possible to use it. If it doesn't exist - we don't need the cache.

test/integration/pack/test_pack.py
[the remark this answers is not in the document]
dokshina: Paths like `.rocks`  contain special characters and should be quoted. It looks ugly for map keys. Github action [`cache`](https://github.com/actions/cache) uses `path` key for that purpose. It can be more convenient for users who configure GitHub CI. I suggest just thinking about it.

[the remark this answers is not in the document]
mRrvz: Agree

cli/pack/app_dir.go
```diff
@@ -106,48 +135,142 @@ func copyFromCache(paths CachePaths, destPath string, ctx *context.Ctx) {
...
+	var keyHash string
+	projectPathHash := common.StringSHA1Hex(ctx.Project.Path)[:cntFirstSymbolsFromHash]
+
+	switch {
+	case params.AlwaysCache == true:
+		keyHash = "always"
+	case params.KeyPath != "":
+		pathFromProjectRoot := filepath.Join(ctx.Project.Path, params.KeyPath)
+		if _, err := os.Stat(pathFromProjectRoot); err == nil {
+			if keyHash, err = common.FileSHA1Hex(pathFromProjectRoot); err != nil {
```
dokshina, line 216:
```go
if _, err := os.Stat(); err != nil {
    return "", fmt.Errorf("Failed to use specified key-path: %s", err)
}

if keyHash ...
```

cli/pack/app_dir.go
```diff
@@ -106,48 +135,142 @@ func copyFromCache(paths CachePaths, destPath string, ctx *context.Ctx) {
...
+				return "", fmt.Errorf("Failed to get hash from file content for path %s: %s", params.Path, err)
+			}
+		} else if os.IsNotExist(err) {
+			return "", fmt.Errorf("Specified key-path file %s for the path %s does not exist", pathFromProjectRoot, params.Path)
+		} else {
+			return "", fmt.Errorf("Failed to get specified file for path %s: %s", params.Path, err)
+		}
+
+		keyHash = keyHash[:cntFirstSymbolsFromHash]
+	default:
```
dokshina, line 226: Better is to add `case params.Key != ""` and a panic by default.

cli/pack/app_dir.go
```diff
@@ -106,48 +135,142 @@ func copyFromCache(paths CachePaths, destPath string, ctx *context.Ctx) {
...
+		keyHash = common.StringSHA1Hex(params.Key)[:cntFirstSymbolsFromHash]
+	}
+
+	return filepath.Join(ctx.Cli.CacheDir, projectPathHash, params.Path, keyHash), nil
+}
+
+func validateCacheParams(params *CachePathParams) error {
+	if params.Key != "" && params.KeyPath != "" {
```
dokshina, line 234: Look at `common.OnlyOneIsTrue`.

cli/pack/app_dir.go
```diff
@@ -106,48 +135,142 @@ func copyFromCache(paths CachePaths, destPath string, ctx *context.Ctx) {
...
+		if err := validateCacheParams(&params); err != nil {
+			return nil, err
+		}
+
+		cachePath, err := calculateCachePath(ctx, &params)
+		if err != nil {
+			return nil, err
+		}
+
+		cachePaths[params.Path] = cachePath
```
dokshina, line 270: What if some path passed twice?

cli/pack/app_dir.go
mRrvz: It seems to me that using this function (and especially with ``!`` operator), only complicates the readability of the code.

cli/pack/app_dir.go
mRrvz: Added check and error in case of duplicate paths

cli/pack/app_dir.go
dokshina: Are you sure?
The whole function can be replaced by

```go
If not common.OnlyOneIsTrue([]bool{params.Key != "", params.KeyPath != "", params.AlwaysCache}) {
    return fmt.Errorf(cacheParamsErrorMsg, params.Path)
}
```

cli/pack/app_dir.go
```diff
@@ -173,7 +302,7 @@ func rotateCacheDirs(ctx *context.Ctx) error {
 }
 
 func updateCache(paths CachePaths, ctx *context.Ctx) error {
-	if ctx.Pack.NoCache {
+	if ctx.Pack.NoCache || paths == nil {
```
dokshina, line 305: `len(paths) == 0`?

cli/pack/app_dir.go
```diff
@@ -106,48 +135,148 @@ func copyFromCache(paths CachePaths, destPath string, ctx *context.Ctx) {
...
 
-	return CachePaths{
-		".rocks": filepath.Join(ctx.Cli.CacheDir, projectPathHash, ".rocks", rockspecHash)}, nil
+		cachePath, err := calculateCachePath(ctx, &params)
+		if err != nil {
+			return nil, err
+		}
+
+		if _, found := cachePaths[params.Path]; found {
+			return nil, fmt.Errorf("You have specified the caching path %s multiple times", params.Path)
```
dokshina, line 273: `Cache path '%s' specified multiple times'?

test/integration/pack/test_pack.py
```diff
@@ -1732,9 +1733,291 @@ def test_rocks_noncache_flag(cartridge_cmd, light_project, tmpdir, pack_format):
...
+        project.path
+    ]
+
+    if platform.system() == 'Darwin':
+        cmd.append('--use-docker')
+
+    rc, output = run_command_and_get_output(cmd, cwd=tmpdir)
+    assert rc == 0
+    assert "Using cached path" not in output
+    assert "Failed to process pack-cache-config.yml file which contain cache paths" not in output
```
dokshina, line 1764: Make this check simpler: `assert "Failed to process pack-cache-config.yml file" not in output`

You don't need a whole message.

test/integration/pack/test_pack.py
```diff
@@ -1732,9 +1733,291 @@ def test_rocks_noncache_flag(cartridge_cmd, light_project, tmpdir, pack_format):
...
+    assert len(os.listdir(get_rocks_cache_path())) == 0
+
+    # Invalid key-path file (non exists)
+    with open(new_cache_yml_path, "w") as f:
+        yaml.dump([{"path": ".rocks", "key-path": "invalid_path"}], f)
+
+    replace_project_file(project, "pack-cache-config.yml", new_cache_yml_path)
+    rc, output = run_command_and_get_output(cmd, cwd=tmpdir)
+    assert rc == 1
+    assert "Failed to get specified file for path .rocks" in output
```
dokshina, line 1919: This error looks strange. 

`Failed to get specified cache key file for .rocks`?

cli/pack/app_dir.go
mRrvz: Agree

dokshina (approved):
Great!

PR #590, opened 2021-07-15
Lint: added unused variables check and removed unused variables from …

mRrvz:


dokshina (commented):
Seems that you forgot to add linter config changes.

dokshina (approved):

PR #588, opened 2021-07-11
Implement ``cartridge failover`` command

mRrvz:
Add ``cartridge failover`` command to manage cartridge failover. 
Closes #507

PR #593, opened 2021-07-19
Functions for managing the cluster are moved to a separate package

mRrvz:
Functions for managing the cluster are moved to a separate package 
called ``cluster``.  Functions from ``replicasets/membership`` and 
``replicasets/common`` were moved there. These functions may be 
used in the future in other packages, for example #588


cli/cluster/cluster.go
```diff
@@ -0,0 +1,244 @@
...
+		return nil, fmt.Errorf("Failed to get membership members: %s", err)
+	}
+
+	membershipInstances := make(MembershipInstances)
+	for _, membershipInstance := range membershipInstancesSlice {
+		membershipInstances[membershipInstance.URI] = membershipInstance
+	}
+
+	return &membershipInstances, nil
+}
```
dokshina, line 244: Let's separate it to three files - `membership.go`, `instances.go` and `cluster.go`.

dokshina (commented):
Good!

dokshina:
I've tried to use the new command:

```bash
# cartridge failover set stateful --state-provider stateboard                                                                           
   ⨯ Please, specify `stateboard_params` when using stateboard provider
```

But this error message doesn't help me to understand that I should specify this 

```bash
$ cartridge failover set  stateful --state-provider stateboard \
    --provider-params '{"uri": "localhost:4401", "password": "passwd"}'
```

but not `--provider-params '{"stateboard_params": {"uri": "localhost:4401", "password": "p"}}'`.

I suggest `Please, specify params for stateboard state provider, using --provider-params '{"uri": "localhost:4401", "password": "passwd"}''`. It should be simply copy-pastable.

dokshina (approved):

PR #602, opened 2021-07-26
Changed docker.io/go-docker package to github.com/docker/docker/client

### LINKED ISSUE #600: NewEnvClient function is deprecated
for example we use this function here: https://github.com/tarantool/cartridge-cli/blob/fde10b1d11a254fc976c2430870a9dd3d0e7a0d1/cli/docker/common.go#L20

https://pkg.go.dev/github.com/docker/docker/client?utm_source=godoc#NewEnvClient

mRrvz:
Closes #600


mRrvz:
Is it normal that go.sum got so big? @dokshina

Totktonada:
> Is it normal that go.sum got so big? @dokshina

I don't know for sure, but I see the similar go.sum size increase in https://github.com/nektos/act/pull/525. AFAIU, those are testing dependencies of the docker client. I don't know, whether we can get rid of them.

dokshina:
Running `go mod tidy` can decrease `go.sum` size.

Issue #475, opened 2021-02-18
Support for custom app versioning (non-semantic)


RunsFor:
Right now, `cartridge-cli` requires cartridge apps to have semantic versioning even for **tar.gz** packages. However, customers may have their own version policies for their apps. Can we add support for custom version policies?

Here is a reproducer:

```bash
$ cartridge create --name myapp
   • Create application myapp
   • Generate application files
   • Initialize application git repository
   • Application "myapp" created successfully
$ cd myapp
$ touch file
$ git add file
$ git commit -m "update repo"
[master 091637d] update repo
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 file
$ git tag v0.1.0
$ cartridge pack tgz .
   • Packing myapp into tgz
   ⨯ Version should be semantic (major.minor.patch[-count][-commit])
```

Totktonada:
I would propose to solve #415 and suggest to pass `VERSION` variable if tags are different from X.Y.Z schema (say, have 'v' prefix).

Totktonada:
However, okay, those are different issues. Your (and #552) proposal is about arbitrary versioning.

Totktonada:
To be honest, I don't know why the decision about enforcing semver was made. So I don't know, whether it'll be appropriate to change it.

@dokshina Can you shed some light there?

Kasen:
I know that history. This decision was made by @knazarov cause forcing of right practice is a good thing.

PR #604, opened 2021-07-26
Change cartridge pack naming strategy

### LINKED ISSUE #303: pack rpm should respect rpm naming rules
`[name]-[version]-[release].[arch].rpm`

https://www.thegeekdiary.com/understanding-rpm-versions-and-naming-schemes/

mRrvz:
Reworked by @DifferentialOrange 

pack: change artifact naming strategy

After this patch, artifact names will be as follows:
  * RPM: `<app-name>-<version>[.<suffix>]-1.<arch>.rpm`,
  * DEB: `<app-name>_<version>[.<suffix>]-1_<arch>.deb`,
  * TGZ: `<app-name>-<version>[.<suffix>].<arch>.tar.gz`.
With this change, RPM and DEB packages will satisfy naming
conventions [1, 2, 3]. `<version>` is `git describe` output normalized
to `X.Y.Z.N` or non-transformed ``--version`` value. Previously it was
transformed to `X.Y.Z-N-gHASH` with ability to set `X.Y.Z-N`,
`X.Y.Z-gHASH` with `--version`.

After this patch, ``--version`` value is used as provided without
transformation. Previously it had the same restrictions and
transformations as `git describe --tags --long` output. It can be used
to set non-standard values (values still should satisfy conventions).

Package architecture is `all` for DEB and `x86_64` for RPM and TGZ
to preserve backward compatibility. Restricting is not provided now,
refer to #675.

Since ``--version`` is no longer forced to be semantic, it may be used
to workaround #475.

1. https://www.debian.org/doc/debian-policy/ch-controlfields.html#version
2. https://www.debian.org/doc/manuals/debian-reference/ch02.en.html#_debian_package_file_names
3. http://ftp.rpm.org/max-rpm/ch-rpm-file-format.html

Closes #303, part of #475


cli/pack/common_test.go
```diff
@@ -26,10 +28,10 @@ func TestGetPackageFullname(t *testing.T) {
 	assert.Equal("myapp-1.2.3-4.tar.gz", getPackageFullname(&ctx))
 
 	ctx.Pack.Type = RpmType
-	assert.Equal("myapp-1.2.3-4.rpm", getPackageFullname(&ctx))
+	assert.Equal(fmt.Sprintf("myapp-1.2.3-4.%s.rpm", runtime.GOARCH), getPackageFullname(&ctx))
 
 	ctx.Pack.Type = DebType
-	assert.Equal("myapp-1.2.3-4.deb", getPackageFullname(&ctx))
+	assert.Equal(fmt.Sprintf("myapp_1.2.3-4-1_%s.deb", runtime.GOARCH), getPackageFullname(&ctx))
```
Totktonada, line 34: What looks strange for me: `4` is the distribution version on RPM based distros, but `1` is the distribution version on the Deb based distros.

cli/pack/common.go
```diff
@@ -108,33 +111,59 @@ func detectVersion(ctx *context.Ctx) error {
...
-			ctx.Pack.Suffix,
-		)
+func getDebPackageNameBody(ctx *context.Ctx) string {
+	return fmt.Sprintf(
+		"%s_%s-%s%s_%s",
+		ctx.Project.Name,
+		ctx.Pack.VersionRelease,
+		debianVersion,
+		getPackageSuffix(ctx),
+		runtime.GOARCH,
```
dokshina, line 140: When building in docker, this value will depend on the image architecture, so using `runtime` isn't honest here.

cli/pack/common.go
```diff
@@ -108,33 +111,59 @@ func detectVersion(ctx *context.Ctx) error {
...
-		packageFullname,
-		ext,
-	)
+	var packageFullname string
+
+	switch ctx.Pack.Type {
+	case RpmType:
+		packageFullname = getRpmPackageNameBody(ctx)
+	case DebType:
+		packageFullname = getDebPackageNameBody(ctx)
```
dokshina, line 161: Why not simple `get{Rpm,Deb}PackageFullname`? In fact, `ext` depends on pack type.  The default case is used for TGZ only.

cli/pack/common.go
```diff
@@ -108,33 +111,61 @@ func detectVersion(ctx *context.Ctx) error {
...
+	if ctx.Pack.Suffix != "" {
+		suffix = fmt.Sprintf("-%s", ctx.Pack.Suffix)
 	}
 
-	packageFullname := fmt.Sprintf(
-		"%s-%s",
+	return suffix
+}
+
+func getRpmPackageFullname(ctx *context.Ctx) string {
```
Totktonada, line 123: I would like to see here a link to the document you lean on. The same for `getDebPackageFullname()`.

cli/pack/common_test.go
```diff
@@ -19,30 +21,30 @@ func TestGetPackageFullname(t *testing.T) {
 
 	// w/o suffix
 	ctx.Project.Name = "myapp"
-	ctx.Pack.VersionRelease = "1.2.3-4"
+	ctx.Pack.VersionRelease = "1.2.3.g480c55b67"
```
Totktonada, line 24: (Please, take me right: I'm still mostly out of the context of the project. So all my questions could be just because I don't dive into. And, please, don't take me as a source of truth about RPM / Deb packages: I'm not sure about all statements below.)

To be honest, I don't understand this change. I see that `normalizeVersion()` concatenates `Version` and `Release` fields. Okay. The `Version` is in the X.Y.Z form. Okay. The `Release` is either `<count>-<hash>` or `<count>` or `<hash>` or `0` (???). And there I have several questions.

1. The unit test check one case (`Release` is `<count>`), but now it checks another one (`Release` is `<hash>`). It would be easier to understand how your patch changes the behaviour if the inputs of the test case would remain unchanged. Or if you would add more test cases (and don't remove old ones) before the actual change of the code (in a separate commit).
2. As I see `VersionRelease` is formed as `<Version>-<Release>`. Not `<Version>.<Release>`. The test case is incorrect (I mean, it checks impossible situation), right?
3. I don't understand **when** `Release` is just `<count>`, when it is `<hash>`, when it is `<count>-<hash>` and when it is `0`.
4. Why the git hash is in the tarball name? That's quite strange, because a tarball source URL should be predictable by a version, if someone plan to download tarballs automatically. It is quite usual for, say, Gentoo ebuilds.
5. It also looks unusual to see the git hash in the RPM package name (however I don't know for sure). AFAIR, there is some new syntax for adding hash information into a package name, but it is modern and so rarely used now. And it is different (if I remember right, it uses a caret symbol and has its own rules).
6. AFAIR, it is usual for Debian to include a git hash into a 'preview package' or something like that. However (again, if I remember right) a release should not contain it. Whether we handle both situations correctly and it is obvious for users how to build packages in the right way?

cli/pack/common.go
mRrvz: Added

cli/pack/common_test.go
mRrvz: I can't answer the questions related to the function ``normalizeVersion``

1. Fixed
2. Right

cli/pack/common_test.go
```diff
@@ -19,30 +21,30 @@ func TestGetPackageFullname(t *testing.T) {
...
 	ctx.Pack.Type = TgzType
-	assert.Equal("myapp-1.2.3-4-dev.tar.gz", getPackageFullname(&ctx))
+	assert.Equal("myapp-1.2.3.g480c55b67-dev.tar.gz", getPackageFullname(&ctx))
 
 	ctx.Pack.Type = RpmType
-	assert.Equal("myapp-1.2.3-4-dev.rpm", getPackageFullname(&ctx))
+	assert.Equal(fmt.Sprintf("myapp-1.2.3.g480c55b67-dev.%s.rpm", runtime.GOARCH), getPackageFullname(&ctx))
 	ctx.Pack.Type = DebType
-	assert.Equal("myapp-1.2.3-4-dev.deb", getPackageFullname(&ctx))
+	assert.Equal(fmt.Sprintf("myapp_1.2.3.g480c55b67-1-dev_%s.deb", runtime.GOARCH), getPackageFullname(&ctx))
```
DifferentialOrange, line 47: This is invalid. Here `1-dev` is `debian.revision`, but by standard dashes are not supported here: `[a-zA-Z0-9.+~]+`. See [doc](https://www.debian.org/doc/manuals/debian-reference/ch02.en.html#_debian_package_file_names).

DifferentialOrange: Same about RPM: http://ftp.rpm.org/max-rpm/ch-rpm-file-format.html

cli/pack/common.go
DifferentialOrange: Well, why it isn't honest? Docker container uses the kernel and CPU of the host machine (for example, see [this question](https://www.edureka.co/community/65179/do-docker-containers-have-their-own-kernel-and-cpu)). If docker runs under VM, `runtime.GOARCH` will be the same as in VM, but package will also be built under VM setup. It seems like it is fair in all of the cases.

cli/pack/common_test.go
DifferentialOrange: Now hash is stripped from all artifacts and version names are based on strict conventions. See reworked version

DifferentialOrange:
After discussion with @Totktonada we've decided to normalize all packages version. I think it is worth to discuss such changes (`X.Y.Z.N`, hash strip) with our users (mainly Delivery team). But since in is already the evening of Dec 30, seems like the question is left until January.

DifferentialOrange:
Also I leave this PR in WIP state until I return to work in January, so it's not worth to review it for now (especially since tests will be broken). There are too many small things I should verify so I'm not sure I will finished in time today.

cli/pack/common.go
DifferentialOrange: Use `noarch`/`all` to not imply any new restrictions on package, see #675 for extension of this idea

DifferentialOrange:
I've got really tired of restarting flaky tests, so I won't do that for a while

DifferentialOrange:
@Totktonada , @LeonidVas , GitHub do not allow me to rerequest review in status "Awaiting requested review", so I'll ping you with this

cli/pack/common.go
Totktonada: I guess Liza pointing out that docker can run an image from a foreign architecture using qemu. Anyway, it is irrelevant now and I'll resolve this discussion.

cli/pack/common.go
```diff
@@ -28,14 +28,24 @@ var (
...
+
+	// Check if --version or --suffix is allowed to be added in package version.
+	packageVerAllowed = map[string]*regexp.Regexp{
+		// The version number is normally taken verbatim from the package's version.
+		// The only restriction placed on the version is that it cannot contain a dash "-".
+		// http://ftp.rpm.org/max-rpm/ch-rpm-file-format.html
+		RpmType: regexp.MustCompile(`^[^-]+$`),
+		// upstream-version must satisfy `[-a-zA-Z0-9.+:]+`.
+		// https://www.debian.org/doc/manuals/debian-reference/ch02.en.html#_debian_package_file_names
+		DebType: regexp.MustCompile(`^[-a-zA-Z0-9.+:]+$`),
```
Totktonada, line 40: A tilde possibly should be here as well. See https://github.com/knazarov/mkrepo/blob/cdc6ef352476c2ed7ec18cb38583328deaf1baf5/debrepo.py#L450-L460.

CHANGELOG.md
```diff
@@ -5,6 +5,21 @@ All notable changes to this project will be documented in this file.
...
+  * DEB: `<app-name>-<version>[.<suffix>]-1_<arch>.deb`,
+  * TGZ: `<app-name>-<version>[.<suffix>].tar.gz`.
+  `<version>` is `git describe` output normalized to `X.Y.Z.N` or ``--version`` value.
+- **Breaking**: ``cartridge pack --version`` value is used as provided without normalization.
+- **Breaking**: ``cartridge pack`` ``--version`` and ``--suffix`` is validated
+  after RPM/DEB conventions for corresponding type (see
+  https://www.debian.org/doc/manuals/debian-reference/ch02.en.html#_debian_package_file_names
+  and http://ftp.rpm.org/max-rpm/ch-rpm-file-format.html).
+  For example, dashes in rpm version (`1.2.3-0`) is no longer supported.
+- RPM is now built for `noarch` instead of `x86_64`.
```
Totktonada, line 21: Maybe keep it x86_64 for now and implement achitecture detection later? At least we can have compiled modules, which will not run on another architecture. Also we pack tarantool enterprise executable into the application if a user works under tarantool enterprise.

CHANGELOG.md
```diff
@@ -5,6 +5,21 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+- **Breaking**: change ``cartridge pack`` naming strategy:
+  * RPM: `<app-name>-<version>[.<suffix>]-1.<arch>.rpm`,
+  * DEB: `<app-name>-<version>[.<suffix>]-1_<arch>.deb`,
+  * TGZ: `<app-name>-<version>[.<suffix>].tar.gz`.
+  `<version>` is `git describe` output normalized to `X.Y.Z.N` or ``--version`` value.
```
Totktonada, line 14: It would be good to show the old convention to compare.

Totktonada (approved):
The changes are okay for me in general.

CHANGELOG.md
```diff
@@ -5,6 +5,21 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+- **Breaking**: change ``cartridge pack`` naming strategy:
+  * RPM: `<app-name>-<version>[.<suffix>]-1.<arch>.rpm`,
+  * DEB: `<app-name>-<version>[.<suffix>]-1_<arch>.deb`,
+  * TGZ: `<app-name>-<version>[.<suffix>].tar.gz`.
```
Totktonada, line 13: TGZ is not for sources, right? So we should have an architecture here as well?

cli/pack/common.go
DifferentialOrange: Thanks, added

CHANGELOG.md
DifferentialOrange: Restored old behavior

CHANGELOG.md
DifferentialOrange: Added an entry

CHANGELOG.md
DifferentialOrange: I'm a bit stressed out about this one since we no longer will be able to set a TGZ name with `--version`to be **exactly** equal to pre-patch one, but the point seems fair. Added it.

Totktonada: 1. It seems, it is not reflected in the changelog (for tgz).
2. AFAIS, Debian package has 'all' architecture.

CHANGELOG.md
```diff
@@ -5,6 +5,24 @@ All notable changes to this project will be documented in this file.
...
+## [Unreleased]
+
+- **Breaking**: change ``cartridge pack`` naming strategy:
+  * RPM: `<app-name>-<version>[.<suffix>]-1.<arch>.rpm`,
+  * DEB: `<app-name>_<version>[.<suffix>]-1_<arch>.deb`,
+  * TGZ: `<app-name>-<version>[.<suffix>].tar.gz`.
+  `<version>` is `git describe` output normalized to `X.Y.Z.N` or ``--version`` value.
+  (Previously it was `X.Y.Z-N-gHASH` if commits since tag and commit hash are provided,
+  `X.Y.Z-N` if only commits since tag is provided, `X.Y.Z-gHASH` if only commit hash is provided,
+  `X.Y.Z-0` if only version is provided.)
```
Totktonada, line 17: How can I provide it or don't provide? I just run `cartridge pack` in a repository, not?

CHANGELOG.md
```diff
@@ -5,6 +5,24 @@ All notable changes to this project will be documented in this file.
...
+- **Breaking**: ``cartridge pack --version`` value is used as provided without normalization.
```
Totktonada, line 18: What kind of normalization was applied here before? Now this point looks like 'something was changed', but it is not clear, how exactly.

Totktonada: I see it in the `pack.rst`. But here, without necessary context, it was unclear.

CHANGELOG.md
DifferentialOrange: You were able to pass `--version` explicitly, and this value was parsed in code

CHANGELOG.md
DifferentialOrange: `--version` was treated the same as `git describe` result. It was described in

> (Previously it was `X.Y.Z-N-gHASH` if commits since tag and commit hash are provided,
  `X.Y.Z-N` if only commits since tag is provided, `X.Y.Z-gHASH` if only commit hash is provided,
  `X.Y.Z-0` if only version is provided.)

CHANGELOG.md
DifferentialOrange: 1. Fixed
2. What's the question? To replace <arch> with specific values since for now they all are constants?

CHANGELOG.md
DifferentialOrange: Added a "Previously it was" line to the changelog

CHANGELOG.md
DifferentialOrange: Closed after live discussion

CHANGELOG.md
DifferentialOrange: Fixed descriptions after discussion

CHANGELOG.md
DifferentialOrange: Fixed descriptions after discussion

Totktonada:
Ok to push.

LeonidVas (approved):

PR #606, opened 2021-07-28
CI: Add build crearepo_s tool to publishS3 job

mRrvz:
``update_repo`` script [in tarantool repository](https://github.com/tarantool/tarantool/commit/431c043af63179ae7ac145a5ecaec7bd36db6b43#diff-51435e46617eb7267f50ed8c8f174db3818e65442fcf60600e9824b2b424363e) now uses the ``createrepo_c`` tool 
instead of ``createrepo``. The patch adds this tool to publish S3 pipeline.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -69,7 +69,21 @@ jobs:
       - name: Install packages required for publishing script
         run: |
           sudo apt-get -y update
-          sudo apt-get install -y procmail createrepo awscli reprepro
+          sudo apt-get install -y procmail awscli reprepro
+
+      - name: Build createrepo_c tool
+        run: |
+          git clone https://github.com/rpm-software-management/createrepo_c.git && cd createrepo_c/
+          git checkout 0.17.1
+          mkdir build && cd build
+          sudo apt install -y libbz2-dev cmake libmagic-dev \
+            libglib2.0-dev libcurl4-openssl-dev libxml2-dev \
+            libpython3-dev librpm-dev libssl-dev libsqlite3-dev \
+            liblzma-dev zlib1g-dev libzstd-dev doxygen
+          cmake .. -DWITH_ZCHUNK=OFF -DWITH_LIBMODULEMD=OFF
+          make -j $(nproc)
+          sudo make install
+          cd ../..
 
       - name: Import GPG key
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -69,7 +69,20 @@ jobs:
       - name: Install packages required for publishing script
         run: |
           sudo apt-get -y update
-          sudo apt-get install -y procmail createrepo awscli reprepro
+          sudo apt-get install -y procmail awscli reprepro
+
+      - name: Build creatrepo_c tool
```
Totktonada, line 74: Typo: creat -> create.

.github/workflows/release.yml
```diff
@@ -69,7 +69,20 @@ jobs:
       - name: Install packages required for publishing script
         run: |
           sudo apt-get -y update
-          sudo apt-get install -y procmail createrepo awscli reprepro
+          sudo apt-get install -y procmail awscli reprepro
+
+      - name: Build createrepo_c tool
+        run: |
+          git clone https://github.com/rpm-software-management/createrepo_c.git && cd createrepo_c/
```
Totktonada, line 76: Maybe checkout a tag?

Totktonada:
Can we verify it in CI somehow before actually push?

.github/workflows/release.yml
mRrvz: Yes, I think need to switch to 0.17.1, as is done in tarantool repo.

mRrvz:
> Can we verify it in CI somehow before actually push?

I think only if you run the publish s3 script locally. But it's not easy

@Totktonada

Totktonada:
How about this way?

```diff
diff --git a/.github/workflows/release.yml b/.github/workflows/release.yml
index b60e73a..8daeade 100644
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,10 +4,6 @@ name: Release
 on:
   workflow_dispatch:
   push:
-    branches:
-      - master
-    tags:
-      - '*'
 
 env:
   GO_VERSION: 1.15
diff --git a/magefile.publish.go b/magefile.publish.go
index 4e56b79..344b3c8 100644
--- a/magefile.publish.go
+++ b/magefile.publish.go
@@ -110,7 +110,7 @@ func getS3Ctx() (map[string]string, error) {
                return nil, fmt.Errorf("Please, specify %s", s3BucketEnv)
        }
 
-       s3Folder := os.Getenv(s3FolderEnv)
+       s3Folder := "staging/modules"
        if s3Folder == "" {
                return nil, fmt.Errorf("Please, specify %s", s3FolderEnv)
        }
```

Totktonada (approved):
It works, so it is okay. I hope we'll able to change the publication script using HTTP POST uploading via [rws](https://github.com/tarantool/rws/) soon.

PR #607, opened 2021-07-28
CI: Change RELEASE_GITHUB_TOKEN to GITHUB_TOKEN in release job

mRrvz:


Totktonada (approved):
Should work, as I see from https://goreleaser.com/ci/actions/

PR #608, opened 2021-07-29
CI: Add missing asserts and socket file checks

### LINKED ISSUE #560: Refactor ``check_systemd_service`` function
There is no assert for ``check_contains_file`` or ``check_contains_dir`` (e.g. https://github.com/tarantool/cartridge-cli/blob/master/test/utils.py#L1150 ) in ``check_systemd_service``.

mRrvz:
Closes #560

### DIFF
--- a/test/utils.py
+++ b/test/utils.py
@@ -1153,6 +1153,7 @@ def check_systemd_service(container, project, http_port, tmpdir):
 
     replicaset_uuid = create_replicaset(admin_api_url, [advertise_uri], roles)
     wait_for_replicaset_is_healthy(admin_api_url, replicaset_uuid)
+    assert check_contains_file(container, '/var/run/tarantool/%s.pid' % instance_id)
 
     container.restart()
     wait_for_replicaset_is_healthy(admin_api_url, replicaset_uuid)
--- a/test/utils.py
+++ b/test/utils.py
@@ -1101,8 +1101,8 @@ def check_contains_dir(container, dirpath):
     return run_command_on_container(container, command) == 'true'
 
 
-def check_contains_file(container, filepath):
-    command = '[ -f "{}" ] && echo true || echo false'.format(filepath)
+def check_contains_file(container, filepath, is_socket=False):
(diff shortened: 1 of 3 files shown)


test/utils.py
```diff
@@ -1145,9 +1145,8 @@ def check_systemd_service(container, project, http_port, tmpdir):
 
     wait_for_systemd_service(container, service_name)
 
-    check_contains_dir(container, '/var/lib/tarantool/%s' % instance_id)
-    check_contains_file(container, '/var/run/tarantool/%s.control' % instance_id)
-    check_contains_file(container, '/var/run/tarantool/%s.pid' % instance_id)
+    assert check_contains_dir(container, '/var/lib/tarantool/%s' % instance_id)
+    assert check_contains_file(container, '/var/run/tarantool/%s.control' % instance_id, is_socket=True)
```
AnaNek, line 1150: Why there is no `check_contains_file` for `/var/run/tarantool/*.pid`?

test/utils.py
```diff
@@ -1101,8 +1101,8 @@ def check_contains_dir(container, dirpath):
     return run_command_on_container(container, command) == 'true'
 
 
-def check_contains_file(container, filepath):
-    command = '[ -f "{}" ] && echo true || echo false'.format(filepath)
+def check_contains_file(container, filepath, is_socket=False):
+    command = '[ -{} "{}" ] && echo true || echo false'.format('S' if is_socket else 'f', filepath)
```
AnaNek, line 1105: Maybe it would be better to split this function to two functions: `check_contains_regular_file` and `check_contains_socket_file`?

test/utils.py
mRrvz: Problem was that the *.pid* file was not created. But later I found out that it is created after ``box.cfg`` call (after instances bootstrap). Put assert below

test/utils.py
mRrvz: agree

test/utils.py
AnaNek: Ok, thank you for your answer!

AnaNek (approved):

PR #609, opened 2021-07-30
Reverse search in connect commands

### LINKED ISSUE #592: reverse search not working
I often use reverse search to find command in bash history. That is Ctrl + R shortcut and then search term. This is not working with cartridge-cli while being connected to tarantool via `cartridge connect`

mRrvz:
Closes #592


mRrvz (commented):
I forked the [go-prompt fork that we use](https://github.com/binaek89/go-prompt) and implemented my reverse-search inside the library. More details here: https://github.com/mRrvz/go-prompt/compare/v0.2.7-multiline-clearscreen...v0.4.0

go.mod
```diff
@@ -40,10 +40,9 @@ require (
...
 	github.com/vmihailenco/msgpack/v5 v5.1.0
 	github.com/yuin/gopher-lua v0.0.0-20191220021717-ab39c6098bdb
-	golang.org/x/tools v0.0.0-20200609124132-5359b67ffbdf // indirect
 	gopkg.in/yaml.v2 v2.3.0
 )
 
 replace github.com/docker/docker/internal/testutil => gotest.tools/v3 v3.0.0
 
-replace github.com/c-bata/go-prompt => github.com/binaek89/go-prompt v0.2.7-multiline-clearscreen
+replace github.com/c-bata/go-prompt => github.com/mrrvz/go-prompt v0.4.0
```
LeonidVas, line 48: AFAIU we change the dependency to fork with extended functionality. If so, then I would prefer this fork to be inside the "tarantool" organization.

mRrvz: I think we need to involve @Totktonada

mRrvz: Moved to tarantool org repo: https://github.com/tarantool/go-prompt

PR #613, opened 2021-08-03
pack: ignore data, run and log directories

### LINKED ISSUE #494: ignore "/tmp" content when pack application
```
cartridge pack rpm --use-docker
   • Packing lesson into rpm  
   • Temporary directory is set to /Users/o.babin/.cartridge/tmp/pack-3cmctvaknr
   • Initialize application dir
   ⨯ File /Users/o.babin/.cartridge/tmp/pack-3cmctvaknr/package-files/usr/share/tarantool/lesson/.tarantool.cookie has invalid mode: 600. It should have read permissions for all
```

I suggest to ignore content of /tmp directory. I run my application before pack, so it's not a reason to make me call `cartridge clean` manually before pack.

mRrvz:
Reworked by @DifferentialOrange

``cartridge pack`` command now ignores the contents of ``run_dir``,
``data_dir``, ``log_dir`` directories, if default directories used or
set with ``.cartridge.yml``. Since cartridge-cli do not store info on
flags between runs, if flags like ``--run-dir`` were used, contents
of these directories will not be ignored.

Closes #494

### DIFF
--- a/cli/pack/app_dir.go
+++ b/cli/pack/app_dir.go
@@ -16,6 +16,7 @@ import (
 	"github.com/tarantool/cartridge-cli/cli/build"
 	"github.com/tarantool/cartridge-cli/cli/common"
 	"github.com/tarantool/cartridge-cli/cli/context"
+	"github.com/tarantool/cartridge-cli/cli/project"
 )
 
 const (
@@ -325,6 +326,30 @@ func updateCache(paths CachePaths, ctx *context.Ctx) error {
 }
 
 func copyProjectFiles(dst string, ctx *context.Ctx) error {
+	// We need to ignore run_dir, data_dir and log_dir folders because it
+	// contains files of the running instance and they can then block app pack.
+	// More details: https://github.com/tarantool/cartridge-cli/issues/494
+
+	// ctx uses paths from project.SetSystemRunningPaths to pack application.
+	// To get expected application run_dir, data_dir and log_dir we need to get
+	// project.SetLocalRunningPaths result.
+	var appCtx context.Ctx
+	if err := project.SetLocalRunningPaths(&appCtx); err != nil {
+		return fmt.Errorf("Failed to get local running paths: %s", err)
+	}
+
+	appLocalPaths := []string{
+		appCtx.Running.RunDir,
+		appCtx.Running.DataDir,
+		appCtx.Running.LogDir,
+	}
+
+	appLocalDirs := make([]string, len(appLocalPaths))
+
+	for i, appLocalPath := range appLocalPaths {
+		appLocalDirs[i] = fmt.Sprintf("%s/", appLocalPath)
+	}
+
 	err := copy.Copy(ctx.Project.Path, dst, copy.Options{
 		Skip: func(src string) (bool, error) {
 			if strings.HasPrefix(src, fmt.Sprintf("%s/", ctx.Cli.CartridgeTmpDir)) {
@@ -334,6 +359,13 @@ func copyProjectFiles(dst string, ctx *context.Ctx) error {
 			relPath, err := filepath.Rel(ctx.Project.Path, src)
 			if err != nil {
 				return false, fmt.Errorf("Failed to get file rel path: %s", err)
+			}
+
+			// Ignore run_dir, data_dir and log_dir folders.
+			for i, appLocalPath := range appLocalPaths {
+				if src == appLocalPath || strings.HasPrefix(src, appLocalDirs[i]) {
+					return true, nil
+				}
 			}
 
 			if relPath == ".rocks" || strings.HasPrefix(relPath, ".rocks/") {
--- a/cli/project/files.go
+++ b/cli/project/files.go
@@ -246,7 +246,7 @@ func SetLocalRunningPaths(ctx *context.Ctx) error {
 		GetAbs:          true,
 	})
 	if err != nil {
-		return fmt.Errorf("Failed to detect data dir: %s", err)
+		return fmt.Errorf("Failed to detect log dir: %s", err)
 	}
 
 	// set entrypoints
--- a/test/integration/pack/test_pack.py
+++ b/test/integration/pack/test_pack.py
@@ -9,8 +9,8 @@ import tarfile
 
 import pytest
 import yaml
-from project import (remove_project_file, replace_project_file,
-                     set_and_return_whoami_on_build)
+from project import (RUNDIR_CLI_CONF, remove_project_file,
+                     replace_project_file, set_and_return_whoami_on_build)
 from utils import (Archive, assert_dependencies_deb, assert_dependencies_rpm,
                    assert_distribution_dir_contents, assert_filemodes,
                    assert_files_mode_and_owner_rpm,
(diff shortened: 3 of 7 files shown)


test/integration/pack/test_pack.py
```diff
@@ -2184,3 +2184,21 @@ def test_net_msg_max_invalid_value(cartridge_cmd, project_without_dependencies,
...
+@pytest.mark.parametrize('pack_format', ['tgz'])
+def test_ignore_tmp_dir_content(cartridge_cmd, project_without_dependencies, pack_format, tmpdir):
+    project = project_without_dependencies
+
+    with open(os.path.join(project.path, "tmp", "tmpfile"), "w") as f:
+        f.write("test text")
+
+    cmd = [cartridge_cmd, "pack", pack_format, project.path]
+
+    rc, output = run_command_and_get_output(cmd, cwd=tmpdir)
```
AnaNek, line 2198: How about `subprocess.run(cmd, cwd=project.path)` since you don't need `output`?

cli/pack/app_dir.go
```diff
@@ -336,6 +336,10 @@ func copyProjectFiles(dst string, ctx *context.Ctx) error {
 				return false, fmt.Errorf("Failed to get file rel path: %s", err)
 			}
 
+			if strings.HasPrefix(relPath, "tmp/") {
```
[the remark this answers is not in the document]
LeonidVas: Please, comment in the code on what you are doing and why. It will be useful in the future.

LeonidVas (changes requested):
Hi! Thank you for the patch.
I have a several common questions:
1) AFAIU before push to master all commits will be squashed to one. What is the title and message commit will be have?
2) Do you have some rules of a commit(PR) decoration?

cli/pack/app_dir.go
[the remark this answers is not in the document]
mRrvz: ``run_dir`` (by default) is in ``tmp`` folder. But, besides ``run_dir``, ``data and ``log`` are also stored there. I'm not sure if we can ignore them either.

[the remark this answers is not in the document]
mRrvz: Cartridge CLI allows to describe ``run_dir`` in two ways:

* pass ``--run_dir`` flag directly
* describe it in ``.cartridge.yml`` file

In second we case we can get ``run_dir`` from .cartridge.yml (and ignore this folder). 

But, in first case, if you user always specify this parameters with ``--run_dir`` flag, we cant determine the real ``run_dir``. 

I suggest to just ignore the ``/tmp`` folder (like in current patch). Or ignore ``/tmp`` and parameters set in ``.cartridge.yml``. I think first option is better.

mRrvz:
@LeonidVas 

1) I usually copy from the description to the pull request. 
2) We don't have any commit rules. Except that we squash them into one before pushing.

test/integration/pack/test_pack.py
mRrvz: Fixed

cli/pack/app_dir.go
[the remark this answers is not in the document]
mRrvz: I added ignoring ``tmp/data``, ``tmp/log``, ``tmp/run`` to the code.

> We can't understand inside the program with what flag it was run? Sounds strange.

Yes, because this flags specifying when user type ``cartridge start`` command (not ``cartridge pack``.

cli/pack/app_dir.go
```diff
@@ -336,6 +337,15 @@ func copyProjectFiles(dst string, ctx *context.Ctx) error {
 				return false, fmt.Errorf("Failed to get file rel path: %s", err)
 			}
 
+			// We need to ignore the run_dir, data_dir and log_dir folders because it
+			// contains the files of the running instance, which can then block application pack.
+			// More details: https://github.com/tarantool/cartridge-cli/issues/494
+			if strings.HasPrefix(relPath, project.DefaultLocalDataDir) ||
```
LeonidVas, line 343: Why can't we get these values from the `Running context` (before calling the function `SetLocalRunningPaths`)?
AFAUI in this case we will also take into account the value from the config and `defaultLocal...Dir`will remain private as they were.

LeonidVas (changes requested):
I suggest in commit title to use the "names" of the directory(`data`, `log`, `run`) instead of paths.

cli/pack/app_dir.go
mRrvz: If you talk about ``ctx.Running.RunDir`` (and others), they point to the wrong directories. For example ctx.Running.RunDir points to `/var/run/tarantool`. This is set to into `SetSystemRunningPaths` function (`SetLocalRunningPaths` function isn't called wheb pack processing). Before `SetSystemRunningPaths` call, this fields is empty.

mRrvz:
> I suggest in commit title to use the "names" of the directory(data, log, run) instead of paths.

Why? But in reality we don't ignore ``run_dir``: for example, if we specify ``--run_dir`` using a flag, we can't ignore it, because we don't know anything about it (``cartridge start`` and ``cartridge pack`` are two different program calls)

In this case, I suggest adding the word default: 

> Ignore default run_dir, data_dir, log_dir directories when ...

cli/pack/app_dir.go
LeonidVas: >`SetLocalRunningPaths` function isn't called when pack processing
Yep, but why can't we start doing this? In this case, `ctx.Running.RunDir` will contain the actual value.

mRrvz: Agree, change `SetSystemRunningPaths` -> `SetLocalRunningPaths`

mRrvz: In fact, we cannot make this replacement. Some of the paths that are formed in SetSystemRunningPaths are needed during the packaging process. I suggest leaving the changes as they are now (or just ignore the ``tmp/`` folder)

LeonidVas: Wow, how difficult it is. I will try to propose yet one idea. We can add to `files.go` a new function `GetLocalRunningPaths` which will return all "local" paths, call it inside `copyProjectFiles` and exclude the corresponding paths.

mRrvz: Added.

cli/project/files.go
```diff
@@ -281,6 +282,12 @@ func SetLocalRunningPaths(ctx *context.Ctx) error {
 	return nil
 }
 
+// GetDefaultLocalRunningPaths returns a list of default
+// local paths RunDir, DataDir, LogDir.
+func GetDefaultLocalRunningPaths() []string {
```
LeonidVas, line 287: Why "Get**Default**LocalRunningPaths" ? AFAIU we can return one of the default or value from config as in "SetLocalRunningPaths". (If you do so, remove "tmp..." from commit title and changelog).

cli/project/files.go
```diff
@@ -15,10 +15,11 @@ const (
 	defaultEntrypoint           = "init.lua"
 	defaultStateboardEntrypoint = "stateboard.init.lua"
 
+	defaultLocalRunDir  = "tmp/run"
```
LeonidVas, line 18: Unnecessary moving of variables.

cli/pack/app_dir.go
```diff
@@ -336,6 +339,15 @@ func copyProjectFiles(dst string, ctx *context.Ctx) error {
 				return false, fmt.Errorf("Failed to get file rel path: %s", err)
 			}
 
+			// We need to ignore the run_dir, data_dir and log_dir folders because it
+			// contains the files of the running instance, which can then block application pack.
+			// More details: https://github.com/tarantool/cartridge-cli/issues/494
+			for _, path := range defaultLocalPaths {
```
LeonidVas, line 345: Why not just `for _, path := range project.GetDefaultLocalRunningPaths()` ?

mRrvz: In this case we will call ``project.GetDefaultLocalRunningPaths()`` many times (versus only one call) - because this function called for every path in project

cli/project/files.go
mRrvz: Fixed

DifferentialOrange:
If `--run-dir` was passed, cli do (or at least pretends it to) not know what is current run-dir:
```bash
georgymoiseev@georgymoiseev-HP-ProBook-440-G5:~/Development/sandbox/testrundir$ cartridge start -d --run-dir rundir
   • testrundir-stateboard... OK
   • testrundir.router... OK
   • testrundir.s2-replica... OK
   • testrundir.s1-replica... OK
   • testrundir.s2-master... OK
   • testrundir.s1-master... OK

georgymoiseev@georgymoiseev-HP-ProBook-440-G5:~/Development/sandbox/testrundir$ cartridge enter router
   ⨯ Instance router is not running

georgymoiseev@georgymoiseev-HP-ProBook-440-G5:~/Development/sandbox/testrundir$ cartridge enter router --run-dir rundir
connected to testrundir.router
testrundir.router> 
^C

georgymoiseev@georgymoiseev-HP-ProBook-440-G5:~/Development/sandbox/testrundir$ cartridge stop
   • testrundir-stateboard... SKIPPED
   • testrundir.router... SKIPPED
   • testrundir.s1-master... SKIPPED
   • testrundir.s1-replica... SKIPPED
   • testrundir.s2-master... SKIPPED
   • testrundir.s2-replica... SKIPPED
   • testrundir-stateboard: Process is not running
   • testrundir.router: Process is not running
   • testrundir.s1-master: Process is not running
   • testrundir.s1-replica: Process is not running
   • testrundir.s2-master: Process is not running
   • testrundir.s2-replica: Process is not running

georgymoiseev@georgymoiseev-HP-ProBook-440-G5:~/Development/sandbox/testrundir$ cartridge stop --run-dir rundir
   • testrundir-stateboard... OK
   • testrundir.router... OK
   • testrundir.s1-master... OK
   • testrundir.s1-replica... OK
   • testrundir.s2-master... OK
   • testrundir.s2-replica... OK
```

DifferentialOrange:
If run-dir is set with `.cartridge.yml`, it is recognized by all commands implicitly
```bash
georgymoiseev@georgymoiseev-HP-ProBook-440-G5:~/Development/sandbox/testrundir$ cat .cartridge.yml 
---
# Here you can specify default parameters for local running, such as:

# cfg: path-to-cfg-file
# log-dir: path-to-log-dir
run-dir: rundir2
# data-dir: path-to-data-dir

stateboard: true # If no arguments are supplied, --stateboard flag is true

georgymoiseev@georgymoiseev-HP-ProBook-440-G5:~/Development/sandbox/testrundir$ cartridge start -d
   • testrundir-stateboard... OK
   • testrundir.router... OK
   • testrundir.s2-master... OK
   • testrundir.s1-replica... OK
   • testrundir.s1-master... OK
   • testrundir.s2-replica... OK

georgymoiseev@georgymoiseev-HP-ProBook-440-G5:~/Development/sandbox/testrundir$ cartridge enter router
connected to testrundir.router
testrundir.router> 
^C
testrundir.router> 

georgymoiseev@georgymoiseev-HP-ProBook-440-G5:~/Development/sandbox/testrundir$ cartridge stop
   • testrundir-stateboard... OK
   • testrundir.s2-replica... OK
   • testrundir.router... OK
   • testrundir.s1-master... OK
   • testrundir.s1-replica... OK
   • testrundir.s2-master... OK
```

DifferentialOrange:
Oh, well, it was already discussed in folded comments.

cli/project/files.go
DifferentialOrange: No longer relevant

DifferentialOrange:
> I suggest in commit title to use the "names" of the directory(`data`, `log`, `run`) instead of paths.

Changed

DifferentialOrange:
@Totktonada , @LeonidVas , GitHub do not allow me to rerequest review in status "Awaiting requested review", so I'll ping you with this

Totktonada (approved):
It looks meaningful (and I can't say more here). No objections from me.

LeonidVas (approved):

PR #616, opened 2021-08-04
Doc: add description for `--spec` flag

### LINKED ISSUE #615: Doc: add description for `--spec` flag in README
We have a flag for `cartridge build` and `cartridge pack` to pass rockspec file
to `tarantoolctl build` like: _cartridge build --spec app-name-scm-2.rockspec_.
But there is no info about this feature in docs.

AnaNek:
Closes #615


README.rst
```diff
@@ -778,6 +782,8 @@ The options (``[flags]``) are as follows:
 * ``--unit-params-file`` (used for ``rpm`` and ``deb`` packages) is is the path to the file that contains
   unit parameters for the ``systemd`` unit files.
 
+* ``--spec`` is the path to rockspec to use for current build.
```
mRrvz, line 785: I think you need to place this flag in the same order as in ``cartidge pack --help`` (after ``--name`` flag)

README.rst
```diff
@@ -355,6 +355,13 @@ To build your application locally (for local testing), say this in any directory
 
     cartridge build [PATH] [flags]
 
+The following options (``[flags]``) are supported:
+
+* ``--spec`` is the path to rockspec to use for current build.
+
+**Note**: Please use rockspec name in lower case when you build
+your application with ``--spec`` flag.
+
```
mRrvz, line 364: Maybe also add this information to ``--help`` command? Something like this:

```
Path to rockspec to use for current build.
Should be in lowercase.
```

And remove **Note** from the docs, but add into description:

```
* ``--spec`` is the path to rockspec to use for current build. Should be in lowercase.
```

README.rst
```diff
@@ -778,6 +787,9 @@ The options (``[flags]``) are as follows:
 * ``--unit-params-file`` (used for ``rpm`` and ``deb`` packages) is is the path to the file that contains
   unit parameters for the ``systemd`` unit files.
 
+**Note**: Please use rockspec name in lower case when you pack
+your application with ``--spec`` flag.
+
```
mRrvz, line 792: I don't like that this note is far from the description of the flag. See comment above.

README.rst
AnaNek: About `--help` flag: this flag is supported by all commands but there is no description for this flag for any command in doc. My point is if we want to add such information this should be applied to description for each command and looks like it is task for another patch.

mRrvz: Agree. We need a separate patch.

mRrvz (approved):
LGTM. Please don't forget to create another patch with changes we talked about above.

PR #618, opened 2021-08-18
Replace x/sys for go 1.17

roopakv:
This updates x/sys for the sdk and therefor adds support for go 1.17. More info here Homebrew/homebrew-core#83413

I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Documentation

Closes #???

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -42,3 +42,5 @@ require (
 replace github.com/docker/docker/internal/testutil => gotest.tools/v3 v3.0.0
 
 replace github.com/c-bata/go-prompt => github.com/tarantool/go-prompt v0.2.6-tarantool
+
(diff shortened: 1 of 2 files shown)


mRrvz (approved):

mRrvz:
@roopakv please answer the question above

roopakv:
@Totktonada

replace because x/sys is a transitive dependency so require wont work, replace works for nested deps.

cannot do x >= with go. The team will need to remove this later

This is not a problem on homebrew, but a problem in x/sys and go 1.17. cartridge-cli needs this change to work with go 1.17

Totktonada:
Thank you for the clarification. I don't mind the change.

PR #629, opened 2021-09-02
Adapt Cartridge CLI to new release policy

### LINKED ISSUE #619: Adapt cartridge cli to new release policy
https://github.com/tarantool/cartridge-cli/blob/e9e99f68ccd90b5ad1cc0c435658c1add3a5c2e5/cli/common/tarantool.go#L21

This one leads to "⨯ Failed to get Tarantool context: Failed to get Tarantool version: Failed to match Tarantool version"

mRrvz:
Reworked version by @DifferentialOrange 

Add struct type for Tarantool version and parsing utility. Struct can be
used to build dependency requirements instead of version string.

Add Tarantool dependency to rpm or deb package only if it is not set up
by user explicitly with --deps or --deps-file. Use version struct
to build minimal and maximal requirements for packages based on new
version policy. Old version policy (2.8 and less) cases are considered
separately. Version struct parsing is strict so "lazy" versions without
number of commits since tag and commit hash (like "2.8.0") are no longer
supported since real Tarantool packages always have them. Use
struct-based approach in integration tests, migrate them to new version
policy.

Tarantool download through the [installer](https://tarantool.io/installer.sh) has changed along with
the announcement of the new release policy. cartridge pack docker
command uses an installer to install Tarantool.

Closes #619

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        tarantool-version: ["1.10", "2.7"]
+        tarantool-version: ["1.10", "2.8", "2.x-latest"]
       fail-fast: false
     steps:
       - uses: actions/checkout@master
@@ -44,9 +44,16 @@ jobs:
           git config --global user.name "Tar Antool"
 
       - name: Install Tarantool
+        if: matrix.tarantool-version != '2.x-latest'
         uses: tarantool/setup-tarantool@v1
         with:
           tarantool-version: '${{ matrix.tarantool-version }}'
+
+      - name: Install latest pre-release Tarantool 2.x
+        if: matrix.tarantool-version == '2.x-latest'
+        run: |
+          curl -L https://tarantool.io/pre-release/2/installer.sh | bash
+          sudo apt-get -y install tarantool
 
       - name: Cache docker images
         uses: satackey/action-docker-layer-caching@v0.0.11
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Removed unnecessary flags (``--rocks``, ``--project-path``) from ``cartridge help`` command.
 - Fixed project build with capital letters in the project name.
 - Fixed display of Docker image pull (``cartridge pack`` command with ``--verbose`` flag).
+- Added support for new Tarantool release policy.
 
 ### Added
 
--- a/cli/commands/pack.go
+++ b/cli/commands/pack.go
@@ -57,22 +57,62 @@ func init() {
 	packCmd.Flags().StringVar(&ctx.Pack.SystemdUnitParamsPath, "unit-params-file", "", UnitParamsFileUsage)
 }
 
+// isExplicitTarantoolDeps returns true if Tarantool was set up by user as a dependency
+// with --deps or --deps-file, false otherwise.
+func isExplicitTarantoolDeps(deps common.PackDependencies) bool {
+	for _, v := range deps {
+		if v.Name == "tarantool" {
+			return true
+		}
+	}
+
+	return false
+}
+
+// Tarantool dependence is added to rpm and deb packages deps, if it
+// wasn't set up explicitly. Dependency conditions is chosen based on
+// tarantool version used in cartridge-cli environment. Since development
+// builds and entrypoint builds normally are not available in package repos,
+// they are not supported as rpm/deb dependency. Minimal required version
+// is environment tarantool version, maximum is next tarantool major version.
+// Both modern and <= 2.8 version policies are supported.
 func addTarantoolDepIfNeeded(ctx *context.Ctx) error {
-	if ctx.Tarantool.TarantoolIsEnterprise || !(ctx.Pack.Type == pack.RpmType || ctx.Pack.Type == pack.DebType) {
+	var version common.TarantoolVersion
+	var minVersion, maxVersion string
+	var err error
+
+	if isExplicitTarantoolDeps(ctx.Pack.Deps) {
 		return nil
 	}
 
-	var tarantoolVersion string
-	if ctx.Pack.Type == pack.RpmType {
-		tarantoolVersion = strings.SplitN(ctx.Tarantool.TarantoolVersion, "-", 2)[0]
-	} else if ctx.Pack.Type == pack.DebType {
-		tarantoolVersion = ctx.Tarantool.TarantoolVersion
+	if ctx.Tarantool.TarantoolIsEnterprise {
+		return nil
 	}
 
-	if err := ctx.Pack.Deps.AddTarantool(tarantoolVersion); err != nil {
-		return fmt.Errorf("Failed to get tarantool dependency: %s", err)
+	if (ctx.Pack.Type != pack.RpmType) && (ctx.Pack.Type != pack.DebType) {
+		return nil
 	}
 
+	if version, err = common.ParseTarantoolVersion(ctx.Tarantool.TarantoolVersion); err != nil {
+		return err
+	}
+
+	if version.IsDevelopmentBuild {
+		return fmt.Errorf("Development build found. If you want to use Tarantool development build" +
+			"as a dependency, set it up explicitly with --deps or --deps-file")
+	}
+
+	if version.TagSuffix == "entrypoint" {
+		return fmt.Errorf("Entrypoint build found. If you want to use Tarantool entrypoint build" +
+			"as a dependency, set it up explicitly with --deps or --deps-file")
+	}
+
+	if minVersion, err = common.GetMinimalRequiredVersion(version); err != nil {
+		return err
+	}
+	maxVersion = common.GetNextMajorVersion(version)
+
+	ctx.Pack.Deps.AddTarantool(minVersion, maxVersion)
 	return nil
 }
 
--- a/cli/common/tarantool.go
+++ b/cli/common/tarantool.go
@@ -14,11 +14,100 @@ import (
 )
 
 var (
+	// Used for shallow validation of a version string.
 	tarantoolVersionRegexp *regexp.Regexp
+	// Used for thorough validation of a version string and to extract string version to a struct.
+	tarantoolVersionFullRegexp *regexp.Regexp
 )
 
 func init() {
 	tarantoolVersionRegexp = regexp.MustCompile(`\d+\.\d+\.\d+[-\w]*`)
+	// Part 1 is semVer X.Y.Z ,
+	// part 2 (optional) is a tag suffix for pre-release,
+	// part 3 is number of commits since tag and commit hash,
+	// part 4 (optional) is enterprise suffix,
+	// part 5 (optional) is development build suffix.
+	tarantoolVersionFullRegexp = regexp.MustCompile(
+		`^(?P<Major>\d+)\.(?P<Minor>\d+)?\.(?P<Patch>\d+)` +
+			`(?:-(?P<TagSuffix>alpha\d+|beta\d+|rc\d+|entrypoint))?` +
+			`-(?P<CommitsSinceTag>\d+)-(?P<CommitHashId>g[0-9a-f]+)` +
+			`(?:-(?P<EnterpriseSDKRevision>r\d+)(?:-(?P<EnterpriseIsOnMacOS>macos))?)?` +
+			`(?:-(?P<IsDevelopmentBuild>dev))?$`)
+}
+
+// findNamedMatches processes regexp with named capture groups
+// and transforms output to a map. If capture group is optional
+// and was not found, map value is empty string.
+func findNamedMatches(regex *regexp.Regexp, str string) map[string]string {
+	match := regex.FindStringSubmatch(str)
+
+	results := map[string]string{}
+	for i, name := range match {
+		// i == 0 is input string.
+		if i != 0 {
+			results[regex.SubexpNames()[i]] = name
+		}
+	}
+	return results
+}
+
+type TarantoolVersion struct {
+	Major                 uint64
+	Minor                 uint64
+	Patch                 uint64
+	TagSuffix             string
+	CommitsSinceTag       uint64
+	CommitHashId          string
+	EnterpriseSDKRevision string
+	EnterpriseIsOnMacOS   bool
+	IsDevelopmentBuild    bool
+}
+
+func atoiUint64(str string) (uint64, error) {
+	res, err := strconv.ParseUint(str, 10, 32)
+	if err != nil {
+		return 0, fmt.Errorf("Failed to parse Tarantool version: cast to number error (%s)", err)
+	}
+	return res, nil
+}
+
+// ParseTarantoolVersion extracts Tarantool version string to a TarantoolVersion struct.
+func ParseTarantoolVersion(version string) (TarantoolVersion, error) {
+	var result TarantoolVersion
+	var err error
+
+	matches := findNamedMatches(tarantoolVersionFullRegexp, version)
+	if len(matches) == 0 {
+		return result, fmt.Errorf("Failed to parse Tarantool version: format is not valid")
+	}
+
+	if result.Major, err = atoiUint64(matches["Major"]); err != nil {
+		return result, err
+	}
+
+	if result.Minor, err = atoiUint64(matches["Minor"]); err != nil {
+		return result, err
+	}
+
+	if result.Patch, err = atoiUint64(matches["Patch"]); err != nil {
+		return result, err
+	}
+
+	result.TagSuffix = matches["TagSuffix"]
+
+	if result.CommitsSinceTag, err = atoiUint64(matches["CommitsSinceTag"]); err != nil {
+		return result, err
+	}
+
+	result.CommitHashId = matches["CommitHashId"]
+
+	result.EnterpriseSDKRevision = matches["EnterpriseSDKRevision"]
+
+	result.EnterpriseIsOnMacOS = (matches["EnterpriseIsOnMacOS"] != "")
+
+	result.IsDevelopmentBuild = (matches["IsDevelopmentBuild"] != "")
+
+	return result, nil
 }
 
 // GetTarantoolDir returns Tarantool executable directory
--- a/cli/common/tarantool.go
+++ b/cli/common/tarantool.go
@@ -170,16 +170,32 @@ func GetMajorMinorVersion(version string) string {
 	return majorMinorVersion
 }
 
-// GetNextMajorVersion computes next major version for a given one.
-// For example, for 1.10.3 it's 2
-func GetNextMajorVersion(versionStr string) (string, error) {
-	version, err := goVersion.NewSemver(versionStr)
-	if err != nil {
-		return "", fmt.Errorf("Failed to parse Tarantool version: %s", err)
+// GetMinimalRequiredVersion computes minimal required Tarantool version for a package (rpm, deb).
+func GetMinimalRequiredVersion(ver TarantoolVersion) (string, error) {
+	// Old-style package version policy allowed X.Y.Z-N versions for N > 0 .
+	if (ver.Major == 2 && ver.Minor <= 8) || (ver.Major < 2) {
+		return fmt.Sprintf("%d.%d.%d.%d", ver.Major, ver.Minor, ver.Patch, ver.CommitsSinceTag), nil
 	}
 
-	major := version.Segments()[0]
-	return strconv.Itoa(major + 1), nil
+	if ver.IsDevelopmentBuild {
+		return "", fmt.Errorf("Can't compute minimal required version for a development build")
+	}
+
+	if ver.TagSuffix == "entrypoint" {
+		return "", fmt.Errorf("Can't compute minimal required version for an entrypoint build")
+	}
+
+	if ver.TagSuffix != "" {
+		return fmt.Sprintf("%d.%d.%d~%s", ver.Major, ver.Minor, ver.Patch, ver.TagSuffix), nil
+	}
+
+	return fmt.Sprintf("%d.%d.%d", ver.Major, ver.Minor, ver.Patch), nil
+}
+
+// GetNextMajorVersion computes next Major version for a given one.
+// For example, for 1.10.3 it's 2 .
+func GetNextMajorVersion(ver TarantoolVersion) string {
+	return strconv.Itoa(int(ver.Major) + 1)
 }
 
 func GetCartridgeVersionStr(conn *connector.Conn) (string, error) {
--- a/cli/common/utils.go
+++ b/cli/common/utils.go
@@ -34,28 +34,20 @@ type PackDependency struct {
 
 type PackDependencies []PackDependency
 
-func (deps *PackDependencies) AddTarantool(tarantoolVersion string) error {
-	tarantoolMinVersion := tarantoolVersion
-	tarantoolMaxVersion, err := GetNextMajorVersion(tarantoolMinVersion)
-	if err != nil {
-		return fmt.Errorf("Failed to get next major version of Tarantool %s", err)
-	}
-
+func (deps *PackDependencies) AddTarantool(minVersion, maxVersion string) {
 	*deps = append(*deps, PackDependency{
 		Name: "tarantool",
 		Relations: []DepRelation{
 			{
 				Relation: ">=",
-				Version:  tarantoolMinVersion,
+				Version:  minVersion,
 			},
 			{
 				Relation: "<",
-				Version:  tarantoolMaxVersion,
+				Version:  maxVersion,
 			},
 		},
 	})
-
-	return nil
 }
 
 var bufSize int64 = 10000
--- a/cli/project/dockerfiles.go
+++ b/cli/project/dockerfiles.go
@@ -21,7 +21,10 @@ func init() {
 }
 
 type opensourseCtx struct {
-	MajorMinorVersion string
+	// Type is "release" or "pre-release".
+	Type string
+	// Version is "<Major>.<Minor>" for <= 2.8, "<Major>" for newer versions.
+	Version string
 }
 
 type enterpriseCtx struct {
@@ -154,6 +157,7 @@ func CheckBaseDockerfile(dockerfilePath string) error {
 
 func getInstallTarantoolLayers(ctx *context.Ctx) (string, error) {
 	var installTarantoolLayers string
+	var version common.TarantoolVersion
 	var err error
 
 	if ctx.Tarantool.TarantoolIsEnterprise {
@@ -167,9 +171,16 @@ func getInstallTarantoolLayers(ctx *context.Ctx) (string, error) {
 
 	} else {
 		tmplStr := installTarantoolOpensourceLayers
+
+		version, err = common.ParseTarantoolVersion(ctx.Tarantool.TarantoolVersion)
+		if err != nil {
+			return "", err
+		}
+
 		installTarantoolLayers, err = templates.GetTemplatedStr(&tmplStr,
 			opensourseCtx{
-				MajorMinorVersion: common.GetMajorMinorVersion(ctx.Tarantool.TarantoolVersion),
+				Type:    getInstallerType(version),
+				Version: getVersionForTarantoolInstaller(version),
 			},
 		)
 
@@ -179,6 +190,22 @@ func getInstallTarantoolLayers(ctx *context.Ctx) (string, error) {
 	}
 
 	return installTarantoolLayers, nil
+}
+
+func getVersionForTarantoolInstaller(version common.TarantoolVersion) string {
+	if (version.Major == 2 && version.Minor <= 8) || version.Major < 2 {
+		return fmt.Sprintf("%d.%d", version.Major, version.Minor)
+	}
+
+	return fmt.Sprintf("%d", version.Major)
+}
+
+func getInstallerType(version common.TarantoolVersion) string {
+	if version.TagSuffix != "" {
+		return "pre-release"
+	}
+
+	return "release"
 }
 
 const (
@@ -224,7 +251,7 @@ ENV TARANTOOL_INSTANCE_NAME=default
 `
 
 	installTarantoolOpensourceLayers = `### Install opensource Tarantool
-RUN curl -L https://tarantool.io/installer.sh | VER={{ .MajorMinorVersion }} bash \
+RUN curl -L https://tarantool.io/installer.sh | VER={{ .Version }} bash -s -- --type {{ .Type }} \
     && yum -y install tarantool-devel
 `
 
--- a/test/integration/pack/test_pack.py
+++ b/test/integration/pack/test_pack.py
@@ -21,8 +21,8 @@ from utils import (Archive, assert_dependencies_deb, assert_dependencies_rpm,
                    check_param_in_unit_files, clear_project_rocks_cache,
                    extract_app_files, extract_deb, extract_rpm, find_archive,
                    get_rocks_cache_path, get_rockspec_path, recursive_listdir,
-                   run_command_and_get_output, tarantool_enterprise_is_used,
-                   tarantool_version, validate_version_file)
+                   run_command_and_get_output, tarantool_dict_version,
+                   tarantool_enterprise_is_used, validate_version_file)
 
 
 # ########
@@ -99,13 +99,19 @@ def deb_archive(cartridge_cmd, tmpdir, light_project, request):
 
 @pytest.fixture(scope="session")
 def tarantool_versions():
-    min_deb_version = re.findall(r'\d+\.\d+\.\d+-\d+-\S+', tarantool_version())[0]
-    max_deb_version = str(int(re.findall(r'\d+', tarantool_version())[0]) + 1)
-    min_rpm_version = re.findall(r'\d+\.\d+\.\d+', tarantool_version())[0]
-    max_rpm_version = max_deb_version  # Their format is the same
+    ver = tarantool_dict_version()
+    if (ver['Major'] == 2 and ver['Minor'] <= 8) or (ver['Major'] < 2):
+        min_version = f"{ver['Major']}.{ver['Minor']}.{ver['Patch']}.{ver['CommitsSinceTag']}"
+    else:
+        if ver['TagSuffix'] is not None:
+            min_version = f"{ver['Major']}.{ver['Minor']}.{ver['Patch']}~{ver['TagSuffix']}"
+        else:
+            min_version = f"{ver['Major']}.{ver['Minor']}.{ver['Patch']}"
 
-    return {"min": {"deb": min_deb_version, "rpm": min_rpm_version},
-            "max": {"deb": max_deb_version, "rpm": max_rpm_version}}
+    max_version = f"{ver['Major'] + 1}"
+
+    return {"min": {"deb": min_version, "rpm": min_version},
+            "max": {"deb": max_version, "rpm": max_version}}
 
 
 # #####
@@ -134,7 +140,7 @@ def test_tgz(tgz_archive, tmpdir):
         assert_filemodes(project, extract_dir)
 
 
-def test_rpm(rpm_archive, tmpdir):
+def test_rpm(rpm_archive, tmpdir, tarantool_versions):
     project = rpm_archive.project
 
     # archive files should be extracted to the empty directory
@@ -145,7 +151,7 @@ def test_rpm(rpm_archive, tmpdir):
     extract_rpm(rpm_archive.filepath, extract_dir)
 
     if not tarantool_enterprise_is_used():
-        assert_tarantool_dependency_rpm(rpm_archive.filepath)
+        assert_tarantool_dependency_rpm(rpm_archive.filepath, tarantool_versions)
 
     check_package_files(project, extract_dir)
     assert_files_mode_and_owner_rpm(project, rpm_archive.filepath)
@@ -158,7 +164,7 @@ def test_rpm(rpm_archive, tmpdir):
     assert process.returncode == 0, "RPM signature isn't correct"
 
 
-def test_deb(deb_archive, tmpdir):
+def test_deb(deb_archive, tmpdir, tarantool_versions):
     project = deb_archive.project
 
     # archive files should be extracted to the empty directory
@@ -191,7 +197,7 @@ def test_deb(deb_archive, tmpdir):
             assert os.path.exists(os.path.join(control_dir, filename))
 
         if not tarantool_enterprise_is_used():
-            assert_tarantool_dependency_deb(os.path.join(control_dir, 'control'))
+            assert_tarantool_dependency_deb(os.path.join(control_dir, 'control'), tarantool_versions)
 
         # check if postinst script set owners correctly
         with open(os.path.join(control_dir, 'postinst')) as postinst_script_file:
--- a/test/utils.py
+++ b/test/utils.py
@@ -532,19 +532,16 @@ def assert_dependencies_deb(filename, deps, tarantool_versions, tmpdir):
     assert all(dep in output for dep in deps)
 
 
-def assert_tarantool_dependency_deb(filename):
+def assert_tarantool_dependency_deb(filename, tarantool_versions):
     with open(filename) as control:
         control_info = control.read()
 
         depends_str = re.search('Depends: (.*)', control_info)
         assert depends_str is not None
 
-        min_version = re.findall(r'\d+\.\d+\.\d+-\d+-\S+', tarantool_version())[0]
-        max_version = str(int(re.findall(r'\d+', tarantool_version())[0]) + 1)
-
         deps = depends_str.group(1)
-        assert 'tarantool (>= {})'.format(min_version) in deps
-        assert 'tarantool (<< {})'.format(max_version) in deps
+        assert 'tarantool (>= {})'.format(tarantool_versions["min"]["deb"]) in deps
+        assert 'tarantool (<< {})'.format(tarantool_versions["max"]["deb"]) in deps
 
 
 def assert_dependencies_rpm(filename, deps, tarantool_versions):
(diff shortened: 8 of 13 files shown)


cli/common/utils.go
```diff
@@ -568,6 +572,37 @@ func ParseDependencies(rawDeps []string) (PackDependencies, error) {
 	return deps, nil
 }
 
+func getFormattedTarantoolVersion(version string, packType string) (string, error) {
```
olegrok, line 575: To be honest I'm not sure you need to parse version manually. Why don't you use go-version (https://github.com/tarantool/cartridge-cli/blob/master/go.mod#L21)?

LeonidVas (commented):
Hi! Thank you for the patch. Generally LGTM.
See several comments bellow:
* (`Added nightly-build for CE.`) IMHO, we should test the product only on release versions of tarantool (LTS and the latest), because if the current tarantool master is used for testing, our tests will fails when tarantool is broken (this is not such a rare story). In this case, product development will be blocked until the bug is fixed in tarantool, because we can't push to the master if any tests fail.

cli/commands/pack.go
```diff
@@ -62,17 +62,17 @@ func addTarantoolDepIfNeeded(ctx *context.Ctx) error {
...
 	}
 
-	if err := ctx.Pack.Deps.AddTarantool(tarantoolVersion); err != nil {
-		return fmt.Errorf("Failed to get tarantool dependency: %s", err)
+	tarantoolMaxVersion, err := common.GetNextMajorVersion(ctx.Tarantool.TarantoolVersion)
+	if err != nil {
+		return fmt.Errorf("Failed to get maximum Tarantool version: %s", err)
 	}
 
+	ctx.Pack.Deps.AddTarantool(tarantoolMinVersion, tarantoolMaxVersion)
```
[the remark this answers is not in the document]
mRrvz: I do not quite understand what exactly you are talking about. Can you explain please?

mRrvz:
@LeonidVas 

>  IMHO, we should test the product only on release versions of tarantool (LTS and the latest), because if the current tarantool master is used for testing, our tests will fails when tarantool is broken (this is not such a rare story). In this case, product development will be blocked until the bug is fixed in tarantool, because we can't push to the master if any tests fail.

Agree. Remove this from the patch.

cli/commands/pack.go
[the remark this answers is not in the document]
Totktonada: AFAIU, Leonid means that before the patch `AddTarantool()` accepts a current tarantool version (as opposite to a border / borders), while after the patch it accepts borders.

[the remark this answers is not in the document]
Totktonada: In other words, `GetNextMajorVersion()` goes from the `AddTarantool()` helper to this function.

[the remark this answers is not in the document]
LeonidVas: Yep, why can't we leave the logic with `GetNextMajorVersion` in the `AddTarantool` function.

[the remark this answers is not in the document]
mRrvz: We can move the call `GetNextMajorVersion` to `AddTarantool`, but:

We can't call `pack.GetTarantoolMinVersion` in `AddTarantool()` function (`common` package will import `pack` package -> cycle imports), also we can't describe `GetTarantoolMinVersion` in `common` package, because we need constant [`pack.RpmType`](https://github.com/tarantool/cartridge-cli/blob/fdd349b4f99cdbb4f3f6b91300d97183ff1859b3/cli/pack/common.go#L62). We can't move `GetTarantoolMinVersion` to `AddTarantool()`, and in this case, we anyway have to pass two arguments `tarantoolMinVersion` and  `ctx.Tarantool.TarantoolVersion` (bcs `GetNextMajorVersion` will work only with him - if we pass tarantoolMinVersion, we will get error). And this looks ugly.

LeonidVas (approved):
LGTM.
Please ask before pushing the patch whether Sasha will review.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Removed unnecessary flags (``--rocks``, ``--project-path``) from ``cartridge help`` command.
 - Fixed project build with capital letters in the project name.
 - Fixed display of Docker image pull (``cartridge pack`` command with ``--verbose`` flag).
+- Added support for new Tarantool versioned policy.
```
LeonidVas, line 17: `Tarantool versioned policy` -> `Tarantool release policy`

mRrvz: Fixed

Totktonada:
I'll repeat my feedback that I sent privately Sep, 13.

1. The change of the logic around generating dependency on tarantool in RPM / Deb packages could be extracted into its own commit.
2. The change of the `installer.sh` URL could be extracted into its own commit.
3. We should support `<...>/release/installer.sh`, not only `<...>/pre-release/installer.sh`. I guess the logic could be the following: if a user works on a pre-release, let's enable pre-release. Otherwise enable the release repository (if tarantool version is 2.10.0 or newer).
4. For me it looks meaningful to have a job in CI that will test this new code. Sure, CI may fail on pre-releases, but it is good: we'll know about problems before a release.

Regarding 1-2: just to ease review and future reading of the change. Now it is a bit complicated.

mRrvz:
@Totktonada @LeonidVas Thanks for the review! All issues resolved.

cli/project/dockerfiles.go
```diff
@@ -181,11 +183,40 @@ func getInstallTarantoolLayers(ctx *context.Ctx) (string, error) {
...
+	if err != nil {
+		return ""
+	}
+
+	minor, err := strconv.Atoi(parts[1])
+	if err != nil {
+		return ""
+	}
+
+	if major >= 2 && minor > 8 {
```
Totktonada, line 199: 3.0.0 will not match with this condition.

cli/project/dockerfiles.go
```diff
@@ -181,11 +183,40 @@ func getInstallTarantoolLayers(ctx *context.Ctx) (string, error) {
...
+
 const (
 	DefaultBaseBuildDockerfile   = "Dockerfile.build.cartridge"
 	DefaultBaseRuntimeDockerfile = "Dockerfile.cartridge"
 
-	containerSDKPath = "/usr/share/tarantool/sdk"
+	containerSDKPath              = "/usr/share/tarantool/sdk"
+	defaultOpenSourceInstaller    = "https://tarantool.io/installer.sh"
+	preReleaseOpenSourceInstaller = "https://tarantool.io/pre-release/2/installer.sh"
+	releaseOpenSourceInstaller    = "https://tarantool.io/release/2/installer.sh"
```
Totktonada, line 220: We can use the first URL for all cases:

1. `tarantool <= 2.8`: `VER=X.Y installer.sh --type release|live`.
2. `tarantool > 2.8`: `VER=X installer.sh --type release|pre-release`.

I think it'll simplify the logic in the code.

test/utils.py
```diff
@@ -1394,3 +1394,19 @@ def get_response_data(response):
...
+
+def get_tarantool_installer_cmd(package_manager):
+    short_version = tarantool_short_version()
+    major, minor = short_version.split(".")
+
+    if int(major) >= 2 and int(minor) > 8:
+        installer = "https://tarantool.io/release/2/installer.sh"
+        is_prerelease = any(postfix in tarantool_version() for postfix in ["-alpha", "-beta", "-rcc"])
+        if is_prerelease:
+            installer = "https://tarantool.io/pre-release/2/installer.sh"
```
Totktonada, line 1423: Comments above (3.0.0 and single URL) are applicable here too.

cli/project/dockerfiles.go
mRrvz: Fixed

cli/project/dockerfiles.go
mRrvz: Fixed

test/utils.py
mRrvz: Fixed

.github/workflows/tests.yml
```diff
@@ -44,10 +44,17 @@ jobs:
...
       - name: Install Tarantool
+        if: matrix.tarantool-version != '2.10'
         uses: tarantool/setup-tarantool@v1
         with:
           tarantool-version: '${{ matrix.tarantool-version }}'
 
+      - name: Install Tarantool 2.10
+        if: matrix.tarantool-version == '2.10'
+        run: |
+          curl -L https://tarantool.io/IMYxqhi/pre-release/2/installer.sh | bash
```
Totktonada, line 55: The hash could be safely dropped. AFAIU it is for analytics: this hash is reported then by the feedback daemon together with features usage report. It is undesirable to fill the statistics from CI, because its purpose is to better understand popularity of tarantool versions and features across our customers.

cli/project/dockerfiles.go
```diff
@@ -181,6 +184,42 @@ func getInstallTarantoolLayers(ctx *context.Ctx) (string, error) {
...
+	minor, _ := strconv.Atoi(parts[1])
+
+	if major > 2 || (major >= 2 && minor > 8) {
+		return fmt.Sprintf("%d", major)
+	}
+
+	return fmt.Sprintf("%d.%d", major, minor)
+}
+
+func getOpenSourceInstallerLink(version string) string {
```
Totktonada, line 199: The function returns a repository type, not a link.

.github/workflows/tests.yml
mRrvz: Fixed

cli/project/dockerfiles.go
mRrvz: Fixed

cli/pack/common.go
```diff
@@ -30,6 +31,40 @@ var (
...
+	}
+
+	// RPM format
+	if packType == RpmType {
+		return strings.SplitN(version, "-", 2)[0], nil
+	}
+
+	// Deb format
+	return version, nil
+}
```
Totktonada, line 66: Let me describe the idea behind the function. A developer working on an application on its Linux or Mac OS host. The develover verifies the application on locally installed tarantool version and want to be sure that all required tarantool functionality will be available, when a client installs the application from an RPM or Deb package. We assume that the typical wish is to ensure that tarantool version on the target system is not older than one installed on the developer host. Correct?

The function should return a string (based on `tarantool --version` output), which is suitable to place into the `Requires: tarantool >= {}` the RPM Spec of the application and `Depends: tarantool (>= {})` in the `debian/control` file. Right?

We can construct a table with variants of `tarantool --version` output and RPM/Deb package versions (upstream versions, without distribution's `-1`, `-1.el7` and so on).

| Tarantool version/build | tarantool --version                      | RPM version             | Deb version                 |
| ----------------------- | ---------------------------------------- | ----------------------- | --------------------------- |
| <= 2.8\*                | `X.Y.Z-N-g<commit id>`                   | `X.Y.Z.N`               | `X.Y.Z.N.g<commit id>`      |
| >= 2.10\* (release)     | `X.Y.Z-0-g<commit id>`                   | `X.Y.Z`                 | `X.Y.Z`                     |
| >= 2.10\* (pre-release) | `X.Y.Z-rcN-0-g<commit id>`[^1]           | `X.Y.Z~rcN`[^1]         | `X.Y.Z~rcN`[^1]             |
| >= 2.10\* (nightly)     | `X.Y.Z[-rcN]-N-g<commit id>-dev`[^2][^3] | `X.Y.Z[~rcN].N.dev`[^2] | `X.Y.Z[~rcN].N.dev`[^2][^4] |

[^1]: The same for `alpha` and `beta`.
[^2]: The same for `alpha`, `beta` and the special `entrypoint` mark.
[^3]: It lacks `-dev` suffix until https://github.com/tarantool/tarantool/issues/6183 will be fixed.
[^4]: It also may be meaningful to add `<commit id>` into developer builds of Debian packages. I'm not sure but AFAIR unreleased versions usually have it (unlike releases). The question is to be investigated.

When Tarantool Enterprise is used on a developer host, `cartridge pack` copies tarantool's binary file into the package and don't set the dependency on tarantool. Anyway, if we want to keep `tarantool --version` parsing code universal, it would be good to support parsing of Tarantool Enterprise versions as well. The version is constructed as `<opensource_version>-rM[-macos]` (the `-macos` suffix is optional). The `<opensource_version>` is the `describe --long --abbrev=9 <commit id>` output. `M` is commits count from the sdk repository root to given commit (passed from the sdk repository to the tarantool enterprise build scripts).

<details>
<summary>The similar table for Tarantool Enterprise</summary>

| Tarantool version/build | tarantool --version                 | The same on Mac OS                            |
| ----------------------- | ----------------------------------- | --------------------------------------------- |
| <= 2.8\*                | `X.Y.Z-N-g<commit id>-rM`           | `X.Y.Z-N-g<commit id>-rM-macos`               |
| >= 2.10\* (release)     | `X.Y.Z-0-g<commit id>-rM`           | `X.Y.Z-0-g<commit id>-rM-macos`               |
| >= 2.10\* (pre-release) | `X.Y.Z-rcN-0-g<commit id>-rM`       | `X.Y.Z-rcN-0-g<commit id>-rM-macos`           |
| >= 2.10\* (nightly)     | `X.Y.Z[-rcN]-N-g<commit id>-rM`[^5] | `X.Y.Z[-rcN]-N-g<commit id>-rM-macos`[^5]     |

[^5]: There is no sense in `-dev` for tarantool enterprise, because there is `-rM`.

Notes:

* Tarantool Enterprise is NOT distributed as RPM/Deb package.
* Footnotes from the opensource table are applicable here.
</details>

Details about version relationship operators for RPM and Deb packages (sources: [1][1], [2][2], [3][3]):

1. We can use either upstream version specifier (like `X`, `X.Y`, `X.Y.Z`, `X.Y.Z~beta1`) or full package version (like `X.Y.Z-1`, `X.Y.Z.N-1`, `X.Y.Z-1.el7`).
2. The version is like a border, so we can use, say, `Requires: tarantool >= 1.10.3` despite the fact that tarantool (mostly) has four digit version.

[1]: http://ftp.rpm.org/max-rpm/s1-rpm-depend-manual-dependencies.html
[2]: https://www.debian.org/doc/debian-policy/ch-relationships.html
[3]: https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-version

I propose the following rules to form the `>=` dependency:

| Tarantool version/build    | tarantool --version                   | Dependency                       |
| -------------------------- | ------------------------------------- | -------------------------------- |
| <= 2.8\*                   | `X.Y.Z-N-g<commit id>`                | `>= X.Y.Z`                       |
| >= 2.10\* (release)        | `X.Y.Z-0-g<commit id>`                | `>= X.Y.Z`                       |
| >= 2.10\* (pre-release)    | `X.Y.Z-rcN-0-g<commit id>`            | `>= X.Y.Z~rcN`                   |
| >= 2.10\* (nightly, usual) | `X.Y.Z[-rcN]-N-g<commit id>-dev`      | require `--deps` / `--deps-file` |
| >= 2.10\* (nightly, early) | `X.Y.Z-entrypoint-N-g<commit id>-dev` | require `--deps` / `--deps-file` |

The proposal is short: use three upstream version numbers (plus alpha/beta/rc suffix) to point a release. For release build: just copy `X.Y.Z`. For a pre-release: copy `X.Y.Z-rcN`.

For a developer build we can make a guess that everything that is necessary is in the latest (pre-)release (`X.Y.Z-N-g<commit id>-dev` -> `>= X.Y.Z`, `X.Y.Z-rcN-N-g<commit id>-dev` -> `X.Y.Z-rcN`). But I propose another behaviour, because there is a special case.

The corner case is the following: `X.Y.Z-entrypoint-N-g<commit id>-dev` -> `?`. There is no (pre-)release in this feature set, it is unclear whether there are previous bugfix releases. I propose to fail in the case if no tarantool dependency is given thought `--deps` or `--deps-file`. I think we should do that for all developer builds.

NB: Don't forget to re-check that if tarantool dependency is set via `--deps` / `--deps-file`, it will not be overwritten.

I think that it would be better to parse `tarantool --version` into a structure and form the dependency using the structure content. It would be much more easy to spot the idea rather than get it from code based on string split/join/replace operations.

The structure could be the following:

* Major
* Minor
* Patch
* PrereleaseSuffix (optional)
* Revision
* CommitId
* EnterpriseSdkRevision (optional)
* EnterpriseIsOnMacOS (optional)
* IsDeveloperBuild (optional)

So we'll construct `Major.Minor.Patch + (PrereleaseSuffix ? '~' + PrereleaseSuffix : '')`. In fact, we don't even need any custom logic for old / new policy, just process pre-release suffixes correctly.

I'm not insist on 'full parse and re-construction' approach, but currect code looks obscure for me. I'll share several points around current code.

----

Notes on current implementation:

1. It receives the `version` string and it is not clear whether it is full `tarantool --version` part after 'Tarantool' (or 'Tarantool Enterprise') or some part. In fact, it was changed recently.
2. It produces versions with different number of digits for RPM / Deb and I don't understand why.
3. It adds `-1` and it is not clear why.
4. It replaces a first `-` with `~`, but it'll break, say, `X.Y.Z-N-g<commit id>` (when you have no a pre-release suffix).
5. The `2.9.0-alpha` border is unclear for me. Whether it is sorted before or after `2.9.0`? Should not it have tilda instead of dash? Why `2.9` if we started the new policy from `2.10`?

----

Regarding testing.

I would add a unit test with different variants of `tarantool --version` output. See 6b89c1ffebbd39ec5f3d0e5e19a1200c5386239d for examples or, better, construct more real-like examples using the tables above.

test/utils.py
```diff
@@ -529,7 +529,8 @@ def assert_dependencies_deb(filename, deps, tarantool_versions, tmpdir):
 
     rc, output = run_command_and_get_output(cmd, cwd=tmpdir)
     assert rc == 0
-    assert all(dep in output for dep in deps)
+    # Remove hash from dependencies
+    assert all(dep.split('-')[0] in output for dep in deps)
```
Totktonada, line 533: See https://github.com/tarantool/cartridge-cli/pull/629/commits/ba99b8de9437e39d6c9d7f65e1139b7aed8311aa#r727489690. I think we should not spead the logic of costructing the minimal version across the code so much.

test/utils.py
```diff
@@ -566,7 +569,7 @@ def assert_dependencies_rpm(filename, deps, tarantool_versions):
         for i, dep in enumerate(deps):
             assert rpm.headers['requirename'][i].decode('ascii') == dep[0]
             assert rpm.headers['requireflags'][i] == dep[1]
-            assert rpm.headers['requireversion'][i].decode('ascii') == dep[2]
+            assert rpm.headers['requireversion'][i].decode('ascii').startswith(dep[2])
```
Totktonada, line 572: It looks irrelevant to tarantool versions. Or I miss something?

.github/workflows/tests.yml
```diff
@@ -44,10 +44,17 @@ jobs:
           git config --global user.name "Tar Antool"
 
       - name: Install Tarantool
+        if: matrix.tarantool-version != '2.10'
         uses: tarantool/setup-tarantool@v1
         with:
           tarantool-version: '${{ matrix.tarantool-version }}'
 
+      - name: Install Tarantool 2.10
+        if: matrix.tarantool-version == '2.10'
```
DifferentialOrange, line 53: If there will be Tarantool 2.11, it will be in pre-release installer, so I think `2.10` is not the best label for it.

DifferentialOrange:
Original version of PR: https://github.com/tarantool/cartridge-cli/tree/new-release-policy-support-old

DifferentialOrange:
I have pushed ready to review version of reworks (including commit history and messages). @Totktonada , @LeonidVas , it's time to re-review this one.

test/integration/pack/test_pack.py
```diff
@@ -99,13 +99,19 @@ def deb_archive(cartridge_cmd, tmpdir, light_project, request):
...
+        if ver['TagSuffix'] is not None:
+            min_version = f"{ver['Major']}.{ver['Minor']}.{ver['Patch']}~{ver['TagSuffix']}"
+        else:
+            min_version = f"{ver['Major']}.{ver['Minor']}.{ver['Patch']}"
+
+    max_version = f"{ver['Major'] + 1}"
 
-    return {"min": {"deb": min_deb_version, "rpm": min_rpm_version},
-            "max": {"deb": max_deb_version, "rpm": max_rpm_version}}
+    return {"min": {"deb": min_version, "rpm": min_version},
```
LeonidVas, line 113: Sorry. I don’t understand, rpm and deb versions have become equivalent?

cli/project/dockerfiles.go
```diff
@@ -21,7 +21,8 @@ func init() {
 }
 
 type opensourseCtx struct {
-	MajorMinorVersion string
+	Type    string
```
LeonidVas, line 24: Add descriptions for the fields.

LeonidVas (changes requested):
Hi! Thank you for the patch. If it's work fine I am generally ok with the code.
See several minor comments below.
Also check your changes with `go fmt`, as far as I can see, it has several comments to the code.

DifferentialOrange:
> Also check your changes with `go fmt`, as far as I can see, it has several comments to the code.

There is already a lint tool running with `mage lint` in CI that runs `go vet`. I'm not sure that it is interchangable. How should I use `go fmt`? Running `go fmt`, `go fmt *.go`, `go fmt .`, `go fmt */*/*.go` doesn't work (or everything is alright).

cli/common/tarantool.go
```diff
@@ -14,11 +14,92 @@ import (
...
+	tarantoolVersionRegexp      = regexp.MustCompile(`\d+\.\d+\.\d+[-\w]*`)
+	tarantoolVersionFullRegexp  = regexp.MustCompile(
+		`^(?P<Major>\d+)\.(?P<Minor>\d+)?\.(?P<Patch>\d+)` +       // SemVer X.Y.Z
+		`(?:-(?P<TagSuffix>alpha\d+|beta\d+|rc\d+|entrypoint))?` + // (optional) tag suffixes for pre-release
+		`-(?P<CommitsSinceTag>\d+)-(?P<CommitHashId>g[0-9a-z]+)` + // number of commits since tag and commit hash
+		`(?:-(?P<EnterpriseSDKRevision>r\d+)(?:-(?P<EnterpriseIsOnMacOS>macos))?)?` + // (optional) Enterprise suffixes
+		`(?:-(?P<IsDevelopmentBuild>dev))?$`)                      // (optional) Development build suffix
+}
+
+// findNamedMatches processes regexp with named capture groups
```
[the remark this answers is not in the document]
DifferentialOrange: Added where it seemed relevant.

cli/common/tarantool.go
```diff
@@ -14,11 +14,92 @@ import (
...
+func ParseTarantoolVersion(version string) (TarantoolVersion, error) {
+	var result TarantoolVersion
+	var err error
+
+	matches := findNamedMatches(tarantoolVersionFullRegexp, version)
+	if len(matches) == 0 {
+		return result, fmt.Errorf("Failed to parse Tarantool version: format is not valid")
+	}
+
+	if result.Major, err = atoiUint64(matches["Major"]); err != nil {
```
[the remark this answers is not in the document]
DifferentialOrange: Negative numbers can be parsed with `\d+`, but casting them to unsigned will fail.

cli/common/tarantool.go
```diff
@@ -14,11 +14,92 @@ import (
...
+	Patch                  uint64
+	TagSuffix              string
+	CommitsSinceTag        uint64
+	CommitHashId           string
+	EnterpriseSDKRevision  string
+	EnterpriseIsOnMacOS    bool
+	IsDevelopmentBuild     bool
+}
+
+func atoiUint64(str string) (uint64, error) {
```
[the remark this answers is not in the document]
DifferentialOrange: The main reason is error wrapper to make errors more readable. Considering "What errors we can get here?" see my answer below.

cli/commands/pack.go
```diff
@@ -57,22 +57,55 @@ func init() {
...
+	if isExplicitTarantoolDeps(ctx.Pack.Deps) {
 		return nil
 	}
 
-	var tarantoolVersion string
-	if ctx.Pack.Type == pack.RpmType {
-		tarantoolVersion = strings.SplitN(ctx.Tarantool.TarantoolVersion, "-", 2)[0]
-	} else if ctx.Pack.Type == pack.DebType {
-		tarantoolVersion = ctx.Tarantool.TarantoolVersion
+	if ctx.Tarantool.TarantoolIsEnterprise {
```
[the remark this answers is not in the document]
DifferentialOrange: I don't see any reasons why we should make code less readable and mix up two conditions that are of a different nature.

cli/common/tarantool.go
```diff
@@ -14,11 +14,92 @@ import (
 )
 
 var (
-	tarantoolVersionRegexp *regexp.Regexp
+	tarantoolVersionRegexp      *regexp.Regexp
+	tarantoolVersionFullRegexp  *regexp.Regexp
```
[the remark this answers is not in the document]
DifferentialOrange: Well, the main reason is to change only what we need to change to solve the issue and do not rework whole module (at least in this PR). I've tried to add some comments, but not sure it will be a satisfying answer.

cli/commands/pack.go
```diff
@@ -57,22 +57,55 @@ func init() {
...
+	for _, v := range deps {
+		if v.Name == "tarantool" {
+			return true
+		}
+	}
+
+	return false
+}
+
 func addTarantoolDepIfNeeded(ctx *context.Ctx) error {
```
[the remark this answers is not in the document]
DifferentialOrange: Added

cli/project/dockerfiles.go
DifferentialOrange: Added

cli/common/tarantool.go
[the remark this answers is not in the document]
LeonidVas: Ok.

cli/commands/pack.go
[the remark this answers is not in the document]
LeonidVas: Up to you.

cli/common/tarantool.go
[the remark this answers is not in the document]
LeonidVas: Ok. I think then we should create an issue with the `code health` tag to rework the whole module on new parsing when this PR is merged.

cli/common/tarantool.go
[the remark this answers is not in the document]
LeonidVas: JFYI - the relevant places is described [here](https://github.com/tarantool/tarantool/wiki/Code-review-procedure#code-style)
```
Check your comments, commit title, and even variable names to be grammatically correct. One of the most typical problems in the patches - typos. Enable a spell checker in your editor, if it has any. Start sentences from a capital letter, end with a dot. Everywhere - in the code, in the tests, in the commit message.
```

cli/common/tarantool.go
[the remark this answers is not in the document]
LeonidVas: OK.

cli/project/dockerfiles.go
```diff
@@ -21,7 +21,8 @@ func init() {
 }
 
 type opensourseCtx struct {
-	MajorMinorVersion string
+	Type    string // release or pre-release
```
LeonidVas, line 24: As for me, the style that puts the attribute description on a separate line like [here](https://github.com/tarantool/tvisor/blob/30ba24f8d095b64dad3665ecdc7bf47e91db6d67/supervisor/main.go#L21) is more readable in one hand and the same style is enforced in tarantool core on the other hand. So, if you don't object - do the same. Up to you, I do not insist.

As for me, the style that puts the attribute description on a separate line like [here] (https://github.com/tarantool/tvisor/blob/30ba24f8d095b64dad3665ecdc7bf47e91db6d67/supervisor/main.go#L21) is more readable in one hand ... and the same style is enforced on the other side of the tarantool core. So if you don't mind, do the same. At your discretion, I do not insist.

LeonidVas:
> > Also check your changes with `go fmt`, as far as I can see, it has several comments to the code.
> 
> There is already a lint tool running with `mage lint` in CI that runs `go vet`. I'm not sure that it is interchangable. How should I use `go fmt`? Running `go fmt`, `go fmt *.go`, `go fmt .`, `go fmt */*/*.go` doesn't work (or everything is alright).

```
┌─[✗]─[leonid@vasya-hp]─[~/work/mail/cartridge-cli]
└──╼ go fmt cli/commands/pack.go 
┌─[leonid@vasya-hp]─[~/work/mail/cartridge-cli]
└──╼ go fmt cli/project/dockerfiles.go cli/project/dockerfiles_test.go 
cli/project/dockerfiles.go
cli/project/dockerfiles_test.go
┌─[leonid@vasya-hp]─[~/work/mail/cartridge-cli]
└──╼ go fmt cli/common/tarantool.go cli/common/tarantool_test.go  cli/common/utils.go
cli/common/tarantool.go
cli/common/tarantool_test.go
```

Result diff:
```
diff --git a/cli/common/tarantool.go b/cli/common/tarantool.go
index 9477f06..b4e2095 100644
--- a/cli/common/tarantool.go
+++ b/cli/common/tarantool.go
@@ -15,19 +15,19 @@ import (
 
 var (
 	// Used for shallow validation of a version string.
-	tarantoolVersionRegexp      *regexp.Regexp
+	tarantoolVersionRegexp *regexp.Regexp
 	// Used for thorough validation of a version string and to extract string version to a struct.
-	tarantoolVersionFullRegexp  *regexp.Regexp
+	tarantoolVersionFullRegexp *regexp.Regexp
 )
 
 func init() {
-	tarantoolVersionRegexp      = regexp.MustCompile(`\d+\.\d+\.\d+[-\w]*`)
-	tarantoolVersionFullRegexp  = regexp.MustCompile(
-		`^(?P<Major>\d+)\.(?P<Minor>\d+)?\.(?P<Patch>\d+)` +       // SemVer X.Y.Z
-		`(?:-(?P<TagSuffix>alpha\d+|beta\d+|rc\d+|entrypoint))?` + // (optional) tag suffixes for pre-release
-		`-(?P<CommitsSinceTag>\d+)-(?P<CommitHashId>g[0-9a-z]+)` + // number of commits since tag and commit hash
-		`(?:-(?P<EnterpriseSDKRevision>r\d+)(?:-(?P<EnterpriseIsOnMacOS>macos))?)?` + // (optional) Enterprise suffixes
-		`(?:-(?P<IsDevelopmentBuild>dev))?$`)                      // (optional) Development build suffix
+	tarantoolVersionRegexp = regexp.MustCompile(`\d+\.\d+\.\d+[-\w]*`)
+	tarantoolVersionFullRegexp = regexp.MustCompile(
+		`^(?P<Major>\d+)\.(?P<Minor>\d+)?\.(?P<Patch>\d+)` + // SemVer X.Y.Z
+			`(?:-(?P<TagSuffix>alpha\d+|beta\d+|rc\d+|entrypoint))?` + // (optional) tag suffixes for pre-release
+			`-(?P<CommitsSinceTag>\d+)-(?P<CommitHashId>g[0-9a-z]+)` + // number of commits since tag and commit hash
+			`(?:-(?P<EnterpriseSDKRevision>r\d+)(?:-(?P<EnterpriseIsOnMacOS>macos))?)?` + // (optional) Enterprise suffixes
+			`(?:-(?P<IsDevelopmentBuild>dev))?$`) // (optional) Development build suffix
 }
 
 // findNamedMatches processes regexp with named capture groups
@@ -46,15 +46,15 @@ func findNamedMatches(regex *regexp.Regexp, str string) map[string]string {
 }
 
 type TarantoolVersion struct {
-	Major                  uint64
-	Minor                  uint64
-	Patch                  uint64
-	TagSuffix              string
-	CommitsSinceTag        uint64
-	CommitHashId           string
-	EnterpriseSDKRevision  string
-	EnterpriseIsOnMacOS    bool
-	IsDevelopmentBuild     bool
+	Major                 uint64
+	Minor                 uint64
+	Patch                 uint64
+	TagSuffix             string
+	CommitsSinceTag       uint64
+	CommitHashId          string
+	EnterpriseSDKRevision string
+	EnterpriseIsOnMacOS   bool
+	IsDevelopmentBuild    bool
 }
 
 func atoiUint64(str string) (uint64, error) {
@@ -188,7 +188,7 @@ func GetMinimalRequiredVersion(ver TarantoolVersion) (string, error) {
 
 // GetNextMajorVersion computes next Major version for a given one.
 // For example, for 1.10.3 it's 2.
-func GetNextMajorVersion(ver TarantoolVersion) (string) {
+func GetNextMajorVersion(ver TarantoolVersion) string {
 	return strconv.Itoa(int(ver.Major) + 1)
 }
 
diff --git a/cli/common/tarantool_test.go b/cli/common/tarantool_test.go
index 28ed8ee..ecafc2e 100644
--- a/cli/common/tarantool_test.go
+++ b/cli/common/tarantool_test.go
@@ -64,7 +64,7 @@ type returnValueParseTarantoolVersion struct {
 
 func TestParseTarantoolVersion(t *testing.T) {
 	assert := assert.New(t)
-	
+
 	testCases := make(map[string]returnValueParseTarantoolVersion)
 
 	testCases["1.10.11-0-g543e2a1ec0"] = returnValueParseTarantoolVersion{
@@ -156,7 +156,7 @@ func TestParseTarantoolVersion(t *testing.T) {
 		},
 		nil,
 	}
-	
+
 	testCases["2.8.1-121-g94ad1c2ee"] = returnValueParseTarantoolVersion{
 		TarantoolVersion{
 			Major:                 2,
@@ -284,7 +284,6 @@ func TestParseTarantoolVersion(t *testing.T) {
 	}
 }
 
-
 type returnValueGetMinimalRequiredVersion struct {
 	res string
 	err error
@@ -294,12 +293,12 @@ func TestGetTarantoolMinVersion(t *testing.T) {
 	assert := assert.New(t)
 
 	testCases := map[string]returnValueGetMinimalRequiredVersion{
-		"3.0.0-rc1-0-g7da4b1438":    { res: "3.0.0~rc1", err: nil },
-		"2.10.0-beta1-0-g7da4b1438": { res: "2.10.0~beta1", err: nil },
-		"2.10.0-0-g7da4b1438":       { res: "2.10.0", err: nil },
-		"2.9.0-alpha1-0-g7da4b1438": { res: "2.9.0~alpha1", err: nil },
-		"2.8.2-0-gfc96d10f5":        { res: "2.8.2.0", err: nil },
-		"1.10.11-21-g543e2a1ec0":    { res: "1.10.11.21", err: nil },
+		"3.0.0-rc1-0-g7da4b1438":    {res: "3.0.0~rc1", err: nil},
+		"2.10.0-beta1-0-g7da4b1438": {res: "2.10.0~beta1", err: nil},
+		"2.10.0-0-g7da4b1438":       {res: "2.10.0", err: nil},
+		"2.9.0-alpha1-0-g7da4b1438": {res: "2.9.0~alpha1", err: nil},
+		"2.8.2-0-gfc96d10f5":        {res: "2.8.2.0", err: nil},
+		"1.10.11-21-g543e2a1ec0":    {res: "1.10.11.21", err: nil},
 
 		"2.10.1-entrypoint-0-gc2438eeb1": {
 			res: "",
diff --git a/cli/project/dockerfiles.go b/cli/project/dockerfiles.go
index d4f548a..d7609a2 100644
--- a/cli/project/dockerfiles.go
+++ b/cli/project/dockerfiles.go
@@ -191,7 +191,7 @@ func getInstallTarantoolLayers(ctx *context.Ctx) (string, error) {
 }
 
 func getVersionForTarantoolInstaller(version common.TarantoolVersion) string {
-	if (version.Major == 2 && version.Minor <= 8) ||  version.Major < 2 {
+	if (version.Major == 2 && version.Minor <= 8) || version.Major < 2 {
 		return fmt.Sprintf("%d.%d", version.Major, version.Minor)
 	}
 
diff --git a/cli/project/dockerfiles_test.go b/cli/project/dockerfiles_test.go
index 0e0a17f..49b6f3d 100644
--- a/cli/project/dockerfiles_test.go
+++ b/cli/project/dockerfiles_test.go
@@ -116,31 +116,31 @@ RUN curl -L https://tarantool.io/installer.sh | VER=1.10 bash -s -- --type relea
 	assert.Nil(err)
 	assert.Equal(expLayers, layers)
 
-    // Tarantool Opensource 2.10 pre-release
-    ctx.Tarantool.TarantoolIsEnterprise = false
-    ctx.Tarantool.TarantoolVersion = "2.10.0-beta1-0-g7da4b1438"
+	// Tarantool Opensource 2.10 pre-release
+	ctx.Tarantool.TarantoolIsEnterprise = false
+	ctx.Tarantool.TarantoolVersion = "2.10.0-beta1-0-g7da4b1438"
 
-    expLayers = `### Install opensource Tarantool
+	expLayers = `### Install opensource Tarantool
 RUN curl -L https://tarantool.io/installer.sh | VER=2 bash -s -- --type pre-release \
     && yum -y install tarantool-devel
 `
 
-    layers, err = getInstallTarantoolLayers(&ctx)
-    assert.Nil(err)
-    assert.Equal(expLayers, layers)
+	layers, err = getInstallTarantoolLayers(&ctx)
+	assert.Nil(err)
+	assert.Equal(expLayers, layers)
 
-    // Tarantool Opensource 2.10
-    ctx.Tarantool.TarantoolIsEnterprise = false
-    ctx.Tarantool.TarantoolVersion = "2.10.3-0-gb14387da4"
+	// Tarantool Opensource 2.10
+	ctx.Tarantool.TarantoolIsEnterprise = false
+	ctx.Tarantool.TarantoolVersion = "2.10.3-0-gb14387da4"
 
-    expLayers = `### Install opensource Tarantool
+	expLayers = `### Install opensource Tarantool
 RUN curl -L https://tarantool.io/installer.sh | VER=2 bash -s -- --type release \
     && yum -y install tarantool-devel
 `
 
-    layers, err = getInstallTarantoolLayers(&ctx)
-    assert.Nil(err)
-    assert.Equal(expLayers, layers)
+	layers, err = getInstallTarantoolLayers(&ctx)
+	assert.Nil(err)
+	assert.Equal(expLayers, layers)
 }
 
 func TestGetBuildImageDockerfileTemplateEnterprise(t *testing.T) {

```

In my case, this work is done automaticaly by the sublime plugin. I think the `lint` command must be updated to highlight things like this, or I'm missing something.

cli/common/tarantool.go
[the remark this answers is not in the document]
DifferentialOrange: By "relevant" I mean everything that is a complete sentence and not one-liner comments. I don't how philologically hardcore I need to be here.

cli/project/dockerfiles.go
DifferentialOrange: Okay, every comment is reworked to a complete sentence on a separate line.

cli/project/dockerfiles.go
DifferentialOrange: Reworked, see below

cli/common/tarantool.go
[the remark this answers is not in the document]
DifferentialOrange: Reworked, see below

DifferentialOrange:
> > > Also check your changes with `go fmt`, as far as I can see, it has several comments to the code.
> > 
> > 
> > There is already a lint tool running with `mage lint` in CI that runs `go vet`. I'm not sure that it is interchangable. How should I use `go fmt`? Running `go fmt`, `go fmt *.go`, `go fmt .`, `go fmt */*/*.go` doesn't work (or everything is alright).
> 
> ```
> ┌─[✗]─[leonid@vasya-hp]─[~/work/mail/cartridge-cli]
> └──╼ go fmt cli/commands/pack.go 
> ┌─[leonid@vasya-hp]─[~/work/mail/cartridge-cli]
> └──╼ go fmt cli/project/dockerfiles.go cli/project/dockerfiles_test.go 
> cli/project/dockerfiles.go
> cli/project/dockerfiles_test.go
> ┌─[leonid@vasya-hp]─[~/work/mail/cartridge-cli]
> └──╼ go fmt cli/common/tarantool.go cli/common/tarantool_test.go  cli/common/utils.go
> cli/common/tarantool.go
> cli/common/tarantool_test.go
> ```

Ok, thank you, I've got it. Run with lint, added all changes to corresponding commits.

cli/common/tarantool.go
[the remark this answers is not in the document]
DifferentialOrange: https://github.com/tarantool/cartridge-cli/issues/665

mRrvz:
> Ok, thank you, I've got it. Run with lint, added all changes to corresponding commits.

Are you coding in VIM?)) Popular code editors (for example vscode) contain extensions for golang that use `go fmt` when saving a file

DifferentialOrange:
> > Ok, thank you, I've got it. Run with lint, added all changes to corresponding commits.
> 
> Are you coding in VIM?)) Popular code editors (for example vscode) contain extensions for golang that use `go fmt` when saving a file

No, I use Sublime Text 3. But when I've tried to install some popular linter plugin, it didn't worked from the start and I haven't got a patience to fix it or install another at the time.

LeonidVas (approved):
LGTM.

cli/common/tarantool.go
```diff
@@ -14,11 +14,100 @@ import (
...
 	tarantoolVersionRegexp = regexp.MustCompile(`\d+\.\d+\.\d+[-\w]*`)
+	// Part 1 is semVer X.Y.Z ,
+	// part 2 (optional) is a tag suffix for pre-release,
+	// part 3 is number of commits since tag and commit hash,
+	// part 4 (optional) is enterprise suffix,
+	// part 5 (optional) is development build suffix.
+	tarantoolVersionFullRegexp = regexp.MustCompile(
+		`^(?P<Major>\d+)\.(?P<Minor>\d+)?\.(?P<Patch>\d+)` +
+			`(?:-(?P<TagSuffix>alpha\d+|beta\d+|rc\d+|entrypoint))?` +
+			`-(?P<CommitsSinceTag>\d+)-(?P<CommitHashId>g[0-9a-z]+)` +
```
Totktonada, line 33: `a-z` -> `a-f`. It is hexadecimal.

test/utils.py
```diff
@@ -1394,3 +1388,36 @@ def get_response_data(response):
...
+
+    return ver
+
+
+def tarantool_dict_version():
+    s = tarantool_version()
+    ver = parse_tarantool_version(s)
+
+    assert ver['TagSuffix'] != 'entrypoint'
+    assert ver['IsDevelopmentBuild'] is False
```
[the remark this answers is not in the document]
DifferentialOrange: I don't know, they never had worked for me locally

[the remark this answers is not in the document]
Totktonada: Meh. Okay, so it is out of scope here.

cli/common/tarantool.go
DifferentialOrange: Fixed

DifferentialOrange:
> The logic of changes looks okay for me.
> 
> I see that you're working on it now: let's finish and proceed.

I finished my work, just waited for CI to give me a green light. It isn't, but it does not seems like I have broken it with this PR.

PR #630, opened 2021-09-02
Fix displaying of pulling Docker image

### LINKED ISSUE #627: Verbose flag don't show images pulling logs
If some docker image is pulled then cartridge-cli only displays blank lines like here:

```
   • Building base image ***
Step 1/8 : FROM centos:8

























































 ---> 300e315adb2f
```

Expected:
```
   • Building base image ***
Step 1/8 : FROM centos:8
2d473b07cdd5: Downloading [===========================================>       ]  66.57MB/76.1MB
```

mRrvz:
Removed blank lines and replaced with correct  when 
pulling Docker image in ``cartridge pack`` command 
with ``--verbose`` flag.

Closes #627

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Make Tarantool version check less strict, support `X.Y.Z` notation.
 - Removed unnecessary flags (``--rocks``, ``--project-path``) from ``cartridge help`` command.
 - Fixed project build with capital letters in the project name.
+- Fixed display of Docker image pull (``cartridge pack`` command with ``--verbose`` flag).
 
 ### Added
 
--- a/cli/docker/build.go
+++ b/cli/docker/build.go
@@ -36,7 +36,6 @@ var readerSize = 4096
 
 func printBuildOutput(out io.Writer, body io.ReadCloser) error {
 	rd := bufio.NewReaderSize(body, readerSize)
-	var output map[string]interface{}
 	buf := bytes.Buffer{}
 
 Loop:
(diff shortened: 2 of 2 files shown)


LeonidVas (approved):
LGTM

cli/docker/build.go
```diff
@@ -80,6 +77,31 @@ Loop:
...
+	if outputMap["status"] != nil {
+		output = fmt.Sprintf("%s%s ", output, outputMap["status"])
+	}
+
+	if outputMap["progress"] != nil {
+		output = fmt.Sprintf("%s%s", output, outputMap["progress"])
+	}
+
+	return fmt.Sprintf("%s\n", output)
+}
```
Totktonada, line 103: Everything here looks strange for me. Maybe it is actually the strage API, I don't know. I don't find an information about the format of the return value. I'll ask several questions.

1. We ignore everything if there is the 'stream' field. Is it okay?
2. We don't include keys into the output (we output `<id>: `, not `id: <id>`). Is it okay?
3. We output all fields in different ways (`<id>: `, `<status> `, `<progress>`). Those fields form a single line?
4. It seems, we read the whole json response, so why not to just dump it as is (or dump with prettifying)? Maybe except 'stream', which, I guess, large and would be better to output separately.
5. You said that the response format has different variants. When you see one format and when another? From what it depends? Docker server version? Docker client version? Something else?

Anyway, if it gives more readable result, I don't mind.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Cartridge errors in the ``replicasets`` command are now more readable.
 - Make Tarantool version check less strict, support `X.Y.Z` notation.
 - Removed unnecessary flags (``--rocks``, ``--project-path``) from ``cartridge help`` command.
+- Fixed display of Docker image pull (``cartridge pack`` command with ``--verbose`` flag).
```
Totktonada, line 15: NB: Now it has a conflict with master.

mRrvz: Fixed

cli/docker/build.go
mRrvz: I made the print the same as in the ``docker pull`` command.

1. Yes. For some reason (I don't know why) the entire output is passed either in the "stream" field, or it needs to be assembled piece by piece from other fields.
2. Yes.
3. Yes. 
4. Because there are fields with information that we do not need (at least, it is not on the server when we do a docker pull)
5. Depends on the action being performed on the docker side. I do not know why.

Totktonada: I meet Alexey and now I got the logic. In brief:

* I missed that we process a stream of JSON responses (in a cycle), not just one JSON response.
* So now I understand: responses from `docker pull` have no 'stream' fied, so we process them differently.
* Those `docker pull` responses may or may not appear depending on whether the local image is the newest one.

Thank you for the clarification!

Totktonada (approved):
I didn't tested it, but the logic looks correct.

PR #637, opened 2021-09-23
create: replace vfs with embed

### LINKED ISSUE #631: Replace vfs with embed
Currently go provides native way to bundle static files in binary. Initially we used third-party vfs module (8b0753f4450a30d9e956bbf1869139a8c5eb3fad). Currently it's possible to replace it with built-in Go "embed" module.

https://pkg.go.dev/embed

mRrvz:
This patch must be accepted after https://github.com/golang/go/issues/43854

Closes #631


Totktonada:
Looks like a breaking change. Can you clarify, whether it discards support of some Go versions?

DifferentialOrange:
Maybe it should wait until Go 1.18 stable release, but I'll press review request button since it works

Totktonada (approved):
I have no objections here, but, it seems, we should wait until Go 1.18 will be released.

DifferentialOrange:
Do we include internal changes like this in CHANGELOG or should it only include changes that affect binary artifact user?

DifferentialOrange:
@LeonidVas it's relevant again

PR #639, opened 2021-10-06
dependency: update containerd version to 1.5.7

LeonidVas:
Fix vulnerability bug https://github.com/advisories/GHSA-c2h3-6mxw-7mvq

Vulnerable versions: >= 1.5.0, < 1.5.7

Impact:
A bug was found in containerd where container root directories and
some plugins had insufficiently restricted permissions, allowing
otherwise unprivileged Linux users to traverse directory contents
and execute programs. When containers included executable programs
with extended permission bits (such as setuid), unprivileged Linux
users could discover and execute those programs. When the UID of an
unprivileged Linux user on the host collided with the file owner or
group inside a container, the unprivileged Linux user on the host
could discover, read, and modify those files.


Totktonada (approved):
No objections from me.

dokshina (approved):
LGTM. Why didn't you update the Changelog?

dokshina (approved):

LeonidVas:
> LGTM. Why didn't you update the Changelog?

I was in a hurry, I was wrong. Added description of changes to changelog

PR #654, opened 2021-10-27
Update dependencies in templates

yngvar-antonsson:


Totktonada (approved):
No objections.

Maybe squash commits? The idea looks the same.

Since there is no issue, it would be good to describe, which problem hits you and how it presents itself. If there is no problem, just update to be in line with actual versions: it is okay, but worth to explicitly say.

DifferentialOrange:
Rebased on master (#674 already bumped cartridge and fixed tests due to behavior changes), minor cosmetic reworks in comments.

PR #663, opened 2021-11-25
dependency: update containerd version to 1.5.8

LeonidVas:
Vulnerable versions: >= 1.5.0, < 1.5.8

Impact:
In the OCI Distribution Specification version 1.0.0 and prior and
in the OCI Image Specification version 1.0.1 and prior, manifest
and index documents are ambiguous without an accompanying Content-Type
HTTP header. Versions of containerd prior to 1.4.12 and 1.5.8 treat
the Content-Type header as trusted and deserialize the document according
to that header. If the Content-Type header changed between pulls of the
same ambiguous document (with the same digest), the document may be
interpreted differently, meaning that the digest alone is insufficient
to unambiguously identify the content of the image.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,20 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Changed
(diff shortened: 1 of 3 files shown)


Totktonada (approved):
Can't say much, but I don't object.

PR #664, opened 2021-11-25
dependency: update image-spec version to 1.0.2

LeonidVas:
Vulnerable versions: 1.0.1 and prior

Impact:
In the OCI Image Specification version 1.0.1 and prior, manifest
and index documents are not self-describing and documents with a
single digest could be interpreted as either a manifest or an index.


Totktonada (approved):
No objections from me.

Issue #650, opened 2021-10-14
test_version_update_at_roles_reload fails on 1.10 and 2.7


Totktonada:
1.10 and 2.7 (but not 2.8) fail on `test_version_update_at_roles_reload`:

```
=========================== short test summary info ============================
FAILED test/e2e/test_tgz.py::test_version_update_at_roles_reload - docker.err...
ERROR test/e2e/test_tgz.py::test_tgz - AssertionError: Error during creating ...
=============== 1 failed, 4 passed, 1 error in 489.08s (0:08:09) ===============
```

DifferentialOrange:
Example: https://github.com/tarantool/cartridge-cli/runs/4325924032?check_suite_focus=true

DifferentialOrange:
https://github.com/tarantool/cartridge-cli/runs/4521568155?check_suite_focus=true

DifferentialOrange:
It looks like https://github.com/tarantool/cartridge-cli/runs/4325924032?check_suite_focus=true issues were caused by change in installer politics https://github.com/tarantool/installer.sh/commit/841bd7a9de43adf7663fc3674a94889b2110b91f and it was fixed in #629

DifferentialOrange:
Fails do not reproduce both locally and on GitHub runners

DifferentialOrange:
https://github.com/tarantool/cartridge-cli/runs/4521568155?check_suite_focus=true case looks like temporary issues with Tarantool servers (see logs), issue do not reproduces.

DifferentialOrange:
@Totktonada , I propose this ticket to be "closed by #629"

Totktonada:
Thanks for the investigation! Okay, closing as fixed in PR #629.

PR #668, opened 2021-12-20
Fix image name in tests

DifferentialOrange:
See #650 comments about closing the ticket.

I didn't forget about

- [x] Tests
- Changelog (change is minor)
- Documentation (not needed)

Follows up #650

### DIFF
--- a/test/e2e/test_tgz.py
+++ b/test/e2e/test_tgz.py
@@ -54,7 +54,7 @@ def instance_container_with_unpacked_tgz(docker_client, tmpdir, tgz_archive_with
     with open(os.path.join(build_path, 'Dockerfile'), 'w') as f:
         f.write('\n'.join(dockerfile_layers))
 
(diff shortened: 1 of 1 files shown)


Totktonada:
Am I understand right: it does NOT change any behaviour, just to make the image name corresponding its purpose?

DifferentialOrange:
> Am I understand right: it does NOT change any behaviour, just to make the image name corresponding its purpose?

Yes, it's cosmetic

Totktonada (approved):
Okay so. However I would suggest to avoid 'fix' word for changes, which does not change behaviour: it may be confusing.

LeonidVas (approved):

LeonidVas:
> Okay so. However I would suggest to avoid 'fix' word for changes, which does not change behaviour: it may be confusing.

@DifferentialOrange prefers not to change the commit title.

PR #628, opened 2021-09-02
Update Go to 1.17

mRrvz:
Golang has been updated to the newest release 1.17. 
Updated both local build and CI build. 
Bumped dependency packages (go.mod). 

Closes #625

PR #678, opened 2022-01-19
Update Go to 1.17 in release workflow

DifferentialOrange:
After #628, package is built with Go 1.17 for test workflow.
At the same time, release workflow still uses Go 1.15.
This patch removes this inconsistency.

I didn't forget about

- Tests (run with Go 1.17 since #628)
- Changelog (have the entry since #628)
- Documentation (based on #628, there are no visible changes for a user)

Follows up #628

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,7 +10,7 @@ on:
       - '*'
 
 env:
-  GO_VERSION: 1.15
+  GO_VERSION: 1.17
   GORELEASER_VERSION: v0.146.0
 
 jobs:


DifferentialOrange:
Current release of `goreleaser` is `v1.3.1`. At first I thought that it would be nice to try to migrate from zero ver to "stable" one, but it seems that it is a cosmetic change: https://carlosbecker.com/posts/goreleaser-v1/

LeonidVas (approved):

olegrok (approved):

PR #686, opened 2022-02-04
fix: remove unwanted char in version hash pattern

vsyrovat:
It seems like first 'g' char in version hash is misspelling result and totally unwanted.

It was impossible to specify version like `1.1.1-1-pre`, it was possible only with 'g' char first -- `1.1.1-1-gpre`.


DifferentialOrange:
Well, no, version hash is a result of `git describe --tags --long` and "hash" is last commit hash which is always starts with `g` and followed by 16-base commit id. It doesn't related to version suffix you're want to use. Please, try `--suffix` option or wait for https://github.com/tarantool/cartridge-cli/pull/604 to merge and use any version format you want with `--version`

PR #696, opened 2022-03-24
Fix several CVE

LeonidVas:
Fixed CVE:
* CVE-2021-38561: due to improper index calculation, an incorrectly formatted language tag can cause Parse to panic, due to an out of bounds read. If Parse is used to process untrusted user inputs, this may be used as a vector for a denial of service attack.
* CVE-2021-43565: there's an input validation flaw in ``golang.org/x/crypto readCipherPacket()`` function. An unauthenticated attacker who sends an empty plaintext packet to a program linked with ``golang.org/x/crypto/ssh`` could cause a panic, potentially leading to denial of service.


Totktonada (approved):
LGTM.

PR #699, opened 2022-04-20
Fix work with terminal

LeonidVas:
The patch adds a workaround for the bug "Terminal does not display command after go-prompt exit" (https://github.com/c-bata/go-prompt/issues/228).


0x501D (approved):
LGTM

PR #720, opened 2022-06-15
Add missing TYPE argument

patiencedaur:
Resolves tarantool/doc#2948

The command `cartridge pack --use-docker` was misleading, running it led to an error. Must have been `cartridge pack TYPE --use-docker`, for example, `cartridge pack rpm --use-docker`.

### DIFF
--- a/doc/commands/pack/building-in-docker.rst
+++ b/doc/commands/pack/building-in-docker.rst
@@ -5,7 +5,9 @@ To build your application in Docker, run this:
 
 ..  code-block:: bash
 
-    cartridge pack --use-docker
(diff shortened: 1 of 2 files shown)


LeonidVas (changes requested):
Hi! Thank you for the patch.
Please decorate commit message according to https://github.com/tarantool/tarantool/wiki/Code-review-procedure#commit-message
Also specify in commit body something like `Part of #xxx`, `Closes #xxx` ... (https://www.tarantool.io/ru/doc/latest/dev_guide/developer_guidelines/#how-to-write-a-commit-message)

PR #721, opened 2022-06-23
admin: fix run-dir path

0x501D:
From README.md in default cartridge template:

```
cartridge admin probe \
  --name void \
  --run-dir ./tmp/run \
  --uri localhost:3302
```

But this command will fails:

```
$  ../cartridge admin probe --name void --run-dir ./tmp/run --uri localhost:3302
   ⨯ Failed to connect to application instance: No available sockets found in: ./tmp/run
```

This patch allows you to use a relative path for the parameter
`--run-dir`

### DIFF
--- a/cli/commands/admin.go
+++ b/cli/commands/admin.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"path/filepath"
 	"strings"
 
 	"github.com/apex/log"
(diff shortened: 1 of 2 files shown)


DifferentialOrange (approved):
That's nice!

Maybe we should add a test case? I don't think it should be complicated since there are some tests for run-dir already.

0x501D:
> That's nice!
> 
> Maybe we should add a test case? I don't think it should be complicated since there are some tests for run-dir already.

Test case added.

DifferentialOrange (approved):
That's awesome, thank you!

LeonidVas (approved):

PR #733, opened 2022-10-28
connector: add fix for long socket paths

better0fdead:
Now relative path is used instead of real for console socket. It is used to prevent console socket path from being too long.
Before if connection addres was too long it returned an error, now it works.

Needed for [that issue](https://github.com/tarantool/tt/issues/124)

I didn't forget about

- Tests (CI is still green)
- Changelog (Added)
- Documentation (not worth to mention)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


0x501D (approved):
LGTM

better0fdead:
> Hi! Thank you for the patch.
> Why don't you add a note about the change to the CHANGELOG?

Sry, forgot, added

LeonidVas (approved):

PR #734, opened 2022-11-01
code-health: set CI/CD Go version as string

DifferentialOrange:
There is no reason to set Go version as a number: float value can lead to errors due to float precision, and in each place where the value is used it is converted to a string.

I didn't forget about

- Tests (CI is still green)
- Changelog (not worth to mention)
- Documentation (not worth to mention)


LeonidVas (approved):
I don't mind.

PR #746, opened 2022-12-27
Run post-build stage after cache update

oleggator:
At the moment, the packaging cache is updated after the post-build stage. So if folders such as node_modules, target, etc. are deleted at the post-build stage, they will not get into the cache.

In this commit, the packaging cache update stage runs before the post-build stage.

I didn't forget about

- [x] Tests
- [x] Changelog
- [ ] Documentation


psergee (approved):
Thanks for the patch. LGTM.

PR #752, opened 2023-01-16
bench: cluster patch

Kirill-Churkin:
user@cartridge-cli % ./cartridge bench --leader=127.0.0.1:3301 --replica=127.0.0.1:3302
Tarantool 2.10.2 (Binary) b6820873-f007-4413-b2f0-69fe833c0e14 

Parameters:
        Leaders:
                127.0.0.1:3301
        Replicas:
                127.0.0.1:3302
        user: guest
        connections: 10
        simultaneous requests: 10
        duration: 10 seconds
        key size: 10 bytes
        data size: 20 bytes
        insert: 100 percentages
        select: 0 percentages
        update: 0 percentages

Data schema
| key                            | value
| ------------------------------ | ------------------------------
| random(10)                     | random(20)
Benchmark start
...
Benchmark stop.

Results:
        Success operations: 760386
        Failed  operations: 0
        Request count: 760474
        Time (seconds): 9.979464
        Requests per second: 76203


cli/bench/types.go
```diff
@@ -16,15 +18,31 @@ type Results struct {
 	requestsPerSecond    int     // Cumber of requests per second - the main measured value.
 }
 
+// RotaryConnectionsPool describes round-cycled connection pool.
```
DifferentialOrange, line 21: For the future, I think you may consider developing bench as [`tt`](https://github.com/tarantool/tt) extension. `tt` uses [tarantool/go-tarantool](https://github.com/tarantool/go-tarantool) as a connector, and go-tarantool has built-in support of connection pools. It is also worth to mention that you should be able to develop an extention within your own repo.

cli/bench/cluster.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	benchData.startTime = time.Now()
+
+	// Start detached connections.
+	for i := 0; i < ctx.Connections; i++ {
+		benchData.waitGroup.Add(1)
+		go func() {
+			defer benchData.waitGroup.Done()
+			requestsSequence := RequestsSequence{
```
DifferentialOrange, line 181: Now requests generation is the part of the bench run tracked by timer. I think it's better to build request data before starting the timer, it possible.

DifferentialOrange (commented):
Thank you for your contribution! Seems legit at most, I have left a few comments.

DifferentialOrange (changes requested):
Just a status clarification, since GitHub treats this as "need review"

sergos:
@LeonidVas please consider opportunity to transfer/adapt the test to tt perf bench.

PR #765, opened 2023-04-19
bugfix: specifying confusing instance names

### LINKED ISSUE #763: Support specifying confusing instance names
### 1. Application name as instance name

Now we don't support instance names which coincide with application names:

```bash
cat ./instances.yml 
---
myapp.myapp:
  advertise_uri: localhost:3301
  http_port: 8081
```
```bash
cartridge start -d
   • myapp.myapp... OK
```
```bash
cartridge status
   • myapp.myapp: RUNNING
```
```bash
cartridge status myapp
   ⨯ Application name is specified. Please, specify instance name(s)
```

### 2. Instance name with dots

The reason is this code in `GetInstancesFromArgs` method:
https://github.com/tarantool/cartridge-cli/blob/80f0f6374c320e36806e247c406a49665fc41e57/cli/common/utils.go#L385-L387

We also do not support instance names with dots:

```bash
cat ./instances.yml 
---
myapp.my.app:
  advertise_uri: localhost:3301
  http_port: 8081
```
```bash
cartridge start -d
   • myapp.my.app... OK
```
```bash
cartridge status
   • myapp.my.app: RUNNING
```
```bash
cartridge status my.app
   ⨯ [APP_NAME].INSTANCE_NAME is specified. Please, specify instance name(s)
```

The reason is this code in `GetInstancesFromArgs` method:
https://github.com/tarantool/cartridge-cli/blob/80f0f6374c320e36806e247c406a49665fc41e57/cli/common/utils.go#L389-L393

### 3. Specifying an instance name multiple times

`GetInstancesFromArgs` also doesn't allow to specify an instance multiple times:
https://github.com/tarantool/cartridge-cli/blob/80f0f6374c320e36806e247c406a49665fc41e57/cli/common/utils.go#L400-L402

------------------------------------------------------------------------------------------------------

### Proposed solution

The reason of the issues 1 and 2 is the desire to help a user if they specified `application_name` or `application_name.instance_name` instead of `instance_name`. It results in inability to use some instance names in some commands which are (likely) supported by any other command. We need to

1. allow a user to use such names
2. after verifying that any other command is fine with that.

To preserve existing helpful error messages in case user really specified `application_name` or `application_name.instance_name` instead of `instance_name` (which is rather often), we'll need to rework our approach. If user had provided an invalid info, the command should fail. And if the command had failed,

3. we additionally check if they had specified `application_name` or `application_name.instance_name` instead of `instance_name` and, if they are, add this info to error message.

Regarding issue 3, we may simply

4. start to skip duplicate names instead of throwing an error,

but, since there are no such request from customers yet, we may ignore this one for now.

Based on https://jira.vk.team/browse/TNT-765

GRISHNOV:
Added support for instance names:
* Application name as instance name
* Instance name with dots

Closes #763

### DIFF
--- a/cli/common/utils.go
+++ b/cli/common/utils.go
@@ -377,21 +377,11 @@ func InsertInStringSlice(s []string, i int, elem string) []string {
 	return res
 }
 
-func GetInstancesFromArgs(args []string, projectName string) ([]string, error) {
+func GetInstancesFromArgs(args []string) ([]string, error) {
 	foundInstances := make(map[string]struct{})
 	var instances []string
 
 	for _, instanceName := range args {
-		if instanceName == projectName {
-			return nil, fmt.Errorf(appNameSpecifiedError)
-		}
-
-		parts := strings.SplitN(instanceName, ".", 2)
-
-		if len(parts) > 1 {
-			return nil, fmt.Errorf(instanceIDSpecified)
-		}
-
 		if instanceName == "stateboard" {
 			return nil, fmt.Errorf("Please, specify flag --stateboard for processing stateboard instance")
 		}
@@ -407,6 +397,20 @@ func GetInstancesFromArgs(args []string, projectName string) ([]string, error) {
 	}
 
 	return instances, nil
+}
+
+func ErrWrapCheckInstanceNameCommonMisprint(instanceNames []string, projectName string, err error) error {
+	for _, instanceName := range instanceNames {
+		if instanceName == projectName {
+			return fmt.Errorf(err.Error() + ": " + appNameSpecifiedError)
+		}
+		instanceNameParts := strings.SplitN(instanceName, ".", 2)
+		if instanceNameParts[0] == projectName {
+			return fmt.Errorf(err.Error() + ": " + instanceIDSpecified)
+		}
+	}
+
+	return err
 }
 
 func GetStringSlicesDifference(s1, s2 []string) []string {
@@ -582,7 +586,9 @@ func ContainsUpperSymbols(src string) bool {
 
 const (
 	appNameSpecifiedError = "Application name is specified. " +
-		"Please, specify instance name(s)"
-	instanceIDSpecified = `[APP_NAME].INSTANCE_NAME is specified. ` +
-		"Please, specify instance name(s)"
+		"You may have made a typo. " +
+		"Please, try to specify instance name(s)"
+	instanceIDSpecified = "[APP_NAME].INSTANCE_NAME is specified. " +
+		"You may have made a typo. " +
+		"Please, try to specify instance name(s)"
 )
--- a/cli/common/utils_test.go
+++ b/cli/common/utils_test.go
@@ -114,40 +114,45 @@ func TestGetInstancesFromArgs(t *testing.T) {
 	var args []string
 	var instances []string
 
-	projectName := "myapp"
-
-	// wrong format
-	args = []string{"myapp.instance-1", "myapp.instance-2"}
-	_, err = GetInstancesFromArgs(args, projectName)
-	assert.EqualError(err, instanceIDSpecified)
-
-	args = []string{"instance-1", "myapp.instance-2"}
-	_, err = GetInstancesFromArgs(args, projectName)
-	assert.EqualError(err, instanceIDSpecified)
-
-	args = []string{"myapp"}
-	_, err = GetInstancesFromArgs(args, projectName)
-	assert.True(strings.Contains(err.Error(), appNameSpecifiedError))
-
 	// duplicate instance name
 	args = []string{"instance-1", "instance-1"}
-	_, err = GetInstancesFromArgs(args, projectName)
+	_, err = GetInstancesFromArgs(args)
 	assert.True(strings.Contains(err.Error(), "Duplicate instance name specified: instance-1"))
 
 	// instances are specified
 	args = []string{"instance-1", "instance-2"}
-	instances, err = GetInstancesFromArgs(args, projectName)
+	instances, err = GetInstancesFromArgs(args)
 	assert.Nil(err)
 	assert.Equal([]string{"instance-1", "instance-2"}, instances)
+}
 
-	// specified both app name and instance name
-	args = []string{"instance-1", "myapp"}
-	instances, err = GetInstancesFromArgs(args, projectName)
-	assert.EqualError(err, appNameSpecifiedError)
+func TestCheckInstanceNameTypoWithProjectName(t *testing.T) {
+	t.Parallel()
 
-	args = []string{"myapp", "instance-1"}
-	instances, err = GetInstancesFromArgs(args, projectName)
-	assert.EqualError(err, appNameSpecifiedError)
+	assert := assert.New(t)
+
+	var err error
+	var instanceName string
+	var projectName string
+	var errDesc string
+
+	instanceName = "mytestapp.mytestinstance"
+	projectName = "mytestapp"
+	errDesc = fmt.Sprintf("Instance %s is not running", instanceName)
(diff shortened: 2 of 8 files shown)


cli/connect/connect.go
```diff
@@ -33,6 +34,12 @@ func Enter(ctx *context.Ctx, args []string) error {
 
 	process := running.NewInstanceProcess(ctx, instanceName)
 	if !process.IsRunning() {
+		instanceNameParts := strings.SplitN(instanceName, ".", 2)
```
DifferentialOrange, line 37: Let's use the same approach as it was in util before: separate check for applicationName instead of instanceName and applicationName.instanceName instead of instanceName.

cli/replicasets/failover_priority.go
```diff
@@ -21,7 +21,7 @@ func SetFailoverPriority(ctx *context.Ctx, args []string) error {
 		return err
 	}
 
-	if ctx.Replicasets.FailoverPriorityNames, err = common.GetInstancesFromArgs(args, ctx.Project.Name); err != nil {
+	if ctx.Replicasets.FailoverPriorityNames, err = common.GetInstancesFromArgs(args); err != nil {
```
DifferentialOrange, line 24: So we don't get any additional info for `replicasets` command error? I think we should

cli/connect/connect.go
DifferentialOrange: You may extract it to a separate util like the one we used before, since it seems like we may reuse it in several places

cli/common/utils_test.go
```diff
@@ -114,40 +114,16 @@ func TestGetInstancesFromArgs(t *testing.T) {
 	var args []string
 	var instances []string
 
-	projectName := "myapp"
-
-	// wrong format
-	args = []string{"myapp.instance-1", "myapp.instance-2"}
```
DifferentialOrange, line 120: If it is possible and not too time-consuming, we should have tests for our new approach.

DifferentialOrange (commented):
Thank you for the patch. At the first glance, it seems like we do not yet covered every case we should. Please, correct me if I'm wrong.

cli/replicasets/failover_priority.go
GRISHNOV: Done, added for replicasets `join`, `expel` and `set-failover-priority`

cli/common/utils_test.go
GRISHNOV: Added unit test

cli/common/utils.go
```diff
@@ -409,6 +399,17 @@ func GetInstancesFromArgs(args []string, projectName string) ([]string, error) {
 	return instances, nil
 }
 
+func CheckInstanceNameTypoWithProjectName(instanceName string, projectName string, errDesc string) error {
+	instanceNameParts := strings.SplitN(instanceName, ".", 2)
+	if instanceNameParts[0] == projectName {
+		return fmt.Errorf(errDesc+
+			"You may have specified an unnecessary prefix as the application name, "+
```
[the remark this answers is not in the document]
GRISHNOV: Done

DifferentialOrange (commented):
Sorry, I have forgot to submit this batch of comments

cli/common/utils.go
```diff
@@ -409,6 +399,17 @@ func GetInstancesFromArgs(args []string, projectName string) ([]string, error) {
 	return instances, nil
 }
 
+func CheckInstanceNameTypoWithProjectName(instanceName string, projectName string, errDesc string) error {
+	instanceNameParts := strings.SplitN(instanceName, ".", 2)
+	if instanceNameParts[0] == projectName {
```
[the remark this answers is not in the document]
GRISHNOV: Done

cli/common/utils.go
```diff
@@ -409,6 +399,17 @@ func GetInstancesFromArgs(args []string, projectName string) ([]string, error) {
 	return instances, nil
 }
 
+func CheckInstanceNameTypoWithProjectName(instanceName string, projectName string, errDesc string) error {
```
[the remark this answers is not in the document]
GRISHNOV: Done, now `CheckInstanceNameCommonMisprint`

cli/running/process.go
```diff
@@ -390,6 +390,16 @@ func NewStateboardProcess(ctx *context.Ctx) *Process {
 
 func (process *Process) Log(follow bool, n int) error {
 	if _, err := os.Stat(process.logFile); err != nil {
+		if os.IsNotExist(err) {
+			logFilePathParts := strings.SplitN(process.logFile, "/", -1)
+			logFileName := logFilePathParts[len(logFilePathParts)-1]
+			logFilePathParts[len(logFilePathParts)-1] = strings.SplitN(logFileName, ".", 2)[1]
+			possibleLogFilePath := strings.Join(logFilePathParts, "/")
+			if _, err := os.Stat(possibleLogFilePath); err == nil {
+				return fmt.Errorf("Failed to use process log file: " +
```
DifferentialOrange, line 399: Can you reuse existing util here too?

cli/replicasets/expel.go
```diff
@@ -64,6 +65,10 @@ func Expel(ctx *context.Ctx, args []string) error {
 
 	for instanceName, instanceFound := range instancesToExpelMap {
 		if !instanceFound {
+			if err := common.CheckInstanceNameTypoWithProjectName(instanceName, ctx.Project.Name,
+				fmt.Sprintf("Instance %s isn't found in cluster: ", instanceName)); err != nil {
```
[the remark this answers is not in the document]
DifferentialOrange: Yes. I think we may simply return the error in wrapper. Wrapper should add useful comment, if it makes sence, and return the same unwrapped error otherwise.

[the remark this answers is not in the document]
GRISHNOV: Done

cli/running/process.go
GRISHNOV: Hmm, it looks like it can only be moved to `utils`, since the check logic is different here

DifferentialOrange: You can wrap this error on the upper level in `cli/running/running.go`. You can do the same for `Status` too (we decided not to wrap this one in the start, but if you implement this in `Log`, implementing this for `Status` would be a 10 seconds task).

GRISHNOV: Done for `Log`

AFAIU, it is not possible to implement this check at the same level for the`Status`

May be it is possible to do this at a lower level (like for `Log` before), but now we have a print:
```
./cartridge status mytestapp.mytestapp                             
   • mytestapp.mytestapp.mytestapp: NOT STARTED
```

I mean, here users can see that there is an extra unnecessary prefix of the specified name before `NOT STARTED`. Do we need a check here?

cli/common/utils.go
```diff
@@ -582,7 +586,9 @@ func ContainsUpperSymbols(src string) bool {
 
 const (
 	appNameSpecifiedError = "Application name is specified. " +
-		"Please, specify instance name(s)"
-	instanceIDSpecified = `[APP_NAME].INSTANCE_NAME is specified. ` +
-		"Please, specify instance name(s)"
+		"You may have made a misprint. " +
```
DifferentialOrange, line 589: I think the "typo" word is more widely used in such cases

cli/running/process.go
DifferentialOrange: Let's forget about the status then

cli/common/utils.go
GRISHNOV: Done

cli/running/process.go
DifferentialOrange: It seems to be even more interesting than that: you can start and stop instances with arbitrary names, see https://github.com/tarantool/cartridge-cli/issues/766

We decided not to cover this issue in this PR

DifferentialOrange (approved):

PR #775, opened 2023-08-01
cd: update go version

psergee:
The previous version of go had a number of security issues. Upgraded to 1.20.5.


0x501D (approved):
Thank you for the patch!

PR #801, opened 2024-11-22
fix: stdout logging

slavische:
Redirect logging to stdout instead of syslog.

### DIFF
--- a/cli/create/templates/cartridge/init.lua
+++ b/cli/create/templates/cartridge/init.lua
@@ -37,13 +37,14 @@ end
 local cartridge = require('cartridge')
 
 local ok, err = cartridge.cfg({
-    roles = {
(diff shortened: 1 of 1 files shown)


oleg-jukovec:
Please write in more detail why we need this change?

slavische:
> Could you please describe the issue you are experiencing?

When developing locally it is not very convenient to see logs in syslog, in earlier versions logs were written to stdout. I would like to return this behavior.

oleg-jukovec:
Thank you for the patch.

The utility is deprecated and will likely be moved to a public archive soon. Please, use the [tt](https://github.com/tarantool/tt) for [cartridge applications](https://github.com/tarantool/tt/blob/master/doc/examples.md#creating-cartridge-application) instead. We will be glad to see your PRs there.

